PR #5, opened 2026-02-24
Fix packaged app signing/resource layout for 0.4.2

T0mSIlver:
## Summary
- move copied SwiftPM resource bundles into `Contents/Resources` so the app bundle remains structurally signable
- patch ShortcutRecorder bundle lookup during packaging so the packaged app can resolve `ShortcutRecorder_ShortcutRecorder.bundle` from app resources
- strip extended attributes from packaged contents before signing, then ad-hoc sign + verify to avoid `source=no usable signature`

## Validation
- `swift test`
- `./scripts/package_app.sh debug 0.4.2 1`
- `codesign --verify --deep --strict --verbose=2 dist/localvoxtral.app`
- `spctl --assess --type execute -vv dist/localvoxtral.app` (rejected as expected without Developer ID/notarization, but no `no usable signature`)

### DIFF
--- a/scripts/package_app.sh
+++ b/scripts/package_app.sh
@@ -4,6 +4,20 @@ set -euo pipefail
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "$ROOT_DIR"
 
+patch_shortcutrecorder_bundle_lookup() {
+  local sr_common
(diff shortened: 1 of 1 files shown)


T0mSIlver:
Superseded by #6, which is rebased on current main and contains only the 0.4.2 signing/resource fix.

PR #16, opened 2026-03-19
fix: overlay clipping, Fn hotkey re-enable, escape cancel, click guard

joostliebregts:
Hey! I've been using localvoxtral daily for dictation and ran into a few small bugs. Fixed them for myself and figured I'd share in case they're useful to you. Totally understand if you'd rather handle these differently — just sharing what worked for me.

## Fixes

- **Overlay bottom clipping**: `measureContentHeight` counted 1 VStack spacing gap instead of 2, making the panel 8pt too short. Bottom line got clipped during the 4→5 line scroll transition. Also added 12pt bottom padding inside the ScrollView content as buffer.

- **Fn modifier-only hotkey dies after focus loss**: macOS disables `.listenOnly` CGEventTaps after the app loses focus or after a timeout. Added handling for `tapDisabledByTimeout` / `tapDisabledByUserInput` events to auto-re-enable the tap.

- **Escape cancel still pastes**: `finishStoppedSession` committed the overlay buffer text even when `wasCancelled` was true. Now skips both pending segment promotion and overlay commit when cancelled.

- **Overlay body clicks steal focus**: Clicking on the overlay panel body (below drag handle) forwarded mouse events to the SwiftUI view, which could steal focus from the target app and disrupt AX text insertion. Now swallowed.

## Test plan

- [ ] Dictate 5+ lines, verify bottom line stays visible during scroll transition
- [ ] Set Fn as modifier-only key, switch to another app, tap Fn again — should still work
- [ ] Start dictation, press Escape — text should NOT be pasted
- [ ] Click on overlay panel body during dictation — should not affect target app focus
---

### DIFF
--- a/Sources/localvoxtral/DictationOverlayController.swift
+++ b/Sources/localvoxtral/DictationOverlayController.swift
@@ -27,9 +27,23 @@ private final class OverlayContainerView: NSView {
     override var isOpaque: Bool { false }
 }
 
+/// A panel that swallows mouse events on its body so they don't steal
+/// keyboard focus or become the key/main window. This prevents
+/// interference with Accessibility-based text insertion.
+private final class NonActivatingPanel: NSPanel {
+    override var canBecomeKey: Bool { false }
+    override var canBecomeMain: Bool { false }
+
+    override func mouseDown(with event: NSEvent) {
+        // Swallow all mouse clicks on the overlay body.
+        // Forwarding to super can steal focus from the target app
+        // and disrupt AX text insertion.
+    }
+}
+
 @MainActor
 final class DictationOverlayController {
-    private let panel: NSPanel
+    private let panel: NonActivatingPanel
     private let hostingView: TransparentHostingView<DictationOverlayView>
     private let panelWidth: CGFloat = 420
     private let maximumPanelHeight: CGFloat = 420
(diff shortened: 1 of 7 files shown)


T0mSIlver:
Hey @joostliebregts — sorry for the long silence, development was paused and has just resumed in earnest. Thanks a lot for these fixes, and for daily-driving the app. Escape-to-cancel and the overlay clipping fix are exactly the polish I want to land first.

Plan for this PR:
- `ModifierOnlyHotKeyManager.swift` should come out of this branch — nothing here references it, and it collides with the richer version in #17, where it belongs.
- The `.cancelled` phase case is added but never set — either wire it into the cancel path or drop it for now.
- One behavior question on the click guard: the overlay used to pass clicks through, now it swallows them. Intentional beyond preventing focus-steal? I want to be sure clicks over the overlay region behave sanely and dragging isn't affected.

The repo now runs CI on every PR (build + tests + live-backend integration), so a rebase onto main gets you a green check. Happy to make these trims myself and push to your branch if you prefer — you offered "take it as inspiration", but I'd rather keep your authorship.

T0mSIlver:
#30 (now merged) landed the trimmed version of this PR with your commit and authorship intact: the overlay clipping fix, the click guard, and escape-to-cancel — the latter reworked onto a Carbon hotkey after real-device testing showed macOS refuses keyboard event taps without the Input Monitoring permission (details in #30). The Fn re-enable piece belongs with #17's mechanism and will be handled there. Thanks again for these — closing as superseded by #30.

PR #17, opened 2026-03-19
feat: tap-vs-hold gesture for modifier-only hotkey

joostliebregts:
Hey! Built this for myself because I wanted two dictation modes on a single key. Tap Fn for overlay buffer (toggle on/off), hold Fn for live auto-paste (push-to-talk style). Works great for my workflow but totally get it if it's not your thing.

## What it does

- **Tap** (press + release before threshold) → starts/stops overlay buffer mode (toggle)
- **Hold** (press past threshold) → starts live auto-paste, stops on release (push-to-talk)
- **Hold delay** configurable in Settings (150–800ms, default 350ms) so you can tune what feels right

## How it works

- `ModifierOnlyHotKeyManager` replaces `onPress`/`onRelease` with `onTap`/`onHoldStart`/`onHoldRelease`
- Uses a `DispatchWorkItem` timer to differentiate tap from hold
- `HotKeyManager` wires tap→overlayBuffer and hold→liveAutoPaste
- `DictationViewModel` gets dedicated handlers that bypass the global shortcutMode setting
- Key interruption during pending state cancels the gesture (using Fn as actual modifier combo still works)

## Test plan

- [ ] Tap Fn → overlay buffer starts. Tap again → stops (toggle)
- [ ] Hold Fn → live auto-paste starts. Release → stops
- [ ] Hold Fn, press another key while held → no dictation (modifier combo)
- [ ] Adjust hold delay slider in Settings → feel the difference
- [ ] Existing keyboard shortcut paths still work when modifier-only is disabled
---

### DIFF
--- a/Sources/localvoxtral/DictationViewModel.swift
+++ b/Sources/localvoxtral/DictationViewModel.swift
@@ -210,6 +210,9 @@ final class DictationViewModel {
     // to be cancelled if the user releases before dictation actually begins.
     @ObservationIgnored
     private var hasActivePushToTalkShortcutSession = false
+    // True when a modifier-only hold gesture started dictation (push-to-talk semantics).
+    @ObservationIgnored
+    private var isModifierOnlyHoldActive = false
 
     init(
         settings: SettingsStore,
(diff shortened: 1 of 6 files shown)


T0mSIlver:
My favorite of the three — tap-vs-hold is a great interaction model, and the dual-shortcut support with right-cmd/right-option handling effectively implements #20 too (worth adding to the description; it's a big chunk of the diff and it's good work).

Plan: #16 lands first (trimmed as discussed there), then this after a rebase — expect small conflicts in `DictationViewModel.swift` since main has moved. My review focus will be the shared mutable state in the hotkey manager (the `nonisolated(unsafe)` statics touched from the event tap and the hold-timer queue); I may rework that part for strict-concurrency friendliness during review, behavior unchanged.

PR #18, opened 2026-03-19
feat: gradual DDC audio ducking fade-in with configurable duration

joostliebregts:
Hey! I use localvoxtral with HDMI audio on a Mac Mini, so ducking goes through the DDC/BetterDisplay path. The instant volume restore after a long dictation session was pretty jarring (especially with headphones), so I added a gradual fade-in.

**Note:** This PR builds on #17 (tap-vs-hold gesture) because the DictationViewModel changes overlap. Can be reviewed independently but merges cleaner after #17.

## What it does

- **DDC fade-in**: Restores monitor volume gradually via BetterDisplay CLI in 15 steps (instead of instant)
- **CoreAudio fade-in**: Same gradual restore for headphones/built-in speakers via GCD loop
- **Fade duration setting**: Configurable 0.3–5.0s in Settings (default 1.5s)
- **Fade cancellation**: Generation counter cancels stale fades when new duck/unduck starts mid-fade
- **Volume drift prevention**: Duck target calculated from stored original volume, not mid-fade level. Rapid start/stop cycles no longer erode the base volume.

## How it works

- `AudioDuckingService` stores `_ddcOriginalVolume` / `_originalVolume` on first duck, clears only on fade completion
- `_fadeGeneration` counter incremented on every duck/unduck — fade loops bail if their generation is stale
- `duckViaDDC` uses `baseVolume = _ddcOriginalVolume ?? currentVolume` so mid-fade re-ducks still target the true original

## Test plan

- [ ] Dictate → stop → volume fades back gradually (not instant)
- [ ] Adjust fade-in slider in Settings → duration changes
- [ ] Rapid start/stop 5x → volume returns to original (no drift)
- [ ] Start new dictation while fade-in is running → fade cancels, volume ducks immediately
---

### DIFF
--- a/Sources/localvoxtral/AudioDuckingService.swift
+++ b/Sources/localvoxtral/AudioDuckingService.swift
@@ -1,0 +1,417 @@
+import AppKit
+import AudioToolbox
+import CoreAudio
+import Foundation
+import os
+
+/// Ducks audio when dictation starts and restores when it stops.
+///
+/// Uses a multi-strategy approach since not all audio output devices
+/// support software volume control (e.g. HDMI/DisplayPort monitors):
+///
+/// 1. Try CoreAudio VirtualMainVolume (headphones, built-in speakers)
+/// 2. Duck DDC monitor volume via BetterDisplay binary (when `useDDC` is true and BetterDisplay is installed)
+/// 3. Fall back to pausing common music apps (Spotify, Music) via NSAppleScript
+/// 4. Resume paused apps on unduck
+final class AudioDuckingService: @unchecked Sendable {
+    private let lock = NSLock()
+    private var _originalVolume: Float?
+    private var _duckedVolume: Float?
+    private var _fadeTask: Task<Void, Never>?
+    private var _pausedApps: [String] = []
+    private var _volumeControlAvailable = false
+    private var _ddcOriginalVolume: Float?
+    private var _ddcDuckedVolume: Float?
+    private var _ddcDucked = false
+    /// Incremented on every duck/unduck to cancel stale fade loops.
+    private var _fadeGeneration: Int = 0
+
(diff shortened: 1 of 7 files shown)


T0mSIlver:
Thanks for this one too — the fade design (generation counter, drift prevention on rapid start/stop) is thoughtful. Two things before it can move:

1. It currently contains all of #17 squashed in, minus its tests. Once #16/#17 land, a rebase should shrink it to the ~400-line ducking feature itself, reviewable on its own merits.
2. The bigger question is product-level: ducking here depends on the BetterDisplay CLI at a hardcoded path plus AppleScript control of Spotify/Music — a heavier dependency footprint than anything else in the app. I'm open to the feature, but it needs to be clearly optional: off by default, degrading gracefully when BetterDisplay isn't installed. Up for reshaping it that way after the rebase?

T0mSIlver:
Thanks for this, and sorry it sat so long.

Closing the PR, not the idea: it has been conflicting since March and it carries the gesture work from #17, so there is no clean path from this branch to main any more. What it found is worth more than the diff — the drift and the mid-fade cancellation are the two problems anyone implementing this hits second, and you hit them first. Both are written into [#375](https://github.com/T0mSIlver/localvoxtral/issues/375) as requirements, with credit to you, along with the thing you actually reported: the instant restore is the jarring part, so the fade is the feature, not a flag on it.

Two constraints that PR could not have known about and that the new one states: fades get an injected clock and sleep seam rather than a real `Task.sleep` (no wall clock in tests), and the BetterDisplay CLI has to be read through `POSIXPipeRead`, because `FileHandle.availableData` raises an uncatchable ObjC exception on descriptor errors and aborted the app in the field.

If you want to take #375 yourself, say so and it is yours.

PR #30, opened 2026-07-02
Trimmed & rebased #16: overlay clipping, escape-cancel, overlay click guard

T0mSIlver:
## What

Trimmed and rebased version of #16, crediting @joostliebregts who did all of
the original work in #16 (commit `1caffbe`, March). The maintainer asked for a
trimmed version rebased onto current `main`, **preserving Joost's authorship**.
His original commit is cherry-picked verbatim (`548c816`, author
`Joost MF Liebregts <developer@purposedesign.nl>`); the trims and one small
correctness fix are follow-up commits.

### Kept from #16 (all by @joostliebregts)
1. **Overlay bottom-clipping fix** — rewrites the overlay body into a
   `ScrollViewReader`/`ScrollView` with a 4-line cap and animated
   scroll-to-bottom, with matching height math in
   `DictationOverlayController.measureContentHeight`
   (`maxScrollableBodyHeight = singleLineHeight * 4 + 8`).
2. **Overlay click guard** — new `NonActivatePanel` (`canBecomeKey`/`canBecomeMain`
   return `false`, `mouseDown` swallows clicks) with `ignoresMouseEvents = false`,
   so overlay clicks no longer steal focus from the target app / disrupt AX text
   insertion.
3. **Escape-to-cancel** — new `EscapeCancelHandler` (session-wide `CGEventTap`
   consuming Escape during dictation) + `wasCancelled` plumbing through
   `DictationViewModel` / `+Session` so cancelled sessions skip segment promotion
   and overlay commit and just dismiss.

### Trims (follow-up commits, why each was dropped)
- **`ModifierOnlyHotKeyManager.swift` removed** (~143 lines). Dead code in this PR:
  nothing references it (only self-references). It belongs to the separate
  Fn/Globe-as-hotkey feature tracked in #17. `git grep ModifierOnlyHotKeyManager`
  now returns nothing.
- **`OverlayBufferPhase.cancelled` + "Cancelled" label removed.** Added by #16 but
  never set anywhere — the cancel path calls `overlayBufferCoordinator.reset()` and
  runs cleanup directly, it never transitions the phase to `.cancelled`. Dead state;
  dropped to keep the state machine minimal. The overlay simply disappears on cancel.
- **"Fn hotkey re-enable" fix deferred to #17.** That fix lived inside the removed
  `ModifierOnlyHotKeyManager`, so it is not part of this trimmed PR. It will land
  with #17 (the Fn-hotkey feature) where it belongs.

### EscapeCancelHandler leak review (per request)
`EscapeCancelHandler` toggles a `nonisolated(unsafe) static isDictatingRef` that
the CGEventTap callback checks; if left `true`, Escape is consumed system-wide.
`isDictatingRef` is set `true` in exactly one place (`startAudioCaptureAfterConnection`,
paired with `isDictating = true`), so the invariant is "mirror `isDictating`".
There are three runtime sites that set `isDictating = false`; two of them
(`stopDictation`, `abortConnectingSession`) already reset the flag and stop the
handler. The third — **`handleDisconnectedEvent` (unexpected websocket drop during
active dictation)** — did **not**, so a connection loss while dictating left the
tap armed and swallowed Escape everywhere until the next session started/stopped.

This is fixed in a follow-up commit (`0899727`) by mirroring the reset on that
third teardown path, with a regression test that arms the flag, drives a
`.disconnected` event during dictation, and asserts the flag is cleared.

### Click-guard analysis (code-level; no GUI available here)
- **Behavior change:** previously `panel.ignoresMouseEvents = true`, so all mouse
  events (clicks, drag, scroll) passed straight through the transparent overlay to
  the app/window behind. Now `ignoresMouseEvents = false` and `NonActivatePanel`
  overrides `mouseDown` with an empty body (swallow), while `canBecomeKey`/`canBecomeMain`
  return `false`. **Clicks on the overlay body are now consumed** (they no longer
  reach the app behind) and the panel never becomes key/main (no focus/activation
  theft, no menu-bar switch).
- **Dragging:** the overlay is **not** draggable, and was not before either — there
  is no `mouseDragged`, no `isMovable`/`isMovableByWindowBackground`. Position is
  locked per-session in `positionPanel`/`lockedPlacement` (top edge fixed, grows
  downward). So this PR does not regress dragging (it never existed).
- **Scroll:** `mouseDown` is overridden but `scrollWheel` is not. Because the body
  is now a SwiftUI `ScrollView`, scroll-wheel events over the overlay now scroll the
  overlay text (intended) rather than passing through to the app behind.

### Required manual checks on the Mac (UI-affecting — please verify before merge)
Until the automated UI tier (tier 2) exists, these need hand verification:
1. Overlay-buffer session visible → click the overlay body: confirm the target app
   keeps keyboard focus (caret stays) and the panel does **not** become key/main
   (menu bar does not switch). Confirm text still commits/inserts correctly.
2. While dictating, press **Escape**: overlay disappears, nothing is inserted.
   Then confirm Escape works normally in another app between sessions (not swallowed
   globally). In particular confirm it works right after a session that ended via a
   connection drop (the leak path now covered by unit test).
3. Long transcript (>4 lines): overlay body scrolls, auto-scrolls to bottom as new
   text arrives, and the bottom line is **no longer clipped** (the original bug).
   Check both above-caret and below-caret placement, and 1–2 line / exactly-4-line
   transcripts.
4. Scroll wheel over the overlay scrolls its own text (new behavior) rather than the
   app behind; overlay remains non-draggable and non-focus-stealing.

## Proof

- [x] Unit suite green — `LV_BUILD_DIR=work/localvoxtral-pr16 ./scripts/remote-build.sh test`
      ```
      Executed 212 tests, with 0 failures (0 unexpected) in 1.363 (1.373) seconds
      ```
      (208 baseline + 4 new: `testUnexpectedDisconnectDuringDictationResetsEscapeCancelFlag`,
      `testCancelledOverlaySessionSkipsSegmentPromotionAndCommit`,
      `testCancelDictationDuringActiveDictationStopsWithoutOverlayCommit`,
      `testCancelDictationIsNoOpWhenNoSessionActive`.)
- [x] Realtime integration green — `LV_BUILD_DIR=work/localvoxtral-pr16 ./scripts/remote-build.sh integration`
      ```
      Executed 7 tests, with 2 tests skipped and 0 failures (0 unexpected) in 16.435 (16.437) seconds
      ```
      (the 2 skips are the gated mic-capture tests — normal.)
- [x] Bug fix: regression test added — failing run **before** the fix and passing run **after**.
      Before (fix reverted):
      ```
      testUnexpectedDisconnectDuringDictationResetsEscapeCancelFlag ... /Users/builder/work/localvoxtral-pr16/Tests/localvoxtralTests/DictationViewModelOverlayLifecycleTests.swift:623: error: ... XCTAssertFalse failed
      Executed 1 test, with 1 failure (0 unexpected) in 0.124 seconds
      ```
      After (fix restored):
      ```
      testUnexpectedDisconnectDuringDictationResetsEscapeCancelFlag ... passed (0.003 seconds)
      Executed 22 tests, with 0 failures (0 unexpected) in 1.221 seconds
      ```
- [ ] UI change: not hand-verified here (no GUI on this dev box). See **Required manual
      checks on the Mac** above — needs maintainer sign-off before merge.


T0mSIlver:
## Escape-cancel: root cause, fix, and how to re-test

### Root cause
Escape did nothing during dictation in the packaged app, yet every unit test passed. I traced the full lifecycle (`EscapeCancelHandler` instantiation → `CGEvent.tapCreate` → run-loop source → callback → `onCancel` → `cancelDictation`) and confirmed the **ViewModel wiring is correct**: the tap is armed in the single shared path `startAudioCaptureAfterConnection` (used by **both** output modes and **both** shortcut modes — push-to-talk and toggle all funnel through `beginDictationSession` → connect → `startAudioCaptureAfterConnection`), and every teardown path (`stopDictation`, `cancelDictation`, `handleDisconnectedEvent`, `abortConnectingSession`) disarms it. So this is a **CGEventTap runtime failure the tests never reached**:

1. **`start()` silently returned when `CGEvent.tapCreate` failed.** An active (`.defaultTap`) event tap requires the process to be Accessibility-trusted *at call time*. `tapCreate` returns `nil` (no exception, no log) when trust is absent — e.g. the first dictation session of a freshly-launched app before AX is granted, or after a TCC state change that needs an app relaunch. The old code was `guard let tap = ... else { return }` — zero diagnostics. (Note: text *insertion* can still work via the Cmd+V keyboard fallback without AX trust, so dictation looked healthy while the tap was dead.)
2. **The callback never handled `tapDisabledByTimeout` / `tapDisabledByUserInput`.** The system periodically disables an active tap and delivers one of those sentinel event types; the callback must re-enable the tap or Escape interception silently dies for the rest of the session.
3. The run-loop source was added with `CFRunLoopGetCurrent()` instead of `CFRunLoopGetMain()`.

### Fix (`EscapeCancelHandler.swift` + `Log.swift`)
- Gate creation on `AXIsProcessTrusted()` first → logs an **actionable** message instead of a silent no-op.
- Distinct log when `tapCreate` returns `nil` despite trust (points at the relaunch-after-grant case).
- **Handle `tapDisabledByTimeout`/`tapDisabledByUserInput`** in the callback and re-enable the tap.
- Pin the source to `CFRunLoopGetMain()`.
- New `Log.escape` category (subsystem `com.localvoxtral`, category `Escape`) with diagnostics for every outcome so the next manual run is conclusive even if the fix is incomplete.

### What to observe in Console.app (filter: `subsystem:com.localvoxtral  category:Escape`)
On a session start you will see **exactly one** of:
- ✅ `Escape CGEventTap created and enabled on the main run loop (trusted=true)` → tap is live; Escape should cancel.
- ❌ `CGEventTap NOT created: process is not Accessibility-trusted…` → grant Accessibility in **System Settings → Privacy & Security → Accessibility** and **relaunch** localvoxtral.
- ❌ `CGEvent.tapCreate returned nil although AXIsProcessTrusted()==true…` → relaunch the app (macOS can require a relaunch after granting Accessibility for the event-tap path).

While dictating, if you press Escape you should see `Escape consumed during active dictation; triggering cancel.` If the system ever disables the tap you will see `Escape CGEventTap disabled by system (type=N); re-enabled.`

### Exact manual re-test steps
1. Open Console.app, click Start, set the search filter to: `subsystem:com.localvoxtral category:Escape`
2. Ensure localvoxtral is granted **Accessibility** (System Settings → Privacy & Security → Accessibility). If you just toggled it, **quit and relaunch** localvoxtral.
3. Start dictation (either shortcut mode; either output mode).
4. Press **Escape** once. Expected: dictation stops immediately, no Escape reaches the focused app, and Console shows `Escape CGEventTap created and enabled…` (at start) + `Escape consumed during active dictation; triggering cancel.` (on press).
5. If instead you see a `NOT created` / `returned nil` line, that line names the exact reason — copy it back.

### Tests (full suite: 244 passed, 0 failures, Swift 6.2 warning-free)
- `EscapeCancelHandlerTests` — `start()` now records a deterministic outcome (DEBUG-only `lastStartOutcome`/`startCallCount`/`stopCallCount` seam) instead of silently no-oping; on an untrusted host (CI) it records `.notTrusted` before attempting `tapCreate`; `stop()` is idempotent. **Regression shown failing-before (silent return ⇒ outcome `none`) and passing-after** (two-run proof in the commit).
- `DictationViewModelOverlayLifecycleTests` — added disarm coverage across **every** teardown path (`stopDictation`, `cancelDictation`, `abortConnectingSession`) in **both** output modes (overlay buffer + live auto-paste), asserting `isDictatingRef` clears and `escapeCancelHandler.stop()` runs.

> The arming point (`startAudioCaptureAfterConnection`) needs a real microphone and can't be driven headlessly, so arming itself is verified by code-reading (single shared path) plus symmetric disarm tests rather than a live arming test. The diagnostic logs close the gap for the real device.

T0mSIlver:
Review finding (codex, P2, confirmed): cancelling while the session was still connecting leaked `wasCancelled` into the next session via `abortConnectingSession()` (which never reaches stopped-session cleanup) — the next overlay session would then silently skip promotion and commit, losing that dictation. Fixed by clearing the flag in `abortConnectingSession()`; regression test `testCancelWhileConnectingDoesNotLeakCancellationIntoNextSession` proven failing-before/passing-after. Suite: 245 tests, 0 failures.

T0mSIlver:
Implemented the Escape-cancel mechanism swap in c03c5f4.

Confirmed root cause from the maintainer log:

```
permission state: AXIsProcessTrusted=true inputMonitoring=denied
CGEvent.tapCreate returned nil although AXIsProcessTrusted()==true.
```

This PR now uses a Carbon `RegisterEventHotKey(kVK_Escape, 0)` registration scoped to the active dictation session instead of an active CGEventTap. That avoids requesting Input Monitoring for a single-key cancel path while preserving the intended behavior: Escape is consumed during dictation and untouched between sessions.

Manual test for the maintainer build:
- Start dictation, press Escape: dictation cancels.
- Between sessions, Escape behaves normally in the focused app.
- Nothing special to watch in Console anymore beyond the permission-state diagnostic line and the Carbon registration/failure log.

Proof:
- `PATH=/tmp/localvoxtral-ssh-bin:$PATH LV_BUILD_DIR=work/localvoxtral-pr16 ./scripts/remote-build.sh test`
- Result: `Executed 254 tests, with 0 failures`.
- Focused coverage includes `EscapeCancelHandlerTests` with double-start, idempotent stop/unregister, and registration-failure outcome assertions.

Issue #23, opened 2026-05-11
Allow postprocessing to revise already live transcribed text


axwnet:
Since text can likely be edited in place, I thought it might be possible to postprocess while live transcribing. This is the unique feature that sets this tool apart - instant transcription.

Example:

```sh
tr “dash” “-“
```

T0mSIlver:
Shipped in #39: the replacement dictionary now applies to Live Auto-Paste **while streaming** — corrections are retyped in place at word boundaries as text lands, rather than in a post-session sweep. Your example works as described: with `dash → -` in `replacement_dictionary.toml` and "Exact match replacements" enabled, saying "call me dash later" produces `call me - later` in the target field mid-dictation.

The one piece intentionally not covered is LLM-based polishing of live-typed text, which needs a different mechanism (whole-utterance rewrite after stop) with real per-app Accessibility risks — that's now tracked in #46. Closing this as delivered; please reopen if the streaming replacements don't cover your use case.

PR #35, opened 2026-07-02
fix: insert trailing-punctuation suffix when live deltas precede the final

T0mSIlver:
## What

In Live Auto-Paste + RealtimeAPI mode, partial deltas are typed live as they arrive. The final transcript frequently carries a trailing addition that never came as a partial delta (typically the final `.`). The `hadLiveDelta` guard in `handleFinalTranscriptEvent` skipped re-insertion entirely to avoid duplicating text, so that trailing addition never reached the target field — the field ended `sparisce` while the canonical event text said `sparisce.`.

Addresses the trailing-punctuation finding from #13's investigation (does **not** claim to fix #13 itself — the mid-word punctuation positioning analyzed there is a separate concern; the live path is append-only).

**Change:** when the final is a *pure extension* of what was already live-inserted (`final == liveInserted + suffix`, using the same preprocessing the live path applies on both sides), insert exactly the missing suffix. When the final *revises* earlier content (not a pure extension), keep today's behavior of inserting nothing — live mode cannot rewrite already-typed text (issue #23's territory).

The "already live-inserted" baseline is derived from the same state the `hadLiveDelta` guard reads (`pendingSegmentText`, with `livePartialText` fallback), captured before the segment reset, so it cannot drift from parallel bookkeeping. Both partials and finals go through `preprocessIncomingTranscriptChunk` (→ `FirstChunkPreprocessor`), so the prefix comparison is on equal footing.

### Note on a sanctioned test change

The pre-existing characterization test `testFinalTranscriptDoesNotDoubleInsertWhenLiveDeltasAlreadyInserted` asserted the **current (buggy)** behavior (`insertedChunks == ["sparisce"]` — the trailing `.` dropped). This PR changes intended behavior, so that assertion is updated to the correct one (`insertedChunks == ["sparisce", "."]`) and the test renamed. This is the sanctioned exception to the "never weaken a test" rule: the old assertion locked in the bug; the new one locks in the fix, and is stricter (it requires *more* correct output, not less). All other assertions in that test (`currentDictationEventText == "sparisce."`, `pendingSegmentText == ""`) are unchanged.

## Proof

- [ ] Realtime integration green — not re-run; this change does not touch the websocket/audio path (tier 1 runs in CI on the self-hosted runner)
- [x] Bug fix: regression test added — failing run before the fix **and** passing run after (both below)
- [ ] UI change: n/a (logic-only change in the live-insertion path; no UI surface changed)

### Regression test — FAILING run (before the fix)

`LV_BUILD_DIR=work/localvoxtral-punct ./scripts/remote-build.sh test --filter RealtimeAPILivePastePunctuationTests` — against current/buggy source with the new regression tests added:

```
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsMultiCharSuffixFromPureExtension]' started.
/Users/builder/work/localvoxtral-punct/Tests/localvoxtralTests/RealtimeAPILivePastePunctuationTests.swift:212: error: -[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsMultiCharSuffixFromPureExtension] : XCTAssertEqual failed: ("["you are", " right"]") is not equal to ("["you are", " right", ", right?"]")
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsMultiCharSuffixFromPureExtension]' failed (0.058 seconds).
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsTrailingPunctuationSuffixFromPureExtension]' started.
/Users/builder/work/localvoxtral-punct/Tests/localvoxtralTests/RealtimeAPILivePastePunctuationTests.swift:198: error: -[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsTrailingPunctuationSuffixFromPureExtension] : XCTAssertEqual failed: ("["sparisce"]") is not equal to ("["sparisce", "."]")
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsTrailingPunctuationSuffixFromPureExtension]' failed (0.002 seconds).
	 Executed 16 tests, with 2 failures (0 unexpected) in 0.091 (0.092) seconds
```

### Regression test — PASSING run (after the fix)

Same command after the fix:

```
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testEmptyFinalTranscriptWithNoLiveDeltasInsertsNothing]' passed (0.011 seconds).
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptIdenticalToLiveTextIsNoOp]' passed (0.002 seconds).
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsMultiCharSuffixFromPureExtension]' passed (0.002 seconds).
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptInsertsTrailingPunctuationSuffixFromPureExtension]' passed (0.001 seconds).
Test Case '-[localvoxtralTests.RealtimeAPILivePastePunctuationTests testFinalTranscriptThatRevisesLiveTextIsNotInserted]' passed (0.001 seconds).
	 Executed 16 tests, with 0 failures (0 unexpected) in 0.032 (0.033) seconds
```

The new cases cover: pure-extension trailing `.` (core fix), multi-char suffix `, right?`, NOT-pure-extension revision (`sparisce` → `sparisci.` stays dropped), empty final (no-op), and final identical to live text (no-op). Pure-function coverage for `livePasteExtensionSuffix` is added to `TextMergingAlgorithmsTests` (7 cases).

### Full unit suite — green after the fix

`LV_BUILD_DIR=work/localvoxtral-punct ./scripts/remote-build.sh test`:

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-07-03 01:05:18.600.
	 Executed 242 tests, with 0 failures (0 unexpected) in 1.340 (1.351) seconds
```

### Strict-concurrency — warning-free

Forced recompile of the two changed source files; no `warning:`/`error:` emitted:

```
[3/7] Compiling localvoxtral TextMergingAlgorithms.swift
[5/7] Compiling localvoxtral DictationViewModel+RealtimeEvents.swift
=== no warning:/error: lines ===
```

### DIFF
--- a/Sources/localvoxtral/DictationViewModel+RealtimeEvents.swift
+++ b/Sources/localvoxtral/DictationViewModel+RealtimeEvents.swift
@@ -137,6 +137,13 @@ extension DictationViewModel {
         let finalizedSegment = resolvedFinalizedSegment(from: processedText)
         let hadLiveDelta = !pendingSegmentText.trimmed.isEmpty
             || !livePartialText.trimmed.isEmpty
+        // Text already typed into the field by the live partial path. Derived
+        // from the same state the `hadLiveDelta` guard reads (accumulated
+        // pending text, with live-partial text as fallback) so it cannot drift
+        // from a parallel bookkeeping. Captured before the reset below.
+        let liveInsertedText = pendingSegmentText.trimmed.isEmpty
+            ? livePartialText
+            : pendingSegmentText
         guard !finalizedSegment.isEmpty else {
             livePartialText = ""
             pendingSegmentText = ""
@@ -154,8 +161,23 @@ extension DictationViewModel {
         pendingSegmentText = ""
         statusText = activeStatusText
 
-        if !hadLiveDelta, isLiveAutoPasteModeEnabled {
-            textInsertion.enqueueRealtimeInsertion(finalizedSegment)
+        if isLiveAutoPasteModeEnabled {
+            if !hadLiveDelta {
+                // No partials were typed live: insert the whole segment.
+                textInsertion.enqueueRealtimeInsertion(finalizedSegment)
+            } else if let liveSuffix = TextMergingAlgorithms.livePasteExtensionSuffix(
+                finalText: processedText,
+                liveInsertedText: liveInsertedText
+            ) {
+                // Partials were already typed live and the final is a pure
+                // extension of them (e.g. a trailing "." that only arrived in
+                // the final): insert only the missing suffix so the trailing
+                // addition reaches the field without duplicating earlier text.
+                // When the final revises earlier content the helper returns nil
+                // and nothing is inserted — live mode cannot rewrite already
+                // typed text.
+                textInsertion.enqueueRealtimeInsertion(liveSuffix)
+            }
             if let accessibilityError = textInsertion.lastAccessibilityError {
                 lastError = accessibilityError
             }
--- a/Sources/localvoxtral/TextMergingAlgorithms.swift
+++ b/Sources/localvoxtral/TextMergingAlgorithms.swift
@@ -121,11 +121,18 @@ enum TextMergingAlgorithms {
         finalText: String,
         liveInsertedText: String
     ) -> String? {
-        guard !liveInsertedText.isEmpty, finalText.hasPrefix(liveInsertedText) else {
+        // The finalized-transcript path trims surrounding whitespace, so mirror
+        // that here: a final of "hello. " must type "." — never a dangling
+        // space the transcript itself discards.
+        let normalizedFinal = finalText.trimmed
+        guard !liveInsertedText.isEmpty, normalizedFinal.hasPrefix(liveInsertedText) else {
             return nil
         }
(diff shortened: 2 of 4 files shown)


T0mSIlver:
Review finding (codex, P2, confirmed): a final with trailing whitespace (e.g. `"hello. "`) typed the whitespace into the field while the finalized-transcript path trims it — diverging field text from the transcript. Fixed by trimming the final before deriving the suffix; whitespace-only extensions are no-ops. Tests `testLivePasteExtensionSuffix_trailingWhitespaceInFinalIsNotTyped` / `_whitespaceOnlyExtensionReturnsNil` proven failing-before (typed `". "` and `".\n"`) / passing-after. Suite: 244 tests, 0 failures.

PR #39, opened 2026-07-03
Use streaming replacements for Live Auto-Paste

T0mSIlver:
## What

Reworks PR #39 from a stop-time Live Auto-Paste sweep into a streaming replacement corrector.

- Uses the existing “Exact match replacements” / `replacementDictionaryEnabled` toggle for Live Auto-Paste too. The sweep-only setting, Settings row, and persistence are removed.
- Loads the replacement dictionary at live session start only when that existing toggle is on; when it is off, live startup does not load the dictionary, capture a target, or install correction state.
- Adds `LiveReplacementCorrector`, which tracks the text the app successfully typed, scans only newly completed boundary tails, and uses a max-key-word-count lookback so multi-word dictionary keys can match as soon as the final word completes.
- Injects corrections in order through the live insertion pipeline: backspaces erase the matched text plus boundary characters by grapheme count, then Unicode keyboard typing inserts the replacement plus the preserved boundary.
- Keeps corrections keyboard-only. No AX field-value write, no selection paste fallback.
- Reuses the span/caret tracking only as a guard: if the target exposes a caret and it diverges, or if realtime insertion/correction reports failure, the corrector logs once at notice level and stands down for the rest of the session.
- Flushes the final unbounded word on live stop after pending/final suffix insertion has run and before cleanup/session persistence. Raw transcript/session records stay unchanged.

Deleted vs reused:

- Deleted: `LiveAutoPasteSweep`, post-session sweep execution, sweep outcome types, AX span rewrite / verification / selection-paste fallback, sweep-specific tests.
- Reused: `ReplacementDictionary` parsing/matching source data and the PR’s caret-span idea as a guard only.
- Replaced tests: the old sweep tests are deleted because the sweep feature is superseded; `DictationViewModelLiveReplacementCorrectorTests` now drives the production live event/insertion path with DEBUG insertion hooks including backspace capture.

## Proof

- [x] Unit suite green — `PATH=/tmp/lv-ssh:$PATH LV_BUILD_DIR=work/localvoxtral-dictsweep ./scripts/remote-build.sh test`
  - The temporary `PATH` prefix only provided an `ssh -F /dev/null` wrapper because this Linux environment has a broken system SSH include: `/etc/ssh/ssh_config.d/20-systemd-ssh-proxy.conf` has bad owner/permissions. The build/test still ran through `scripts/remote-build.sh` on the Mac host with the requested `LV_BUILD_DIR`.
  - Summary: `Test Suite 'Selected tests' passed ... Executed 260 tests, with 0 failures (0 unexpected)`.
- [x] New streaming-corrector coverage:
  - `DictationViewModelLiveReplacementCorrectorTests/testCorrectsWordCompletedAcrossDeltaBoundaries`
  - `testCorrectsMultiWordKeyWithLookbackWindow`
  - `testCorrectsFinalUnboundedWordOnStopFlush`
  - `testFinalTranscriptSuffixCanCompleteAndCorrectWordBeforeStop`
  - `testNoMatchWordsAreUntouched`
  - `testCorrectorDefersWhenCaretHasNotSettled`
  - `testCorrectorStandsDownAfterCaretSettleTimeout`
  - `testCorrectorWaitsForLaggedCaretAndThenCorrects`
  - `testDeltasArrivingDuringInFlightCorrectionAreBufferedUntilAfterCorrection`
  - `testFailedReplacementRetypesOriginalTextBeforeStandingDown`
  - `testBackspaceCountUsesGraphemeCountForEmojiMatch`
  - `testShorterReplacementPreservesBoundary`
  - `testBoundaryCharactersArePreserved`
  - `testReplacementDictionarySettingOffLeavesLivePathUntouchedAndDoesNotLoadDictionary`
- [ ] Realtime integration green — not run locally for this unit-scoped rework; CI/self-hosted runner should exercise the env-gated integration tier.
- [x] UI-affecting change: the rejected Live Auto-Paste post-processing row was removed; manual checks are listed below.

Manual checks before merge on a Mac build:

- Decisive on-device check for the async-lag class: with Live Auto-Paste and “Exact match replacements” enabled, speak a dictionary word in TextEdit and see it visibly correct immediately after its boundary is spoken. Unit tests simulate lagged caret reads, but only the real app proves target-run-loop CGEvent settling.
- Dictate a match split across speech chunks and a multi-word key; verify correction happens only when the completing boundary arrives.
- Stop after an unbounded final word that has a dictionary entry; verify it corrects during stop finalization.
- While dictating, manually type/click/edit in the target field; verify live correction stands down for the rest of that session and raw typed text is not rewritten.
- Repeat the streaming correction in Terminal, a web text field, and TextEdit.
- Disable “Exact match replacements” and repeat a matching phrase; verify there are no backspaces/corrections and live text remains raw.

Refs #23.


T0mSIlver:
Third-vendor adversarial review (Claude, plan-mode static trace): verdict SHIP — span-guard class traced safe end-to-end (UTF-16 accounting consistent; #35's suffix lands in the tracked span before the sweep; setting-off provably inert; Cmd+V fallback can't double-apply). One defense-in-depth hardening applied from finding M1: `replaceFieldSpan` now re-asserts the resolved element still holds the verified content before writing, closing a theoretical focus-change TOCTOU whose worst case was writing into the wrong field. 266 tests, 0 failures, CI green.

T0mSIlver:
Fix iteration pushed in `522f41b` (`Fix async live replacement caret guard`).

Finding -> fix:

- Major caret async race: correction is now a serialized async-settled pipeline step. On a boundary match, the corrector verifies the post-delta caret, polling every 10 ms for up to 15 attempts (150 ms total), posts backspaces, polls for the erased caret, then posts replacement + boundary. While a correction task is in flight, `flushPendingRealtimeInsertion()` leaves incoming deltas buffered in `pendingRealtimeInsertionText`; the finish path processes any pending final correction work and then flushes buffered deltas in order. If the caret is unavailable or never settles, the corrector stands down for the session.
- Minor partial failure erases text: if replacement posting fails after backspaces, the service now best-effort retypes the original erased text before standing down.
- Minor grapheme/UTF-16 deletion mismatch: documented where `backspaceCount` is computed. It remains grapheme-counted for native-field behavior; the next caret guard catches UTF-16-deleting targets and disables further corrections.

Regression evidence for the major:

- Fail-before: on `origin/agent/opencode/dict-sweep` (`ac83eae`) with only `testCorrectorWaitsForLaggedCaretAndThenCorrects` added, `PATH=/tmp/lv-ssh:$PATH LV_BUILD_DIR=work/localvoxtral-dictsweep ./scripts/remote-build.sh test` failed. The regression left `"voxtral "` instead of `"localvoxtral "`, events were only `[.type("voxtral ")]` instead of type/backspace/replacement, and `XCTAssertTrue(debugLiveReplacementCorrectorIsActive)` failed. Summary: `Executed 257 tests, with 3 failures`.
- Pass-after: same command on `522f41b` passed. Summary: `Test Suite 'Selected tests' passed ... Executed 260 tests, with 0 failures (0 unexpected)`.

New/updated tests include:

- `testCorrectorWaitsForLaggedCaretAndThenCorrects`
- `testDeltasArrivingDuringInFlightCorrectionAreBufferedUntilAfterCorrection`
- `testCorrectorStandsDownAfterCaretSettleTimeout`
- `testFailedReplacementRetypesOriginalTextBeforeStandingDown`
- existing instant-caret correction tests still pass unchanged behaviorally

PR body Proof/manual checks were updated. The decisive manual check now calls out speaking a dictionary word in TextEdit and seeing it visibly correct, because unit tests simulate lagged caret reads but cannot prove real target run-loop CGEvent settling.

T0mSIlver:
Fix pushed in `192c2f7` (`Enable dictionary settings in live mode`).

What changed:
- Split Text Processing settings into `Replacements` and `LLM Polishing`.
- Exact match replacement toggle, config folder button, and `replacement_dictionary.toml` notes are enabled in Live Auto-Paste.
- LLM polishing controls keep the Live Auto-Paste gate, with the unavailable explanation scoped to the LLM polishing rows.
- Updated dictionary copy to say replacements apply in both modes: streamed corrections in Live Auto-Paste, before commit in Overlay Buffer.

Testing:
- `LV_BUILD_DIR=work/localvoxtral-dictsweep ./scripts/remote-build.sh test`
- Remote command: `swift test --skip RealtimeAPIVLLMIntegrationTests`
- Result: `Executed 260 tests, with 0 failures (0 unexpected)`.

Test note:
- I did not add a SettingsStore regression because this stale gate was purely SwiftUI view-layer state. `SettingsStore.replacementDictionaryEnabled` already persists independently of `dictationOutputMode`; there was no store-level mode gate to exercise.

Manual UI re-check:
- Open Settings, set Dictation > Output mode to Live Auto-Paste.
- Open Text Processing.
- Confirm Exact match replacements and Open Config Folder are active.
- Confirm LLM polishing rows remain disabled/dimmed and show the Live Auto-Paste unavailable explanation.

T0mSIlver:
Round 3 update pushed in `970e532`.

Root cause for Part A:
- Live Auto-Paste did load the dictionary and did route deltas through `TextInsertionService.recordSuccessfulLiveInsertion`.
- The failure was the next gate: `beginLiveReplacementSession` treated `kAXSelectedTextRangeAttribute` as mandatory. If the target field could not expose a selected-text range at session start, `readCurrentCaretLocation(...)` returned nil and the corrector immediately stood down before any partial delta could be matched.
- That reproduces the device symptom: normal live keyboard insertion still works, overlay finalization still works, but streaming dictionary correction never posts. Web/Electron/custom AX-backed fields are the risky class because they can accept Unicode keyboard events while not exposing `kAXSelectedTextRangeAttribute`.

Fix:
- The caret reader is now an optional guard instead of an arming precondition. If the initial caret is available, the existing UTF-16 caret tracking/deferred settle path still protects corrections. If it is unavailable, the corrector stays armed and posts the immediate backspace/replacement sequence through the same keyboard path that inserted the live text.
- The dictionary is snapshotted per session and reused by live + overlay paths.
- Added sparse notice logs under `Log.corrector` / category `Corrector`:
  - `corrector armed entries=... rules=... caret_guard=on|off`
  - `corrector stand-down reason=...`
  - `corrector correction posted erased_chars=... replacement_chars=...`
  - `corrector settle timeout expected_utf16=...`

Maintainer capture command:

```bash
log stream --style compact --predicate 'subsystem == "com.localvoxtral" && category == "Corrector"'
```

Part B:
- Overlay display/commit buffers now run the same completed-boundary matching core via `LiveReplacementCorrector.completedBoundaryCorrectedText(...)`.
- The raw transcript/session text is not mutated while streaming. Finalize still applies the dictionary once to the raw text.
- Finalization suppresses another streaming pass over already processed text, which prevents chained rules like `foo -> bar`, `bar -> baz` from double-applying at commit.

Regression coverage added:
- `testCorrectorStillAppliesWhenInitialCaretIsUnavailable`
- `testOverlayStreamingReplacementUpdatesDisplayAtWordBoundary`
- `testOverlayFinalizeDoesNotDoubleApplyStreamingReplacementAndSavesRawRecord`

Proof:

```text
PATH=/tmp/ssh-no-global:$PATH LV_BUILD_DIR=work/localvoxtral-dictsweep ./scripts/remote-build.sh test
...
Test Suite 'Selected tests' passed
Executed 263 tests, with 0 failures (0 unexpected)
```

Manual re-checks requested on the device:
- Live Auto-Paste into the same real field that failed before: speak a dictionary key followed by a boundary, expect a `Corrector` armed line and then a `correction posted` line; the visible field should update live.
- Overlay Buffer: speak the same dictionary key followed by a boundary, expect the overlay text to update before stop; stop/commit should insert the replacement once, not a chained/double-applied value.

CI:

```text
gh pr checks 39 --watch
build-test pass 1m29s
```

PR #65, opened 2026-07-04
Polishing UX: launch self-heal, controls consolidated in Endpoints, honest replacements copy

T0mSIlver:
## What

Three owner requests from live #62 testing:

1. **Launch self-heal**: starting the app with LLM polishing already enabled (managed mode) now triggers the polishing warmup immediately — a missing/deleted model re-downloads with the inline progress bar instead of requiring the disable/re-enable dance. Shares `startPolishingWarmup()` with the toggle path; runs only in the real-app init (`startRuntimeServices`), so tests and the ui-smoke launch invariant (its defaults leave polishing off) hold. Warm cache costs ~a second ("Checking model…"); offline launches fail into the status row and heal on the next toggle/dictation.
2. **All polishing control in one place**: the toggle, overlay-only availability card + gating, managed status row (with the download bar) and external config fields now live under **Endpoints → Polishing**, ordered Mode → toggle → status/config. The Mode picker stays outside the availability gating (switchable in any output mode). Text Processing is down to Replacements + Shared Configuration.
3. **Replacements copy**: no longer claims replacements happen only "before commit in Overlay Buffer" — they apply during live corrections and overlay finalization (verified against both call sites).

## Proof

New tests: `testWarmUpPolishingAtLaunchIfNeededEnabledManagedWarmsUpPolishingOnly` / `...DisabledDoesNotWarmUp` / `...ExternalModeDoesNotWarmUp`.

```
Executed 437 tests, with 0 failures (0 unexpected)
```
(implementer codex worker run; PR CI re-verifies)

## Hand-verification owed (macOS, try-pr.sh)

- Quit app, delete `~/.cache/huggingface/hub/models--mlx-community--Qwen3.5-0.8B-8bit`, relaunch with polishing enabled → download bar appears without touching anything.
- Endpoints → Polishing: Mode → toggle → status/bar in managed; fields in external; availability gating in Live output mode leaves the Mode picker usable.
- Text Processing shows only Replacements + Shared Configuration; replacements copy reads correctly.

## Now the combined settings+onboarding PR (owner direction)

This branch now **contains PR #64** (wizard + General pane, both review waves) so one copy standard covers the whole settings/onboarding surface coherently. **Merge #64 first** — this PR then shrinks to its own delta (polishing UX + the copy pass). Suite on the combined tree: **494 tests, 0 failures**.

## Copy pass (Opus subagent) — before → after

| Where | Before | After |
|---|---|---|
| Endpoints → Polishing toggle | "LLM polishing" + subtitle "Send dictation text to an OpenAI-compatible chat completions server." | **"Enable"**, no subtitle |
| Endpoints → Polishing help | "Used only while LLM polishing is enabled." | **deleted** |
| Endpoints → Polishing (disabled state) | "Enable LLM polishing to start this backend." | **deleted** (toggle is directly above) |
| Availability card | "Unavailable in Live Auto-Paste output mode" / "LLM polishing runs before Overlay Buffer commits. Switch Dictation > Output mode…" | "Unavailable in Live Auto-Paste mode" / "Set Dictation > Output mode to Overlay Buffer to enable it." |
| Mode descriptions | "localvoxtral installs and runs voxmlx for dictation on this Mac." (and mlx-lm twin) | "Installs and runs voxmlx on this Mac." / "Installs and runs mlx-lm on this Mac." |
| Dictation → output-mode help | "Used when starting dictation from the menu bar. Keyboard triggers above select their mode directly." | "Keyboard triggers select their mode directly." |
| Auto-copy (now in General) | "Copy the finalized segment to the clipboard after dictation stops." | "Copies to the clipboard when dictation stops." |
| Replacements toggle subtitle | "Apply dictionary replacements during live correction and overlay finalization." | "Applied in both output modes." |
| Config-folder help | "Dictionary changes apply to live corrections and overlay finalization." | **deleted** |
| Prompt-file notes | "System prompt used for polishing." / "User prompt template used for polishing. Remove the {{replacement_dictionary}} placeholder if you do not want to send the dictionary to the LLM." | "System prompt for polishing." / "User prompt template for polishing. Remove {{replacement_dictionary}} to stop sending the dictionary to the LLM." |

Deliberately kept verbose: the onboarding wizard page subtitles (welcome value-prop, why each permission, what downloads) — onboarding is the one surface where explanation is the point; and the Live Auto-Paste caret-overwrite `.help` tooltip (real data-loss caveat).


T0mSIlver:
## Post-review delta (2026-07-05)

Four commits since the Opus review wave:

- **9d9ca63** — dropped the "Applied in both output modes." subtitle from the Exact match replacements toggle (owner request); the fact stays on-screen via the `replacement_dictionary.toml` file note, and the caret-overwrite `.help` tooltip is untouched.
- **f06564a** — both medium review findings fixed:
  1. `warmUpPolishingAtLaunchIfNeeded` now requires `dictationOutputMode == .overlayBuffer`, mirroring the dictation-time gate — a persisted enabled toggle no longer spawns mlx-lm on a Live Auto-Paste launch.
  2. `ui-smoke.sh` writes `settings.llm_polishing_enabled -bool false`, so the lazy-bootstrap launch invariant holds by construction instead of by whatever the runner's persisted defaults contain.
- **e55aac3** — README screenshots recaptured on this branch via `capture-assets.yml` (first successful run; the runner now has its Screen Recording grant). Adds `settings-general.png`, refreshes the three changed panes.
- **be60a82** — `release.yml`/`release.sh` learn `X.Y.Z-rc.N` prereleases dispatchable from a branch (full gate, pipeline-owned tag, invisible to `releases/latest`) so v0.7.2-rc.1 can be hand-tested via the real `curl | bash` install path before merge.

### Proof — regression test for finding 1

Without the gate (fix stashed):

```
Tests/localvoxtralTests/DictationViewModelFailFastUXTests.swift:663: error: -[localvoxtralTests.DictationViewModelFailFastUXTests testWarmUpPolishingAtLaunchIfNeededLiveAutoPasteDoesNotWarmUp] : XCTAssertTrue failed
Tests/localvoxtralTests/DictationViewModelFailFastUXTests.swift:664: error: ... : XCTAssertNil failed: "Task<(), Never>(_task: (Opaque Value))"
Executed 1 test, with 2 failures (0 unexpected) in 0.087 (0.087) seconds
```

With the fix, full suite on the build host:

```
Executed 495 tests, with 0 failures (0 unexpected) in 6.086 (6.114) seconds
```

### Flake observation (pre-existing, not from this branch's changes)

One full-suite run failed `testManagedStartupShowsPolishingModelDownloadProgress` ("Installing polishing backend..." instead of "Downloading polishing model (25%)..."); it passed in isolation and in an immediate full rerun, and CI was green on the identical commit. Smells like the PR #66 family of cross-test state leakage — noting it here so it's on the radar, not weakening anything.

PR #71, opened 2026-07-05
CI proof tooling: coverage summary, dSYM artifacts, on-demand ui-smoke, advisory format lint

T0mSIlver:
## What

- **Coverage visibility**: the unit suite runs with `--enable-code-coverage` and a new step prints the `llvm-cov report` summary over `Sources/` — PR Proof sections can cite real coverage of changed files. Visibility only, no threshold gate.
- **dSYM artifacts**: `package_app.sh` builds with `-Xswiftc -g` (does not change `-O`) and runs `dsymutil` → `dist/localvoxtral.dSYM`. `ci.yml` uploads it (`localvoxtral-dsym`, 30-day retention — field crashes surface days after the try-pr build); `release.yml` attaches `localvoxtral-<tag>.dSYM.zip` to releases.
- **Bug fix (found while proving the above)**: `remote-build.sh package` has been broken since #67 — in a tree without `.git` (rsync excludes it), `git describe` exits 128 and `set -euo pipefail` killed the script *silently* before the intended `0.0.0-dev` fallback could apply. Fixed with `|| true` inside the substitutions.
- **On-demand UI proof**: `ui-smoke.yml` now also triggers when the `needs-ui-smoke` label is added to a same-repo PR. Fork PRs can never reach the self-hosted runner (label event still checks `head.repo`).
- **Format config + advisory lint**: `.swift-format` tuned to the existing style (4-space indent, 120 cols) + `.editorconfig`. CI prints lint findings but stays green (swift-format only fails with `--strict`). Blocking enforcement needs a one-shot tree reformat first — measured at ~2,250 findings (mostly continuation-line indentation), which would conflict with in-flight PRs, so that's a separate owner-scheduled change.

## Proof

- **Packaging bug**, minimal repro (Linux, non-git dir): `set -euo pipefail; V="$(git describe --tags 2>/dev/null | sed 's/^v//')"` → `exit=128`, nothing printed. With `|| true` → `exit=0`, fallback applies.
- **Before fix**: `LV_BUILD_DIR=work/localvoxtral-ci-proof ./scripts/remote-build.sh package` → `exit=128` with zero remote output. **After fix**, same command:
  ```
  Packaged app: /Users/builder/work/localvoxtral-ci-proof/dist/localvoxtral.app
  Debug symbols: /Users/builder/work/localvoxtral-ci-proof/dist/localvoxtral.dSYM
  UUID: A30F4BDF-FA3A-351E-8F2B-50903CB10727 (arm64) .../DWARF/localvoxtral
  ```
  (exit=0 — real DWARF in the dSYM, not an empty shell)
- Workflow YAML validated; this PR's own CI run exercises the coverage + lint + dSYM-artifact steps end-to-end.
- Label trigger: adding `needs-ui-smoke` to this PR should start a UI Smoke run — will verify after opening.
- Lint baseline measured with the Swift 6.2 toolchain: 2,250 findings under the tuned config (1,941 Indentation) vs 23,775 under defaults.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.swift]
+indent_style = space
+indent_size = 4
+
+[*.{yml,yaml,json}]
+indent_style = space
+indent_size = 2
+
+[Makefile]
+indent_style = tab
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -2,10 +2,14 @@
 
 ## `ci.yml`
 
-Every PR and push to main: build, unit tests, live voxmlx integration tests,
-app packaging, launch smoke test, and an installable app artifact
-(`localvoxtral-app`, fetch with `scripts/try-pr.sh`). Same-repo branches run
-on the self-hosted Mac runner; fork PRs run on GitHub-hosted macOS.
+Every PR and push to main: build, advisory format lint (`.swift-format`;
+flips to `--strict` after the one-shot tree reformat), unit tests with a
+coverage summary (`llvm-cov report` over Sources — visibility for the PR
+Proof section, not a gate), live voxmlx integration tests, app packaging,
+launch smoke test, an installable app artifact (`localvoxtral-app`, fetch
+with `scripts/try-pr.sh`), and a `localvoxtral-dsym` artifact (30-day
+retention) for symbolicating field crashes. Same-repo branches run on the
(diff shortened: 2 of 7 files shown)


T0mSIlver:
The red `ui-smoke` check on this PR is **pre-existing breakage on main**, not introduced here — last night's scheduled nightly (on `d0185ef`, before this branch existed) failed with the identical signature, and this PR contains no Swift changes. Details and analysis: #72.

Silver lining: the `needs-ui-smoke` label trigger added by this PR worked end-to-end (label → same-repo guard → self-hosted run) and is what surfaced the silently-red nightly at PR time. The `build-test` check — which exercises this PR's coverage summary, advisory format lint, dSYM artifact, and the package_app.sh exit-128 fix — is green.

PR #82, opened 2026-07-07
Terminal-safe live dictation: target detection, hold-back replacements, newline/tab sanitization

T0mSIlver:
## What

Live Auto-Paste replacement-dictionary entries never applied reliably. The default (non-terminal) path typed text raw, then tried to **backspace and retype** each replacement, gated by reading the target's AX caret. But synthetic CGEvents reach the target *asynchronously*, so the synchronous caret read never matched immediately — every correction deferred waiting for the caret to settle, and a short dictation ended (tearing the deferred task down) before it did. Field logs show **zero** successful corrections across a whole session (crashlog run 28959033459, 2026-07-08): replacements silently did nothing in every non-terminal app. Terminals were worse still — a screen-grid caret can *never* match, so they timed out and stood down.

This PR makes live dictation replacement reliable everywhere by applying replacements **before** typing, for **every** target.

- **`TerminalTargetDetector`** — per-session verdict (bundle-ID allowlist: Terminal.app, iTerm2, Ghostty, Warp incl. channel variants, WezTerm, kitty, Alacritty, Hyper, Tabby, Rio; user `terminal_apps.toml` allowlist; AX-writability probe fallback for unknown apps). Captured in `beginDictationSession` *before* the socket opens (focus can drift during connect). Transient AX failures read as "unavailable", not terminal-like.
- **`LiveHoldBackReplacementStream`** — the sole live-replacement strategy. Text is held back to the trailing partial word + (max rule word count − 1) complete words, corrected, then released. **No caret reads, no backspaces, ever.** Embeds `LiveReplacementCorrector` for matching (no duplicated rule logic).
- **Strategy selection** at live session start: terminal-like → hold-back **with** newline/tab sanitization (a typed `\n` submits an agent prompt, a typed `\t` triggers shell completion — both collapse to a space); non-terminal with rules → hold-back **without** sanitization (a regular editor keeps its newlines); non-terminal with no rules → type directly (zero hold-back delay). Terminal targets arm sanitization even with the dictionary disabled.
- **The guarded backspace corrector is deleted** — along with the deferred-correction task, caret-settle polling, adaptive guarded→hold-back conversion, unguarded blind-backspace path, and the now-dead corrector stand-down / erased-text / backspace-count API. **Net −1228 lines.**
- **Secure Keyboard Entry warning** — one-line popover warning at session start when `IsSecureEventInputEnabled()` (blocks synthetic keystrokes; auto-enabled by Ghostty around password prompts), token-scoped so it never clobbers higher-priority errors and clears at session end.
- Strategy/verdict decisions logged per session (`Log.corrector`, `Log.target`).

Both the original terminal work and this hold-back-everywhere change were independently reviewed (Codex, GPT-5.5 high); all findings fixed.

### Tradeoff (accepted — already true for terminals)

Hold-back can't type a word until it sees the boundary proving it can't be part of a rule match, so the trailing partial word — or up to `maxRuleWordCount` words for a multi-word rule — appears one boundary later, flushed on the next word or on stop. Imperceptible next to "no replacement at all", and it's what terminals already did.

## Proof

Field evidence, guarded path, before this change (crashlog run 28959033459 — every non-terminal session armed then produced **zero** `correction posted`, only `teardown drained` / `restored erased text`):
```
18:21:36 corrector armed strategy=guarded-corrector entries=1 rules=1 caret_guard=on
18:21:39 corrector teardown drained queued final flush          # no "correction posted"
18:22:04 corrector armed strategy=guarded-corrector ...
18:22:08 corrector teardown restored erased text chars=9        # correction abandoned
```
(`grep -c "correction posted"` across the whole run: **0**.)

- [x] **Regression — the exact field bug** (final-word-only replacement, no trailing whitespace) now applied on stop, at both the service and view-model level:
```
TextInsertionServiceLiveStrategyTests testFinalWordOnlyReplacementIsAppliedAtStopNonTerminal passed (0.002s)
DictationViewModelLiveReplacementCorrectorTests testFinalWordOnlyReplacementIsAppliedAtStop passed (0.004s)
```
Fail-before is documented by the (now-deleted) `testCorrectorStandsDownAfterCaretSettleTimeout`, which asserted the field stayed **uncorrected** ("voxtral ") once the caret diverged — the drop, encoded as the old expected behavior.
- [x] **Invariant** — after stop, emitted text == dictionary applied to the full raw transcript (no dropped tail, no fresh-empty-stream bug), asserted for terminal and non-terminal:
```
testStopEmittedTextEqualsDictionaryAppliedToFullRawText passed (0.003s)
testEndSessionFlushesHeldTailWithoutExplicitFinalFlush passed (0.002s)
```
- [x] Stream matrix (pre-existing `LiveHoldBackReplacementStreamTests`, 26 tests): multi-word rule spanning ingest chunks, punctuation boundary, emoji grapheme, newline/tab sanitization on/off, replacement+sanitization compose.
- [x] Unit suite green — `./scripts/remote-build.sh` (578 tests, 2 env-gated skips, **0 failures, 0 warnings**):
```
Executed 578 tests, with 2 tests skipped and 0 failures (0 unexpected) in 6.185 (6.219) seconds
```
- [x] Hand-verification — done twice over: owner field test (2026-07-08, all scenarios, "great") and the automated E2E probe ([run 28974870442](https://github.com/T0mSIlver/localvoxtral/actions/runs/28974870442), TextEdit + Ghostty, real TTS audio + hold gesture — see the probe comment below):
  - **Regular editor (e.g. TextEdit)**: the replacement lands (no backspace flicker); the last word appears on the next word/stop. Log shows `strategy=holdback newline_sanitize=off`.
  - **Ghostty / Claude Code**: replacement lands, newlines collapse to spaces. Log shows `strategy=holdback reason=terminal`.
  - **cmux** (writable AX): now in the **built-in allowlist** (owner request after field-testing) — zero-config sessions get full terminal treatment: replacements AND newline/tab sanitization, log `strategy=holdback reason=terminal newline_sanitize=on`. The user allowlist (`terminal_apps.toml`) remains for other undetectable terminal hosts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Open **Settings** from the menu bar popover:
 - **General** — permission status for Microphone and Accessibility (with grant buttons), copy-final-segment toggle, and Re-run Setup
 - **Endpoints** — Dictation and Polishing sections; each switches independently between `Managed local` (status row with inline install/download progress) and `External URL` (endpoint URL, model name, API key)
 - **Dictation** — the trigger (single modifier key with tap/hold gestures, or per-mode keyboard shortcuts with `Toggle` / `Push to Talk`) and the menu-bar output mode
-- **Text Processing** — exact-match replacements toggle and the shared config folder (`replacement_dictionary.toml`, `llm_system_prompt.toml`, `llm_user_prompt.toml`)
+- **Text Processing** — exact-match replacements toggle and the shared config folder (`replacement_dictionary.toml`, `llm_system_prompt.toml`, `llm_user_prompt.toml`, `terminal_apps.toml` — extra apps to treat as terminals, e.g. ones that embed one)
 - **About** — version, link to this repository, and Export Diagnostics (writes a redacted local report to the Desktop)
 
 The shared config directory lives at `~/Library/Application Support/localvoxtral/config`.
--- a/Sources/localvoxtral/AppConfigStore.swift
+++ b/Sources/localvoxtral/AppConfigStore.swift
@@ -4,6 +4,7 @@ protocol AppConfigServing {
     func configDirectoryURL() -> URL
     func loadReplacementDictionary() -> ReplacementDictionary
     func loadLLMPromptTemplates() -> LLMPromptTemplates
+    func loadTerminalAppBundleIDs() -> [String]
 }
 
 struct ReplacementEntry: Equatable, Sendable {
@@ -329,6 +330,7 @@ struct AppConfigStore: AppConfigServing {
         case replacementDictionary
         case llmSystemPrompt
         case llmUserPrompt
+        case terminalApps
 
         var fileName: String {
             switch self {
@@ -338,6 +340,8 @@ struct AppConfigStore: AppConfigServing {
                 return "llm_system_prompt.toml"
             case .llmUserPrompt:
                 return "llm_user_prompt.toml"
+            case .terminalApps:
+                return "terminal_apps.toml"
             }
         }
 
@@ -408,6 +412,25 @@ struct AppConfigStore: AppConfigServing {
             userPrompt = defaultTemplates.userContent
         }
         return LLMPromptTemplates(systemContent: systemPrompt, userContent: userPrompt)
+    }
+
+    /// User-listed bundle IDs to treat as terminals, on top of
+    /// `TerminalTargetDetector`'s built-in allowlist. Any read/parse failure
+    /// falls back to an empty list (the built-in detection still applies).
+    func loadTerminalAppBundleIDs() -> [String] {
+        let file = ConfigFile.terminalApps
+        let url = userConfigURL(for: file)
+
+        do {
+            ensureConfigFilesExist(at: resolvedConfigDirectoryURL())
+            let data = try Data(contentsOf: url)
+            return try Self.parseTerminalApps(data: data, fileName: file.fileName)
+        } catch {
+            Log.config.error(
+                "Terminal apps config fallback to empty list: \(error.localizedDescription, privacy: .public)"
+            )
+            return []
+        }
     }
 
     private func loadPromptContent(file: ConfigFile, fallback: String) -> String {
@@ -637,6 +660,62 @@ struct AppConfigStore: AppConfigServing {
 
         try finalizeCurrentEntry()
         return ReplacementDictionary(entries: entries)
+    }
+
+    private static func parseTerminalApps(
+        data: Data,
+        fileName: String
+    ) throws -> [String] {
+        guard let text = String(data: data, encoding: .utf8) else {
+            throw AppConfigError.invalidFile(fileName: fileName, reason: "File is not valid UTF-8.")
+        }
+
+        let normalizedText = text.replacingOccurrences(of: "\r\n", with: "\n")
+        let rawLines = normalizedText.components(separatedBy: "\n")
+
+        var bundleIDs: [String]?
+        var lineIndex = 0
+
+        while lineIndex < rawLines.count {
+            let line = uncommented(rawLines[lineIndex]).trimmed
+            lineIndex += 1
+
+            guard !line.isEmpty else { continue }
+            guard bundleIDs == nil else {
+                throw AppConfigError.invalidFile(
+                    fileName: fileName,
+                    reason: "Only a single bundle_ids assignment is supported."
+                )
+            }
+
+            let components = line.split(separator: "=", maxSplits: 1, omittingEmptySubsequences: false)
+            guard components.count == 2, String(components[0]).trimmed == "bundle_ids" else {
+                throw AppConfigError.invalidFile(
+                    fileName: fileName,
+                    reason: "Expected `bundle_ids = [...]` near `\(line)`."
+                )
+            }
+
+            var value = String(components[1]).trimmed
+            while !hasBalancedSquareBrackets(in: value) {
+                guard lineIndex < rawLines.count else {
+                    throw AppConfigError.invalidFile(
+                        fileName: fileName,
+                        reason: "Unterminated bundle_ids array."
+                    )
+                }
+                value += "\n" + uncommented(rawLines[lineIndex]).trimmed
+                lineIndex += 1
+            }
+
+            bundleIDs = try parseStringArray(
+                value,
+                fileName: fileName,
+                fieldName: "bundle_ids"
+            ).map { $0.trimmed }.filter { !$0.isEmpty }
+        }
+
+        return bundleIDs ?? []
     }
 
     private static func parsePromptTemplate(
--- a/Sources/localvoxtral/DictationViewModel+Session.swift
+++ b/Sources/localvoxtral/DictationViewModel+Session.swift
@@ -258,6 +258,11 @@ extension DictationViewModel {
             ? overlayBufferCoordinator.resolveAnchorNow()
             : nil
 
+        // Same timing rationale as the anchor: sample the terminal-like
+        // verdict and Secure Keyboard Entry state while the app the user
+        // started dictation in is still frontmost, not after connect.
+        captureSessionTargetVerdict()
+
         audioChunkBuffer.clear()
         livePartialText = ""
         pendingSegmentText = ""
@@ -303,7 +308,7 @@ extension DictationViewModel {
             isConnectingRealtimeSession = false
             isDictating = true
             escapeCancelHandler.start()
-            refreshSessionTargetVerdictAtSessionStart()
+            applyPreCapturedSessionTargetVerdict()
             statusText = "Listening..."
             restartAudioSendTask()
             restartCommitTask()
@@ -767,6 +772,11 @@ extension DictationViewModel {
         }
 
         if currentErrorToken == .websocketReceiveFailed {
+            lastError = nil
+        }
+        // The Secure Keyboard Entry warning describes state sampled at session
+        // start; a finished session must not leave it wedged in the popover.
+        if currentErrorToken == .secureKeyboardEntryActive {
             lastError = nil
         }
         firstChunkPreprocessor.reset()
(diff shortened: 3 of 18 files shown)


T0mSIlver:
## Field-testing round 1 (owner, try-pr) → user-extensible terminal allowlist

Hand-testing found live replacements dead in **cmux** (`com.cmuxterm.app`): it hosts a terminal but its focused AX element reports a *writable* value, so detection honestly classifies it as an ordinary text field and the guarded corrector dead-ends on the grid caret. Log evidence:

```
Session target verdict: terminalLike=false reason=ax-probe-writable bundle=com.cmuxterm.app
corrector armed strategy=guarded-corrector entries=1 rules=1 caret_guard=on
```

No heuristic can see through an app that claims a writable text field, so `570d60b` adds **`terminal_apps.toml`** to the config folder (bootstrapped template): user-listed bundle IDs are treated as terminals with a distinct `user-bundle-match` verdict reason. Parse failures fall back loudly to the built-in list.

Proof (full suite, integrated):
```
Executed 582 tests, with 2 tests skipped and 0 failures (0 unexpected) in 6.116s
```
New tests: `testTerminalAppsConfigParsesBundleIDs`, `testInvalidTerminalAppsConfigFallsBackToEmptyList`, `testTerminalAppsConfigRejectsUnsupportedKeys`, `testUserBundleIDIsTerminalLikeWithoutProbing`, `testUserBundleIDDoesNotOverrideBuiltInReason`, `testCaptureUsesUserTerminalAppsFromConfigStore`, and the end-to-end cmux repro `testUserAllowlistedBundleSelectsHoldBack` (writable AX probe + config entry → hold-back, zero backspaces).

Codex review of the commit: 1 finding (test fixtures using the production `AppConfigStore` would now touch the real config dir) — fixed in the follow-up commit with hermetic mocks.

Still under investigation from the same field-testing round: non-ASCII (guillemet) garbling in Ghostty live typing — isolated to the CGEvent→Ghostty boundary, probe with the owner pending. Not a regression of the replacement logic (`strategy=holdback reason=terminal` selected correctly, replacement applied).

T0mSIlver:
## Adaptive hold-back conversion (field round 2)

Undetected terminal-hosts (cmux without an allowlist entry) used to dead-end: guarded corrector arms → grid caret diverges → stand-down → replacements silently dead. Now a **caret-related stand-down converts the session to the hold-back stream** for all future text; keyboard-posting failures still truly stand down. Codex review (2 findings) fixed: erased text is restored before any post-backspace conversion, and session teardown drains text queued behind an in-flight correction instead of dropping it — the latter was a **pre-existing** stop-inside-settle-window loss bug, now fixed and pinned.

Proof:
```
FAIL before: ("voxtral voxtral ") is not equal to ("voxtral localvoxtral ")   [conversion regression]
FAIL before: ("") is not equal to ("voxtral ") - the erased text must be restored before converting
PASS after: all of the above
Full suite: Executed 592 tests, with 2 tests skipped and 0 failures (0 unexpected)
```
Key tests: `testCaretDivergenceConvertsToHoldBackAndKeepsApplyingReplacements`, `testPostBackspaceCaretLossRestoresErasedTextBeforeConverting`, `testSessionStopDuringInFlightCorrectionDrainsQueuedTextInsteadOfDroppingIt`, `testKeyboardFailureStandDownDoesNotConvertToHoldBack`.

## Ghostty guillemet garbling: probe status

Scripted probe (4 runs on the self-hosted GUI runner, `ghostty-typing-probe.yml` on the probe branch): Ghostty 1.3.1 renders localvoxtral's exact synthetic event shapes **clean** in every synthetic condition — 20-unit chunks, per-character, synthetic modifier emulation. Remaining suspect: the physically held Right-Command of the hold gesture (unreproducible synthetically; voice-assisted variant pending). Not a PR-82 regression of replacement logic — strategy selection and replacements behave correctly in the field logs.

T0mSIlver:
### Why the guarded backspace corrector was removed (not just gated off)

The guarded corrector wasn't merely flaky — it's structurally unable to do its job, so keeping it as a fallback would only preserve a path we don't trust.

**The mechanism can't win its own race.** It typed transcript text raw, then applied each replacement by posting backspaces + retyping, gated by reading the target app's AX caret to confirm the erase landed where expected. But synthetic `CGEvent`s are delivered to the target app *asynchronously* — the app hasn't processed the keystrokes when we read the caret synchronously right after posting them. So the caret never matches on the first read, and every correction takes the *deferred* path that polls for the caret to settle. A short dictation (a few seconds) ends before it settles; session stop tears the deferred task down; the correction is dropped. This is inherent to "type first, verify with a caret round-trip, then fix" — not a bug you patch, it's the wrong shape.

**Field evidence — zero successful corrections, ever** (crashlog run `28959033459`, a real build of this branch, dictating into non-terminal apps):
```
18:21:36 corrector armed strategy=guarded-corrector entries=1 rules=1 caret_guard=on
18:21:39 corrector teardown drained queued final flush          ← no "correction posted"
18:22:04 corrector armed strategy=guarded-corrector ...
18:22:08 corrector teardown restored erased text chars=9        ← correction abandoned, text unwound
```
`grep -c "correction posted"` across the entire session: **0**. Every non-terminal session armed, deferred, and dropped its replacement. Replacements silently did nothing in every regular app — which is exactly the user report that opened this.

**Hold-back has no such dependency.** It applies replacements to the string *before* typing, so there's no already-typed text to erase, no caret to read, no race to lose. It's the only path that produced replacements in the field (terminals already used it), and it's insertion-path agnostic (works whether text lands via keyboard events or AX).

**Why delete rather than keep behind a flag.** The only thing the guarded path offered was zero-latency in-place correction *when the caret happens to cooperate* — but the field data says it never does within a real dictation. Keeping ~400 lines of caret-guard/deferred-task/backspace/convert machinery as a disabled, now-untested alternate default is precisely the distrusted dead weight this codebase avoids. The hold-back stream already carries the full test matrix (multi-word-across-chunks, punctuation boundaries, emoji graphemes, newline/tab sanitization, failed-release retry), so the "delete once coverage is strong" bar is met.

The independent Codex review (GPT-5.5 high) reached the same conclusion: the deferred-then-teardown loss is inherent, and universal hold-back is the right call. Its one caveat — that a unit test mocking the caret to *match* would hide the bug and "prove the wrong thing" — is why the fail-before here is the field log above and the retired `testCorrectorStandsDownAfterCaretSettleTimeout` (which asserted the field stayed uncorrected once the caret diverged), not a green unit test on the old path.

**Accepted tradeoff:** hold-back can't type a word until it sees the boundary proving it can't be part of a match, so the trailing word (or up to `maxRuleWordCount` words) appears one boundary later — flushed on the next word or on stop. Imperceptible next to "no replacement at all," and it's what terminals already did.

T0mSIlver:
## Hand-verification: automated E2E probe — PASS (TextEdit + Ghostty)

The pending hand-verification is now covered by an automated probe (branch `probe/live-replacement-e2e`, workflow **Live Replacement E2E Probe**): it downloads this PR's CI artifact, stages a `banana → pineapple` `replacement_dictionary.toml` entry, pins the app mic to the BlackHole loopback, dictates via TTS with the real Right-Command **hold** gesture, and reads the results back (TextEdit via AppleScript, Ghostty via its AX tree) plus the per-session strategy lines from the unified log.

Result ([run 28974870442](https://github.com/T0mSIlver/localvoxtral/actions/runs/28974870442)):

```
TextEdit document: 'Please add me the ripe pineapple.'   ← spoken: "Please hand me the ripe banana"
PROBE-RESULT A-TEXTEDIT: PASS (replacement applied)
PROBE-RESULT B-GHOSTTY: PASS (replacement applied)
PROBE-RESULT LOG-NONTERMINAL: PASS (non-terminal hold-back armed, sanitization off)
PROBE-RESULT LOG-TERMINAL: PASS (terminal hold-back armed via allowlist)
PROBE-RESULT LOG-NO-GUARDED: PASS (guarded corrector never armed)
PROBE-RESULT OVERALL: PASS
```

- **TextEdit** covers the exact field regression shape: the sentence *ends* on the dictionary word, so the replacement is applied on the stop flush (the case the guarded corrector dropped 100% of the time). Log: `strategy=holdback … newline_sanitize=off`.
- **Ghostty** replacement landed at the prompt line; log: `strategy=holdback reason=terminal … newline_sanitize=on`.
- No `strategy=guarded-corrector` line anywhere in the session — the deleted path is really gone.

Still manual: the **cmux** check (and note the PR-body bullet was corrected — cmux resolves through the *non-terminal* hold-back path zero-config; sanitization there still needs the `terminal_apps.toml` entry).

⚠️ Verification side-discovery, filed as #87: the CI artifact **crashes at launch on any machine without the builder's `.build` directory** — `package_app.sh`'s resource-bundle patch is silently reverted by its own rebuild under the current toolchain, and CI's launch smoke + try-pr both run where the absolute `.build` fallback path exists, masking it. The probe passes because its packaging-audit step happens to recreate that path in the runner workspace. #87 must be fixed before shipping this (or any) build to another machine.

T0mSIlver:
Fresh independent review of the full PR diff vs main (Codex, GPT-5.5 high, `codex exec review`): **no actionable regressions found**. Also merged main in (picks up #88 — without it every CI artifact of this PR crashes at launch outside the runner workspace, which would have sunk any try-pr hand-test on a cold workspace).

T0mSIlver:
## cmux → built-in allowlist (owner request after field-testing)

Zero-config cmux sessions now get **full terminal treatment**: hold-back replacements AND newline/tab sanitization (`strategy=holdback reason=terminal newline_sanitize=on`). Previously zero-config cmux got replacements only, via the non-terminal path.

Proof:
- Fail-before: the pre-change suite pinned the old behavior — `testCaptureUsesUserTerminalAppsFromConfigStore` asserted unconfigured cmux stays non-terminal, and failed the moment cmux went built-in (`XCTAssertFalse failed`, run before the test updates in `1192ed4`).
- New end-to-end pin: `testCmuxIsTerminalLikeViaBuiltInAllowlist` — replacement + newline sanitization with zero config.
- User-allowlist path still exercised via a non-built-in stand-in bundle (`com.example.myterminal`); `terminal_apps.toml` template example updated.
- Suite: `Executed 579 tests, with 2 tests skipped and 0 failures`; CI green on `1192ed4`; Codex (GPT-5.5 high) review of the commit: no findings.

Worth a 30-second re-test in cmux: dictation should now collapse newlines (no accidental prompt submits) — log shows `newline_sanitize=on`.

PR #84, opened 2026-07-07
Bundled MLX Swift polishing engine replaces the managed mlx-lm wheel

T0mSIlver:
## What

**Replaces the managed mlx-lm Python wheel with a bundled MLX Swift polishing engine** (`localvoxtral-polishd`).

Why: upstream mlx-lm is effectively unmaintained (no release since 2026-04-22, 263 open PRs, the lead maintainer left Apple in February) and every fix we need lives in our fork wheel. Apple's maintenance attention moved to MLX Swift (`ml-explore/mlx-swift-lm`, active releases, Qwen3.5 support). This PR moves polishing onto it and deletes the Python/uv/wheel layer for that backend.

**New `PolishHelper/` package → `localvoxtral-polishd`**: MLXLLM inference + a minimal loopback OpenAI chat/completions server (`/health`, `/v1/chat/completions`) + kqueue parent-PID watchdog. Ships inside the .app (Contents/MacOS), supervised by the existing `BackendProcessSupervisor` on port 8472, same readiness contract as `mlx_lm.server` (`/health` answers only once the model is loaded). Deterministic generation (seeded sampling) so identical requests give identical output, like the old engine's within-state behavior.

**Deliberately a helper process, not in-process inference:** MLX Metal failures abort the host app uncatchably (ml-explore/mlx#3390 — C++ exception thrown inside a libdispatch completion queue), the same failure class as our PR #60 FileHandle crash. Out-of-process keeps a GPU OOM from killing dictation mid-sentence, and keeps the memory contract trivial: every existing stop trigger (polishing disabled, switch to External URL, overlay trigger removed, app quit) still frees all model memory by killing the process.

**Root app:** `ManagedBackendSpec` gains `installKind` (`uvWheel` | `bundledExecutable`). The polishing spec is now bundled — no uv/wheel install step, executable resolves next to the app binary, readiness timeout 1800s → 300s (weights are pre-downloaded by `prepareModel`; the helper never downloads, missing model is a hard actionable error). ensureReady single-flight, stop triggers, status flow, Settings/onboarding/diagnostics seams: unchanged. The `LLMPolishingServicing` HTTP seam is untouched, so the overlay-lifecycle suite runs as-is. Fork patches accounted for: transformers pin → gone (no Python); parent-PID watchdog → native kqueue; prompt caching → deferred follow-up (below).

**Build & packaging:**
- `PolishHelper/` is a separate SwiftPM package so the root `swift build`/`swift test` loop never compiles the MLX C++ core. Helper unit tests are Metal-free: `./scripts/remote-build.sh test --package-path PolishHelper`.
- The shippable helper MUST be built by xcodebuild (SwiftPM CLI cannot compile mlx-swift's Metal kernels); `package_app.sh` owns that lane and bundles `mlx-swift_Cmlx.bundle` (metallib) + tokenizer resources into Contents/Resources, plus a helper dSYM in dist/.
- One-time build-host provision: `xcodebuild -downloadComponent MetalToolchain` (~700 MB; **done on the build host 2026-07-07**, runbook added to `scripts/mac/README.md`). Gotcha: `xcrun --find metal` succeeds even without the component — only invoking `metal` proves it; the catalog fetch fails transiently, retry.
- Deps pinned via committed `PolishHelper/Package.resolved` (mlx-swift-lm 3.31.4 exact, mlx-swift 0.31.4, swift-transformers 1.3.3). All MIT/Apache-2.0, no copyleft.
- CI: helper unit tests + live-model integration on self-hosted lanes; fork PRs skip the helper (`LOCALVOXTRAL_SKIP_POLISHD=1`) to avoid the MetalToolchain download + cold MLX C++ build on hosted runners.
- New lane: `./scripts/remote-build.sh integration-polishd` — spawns the packaged helper with the real pinned model and holds it to the shared eval baseline (corpus factored into `LLMPolishEvalSupport`, used unchanged by `LLMPolishPromptEvalTests`).

## Proof

  ```
  Executed 523 tests, with 2 tests skipped and 0 failures (0 unexpected) in 6.119 (6.149) seconds
  ```
- [x] Helper unit suite green — `./scripts/remote-build.sh test --package-path PolishHelper`
  ```
  Executed 24 tests, with 0 failures (0 unexpected) in 0.112 (0.114) seconds
  ```
- [x] Packaging with bundled helper — `./scripts/remote-build.sh package`
  ```
  Contents/MacOS:
    localvoxtral
    localvoxtral-polishd
  dist/localvoxtral.app: valid on disk
  dist/localvoxtral.app: satisfies its Designated Requirement
  ```
- [x] Helper integration + eval baseline — `./scripts/remote-build.sh integration-polishd`
  ```
  Test Case 'testHelperExitsWhenParentPIDDies' passed (2.172 seconds).
  Test Case 'testHelperMatchesPolishEvalBaselineThroughProductionRequestPath' passed (11.585 seconds).
  Executed 2 tests, with 0 failures (0 unexpected) in 13.757 (13.757) seconds

  == polishd (bundled MLX Swift helper) eval (model: mlx-community/Qwen3.5-0.8B-8bit) ==
  required: 7/7, known-hard passing: 3/7
  ```
  Same-day reference run against the mlx-lm eval service (`./scripts/remote-build.sh eval-llm`, port 8080):
  ```
  == LLM polish prompt eval (model: mlx-community/Qwen3.5-0.8B-8bit, endpoint: http://127.0.0.1:8080/v1/chat/completions) ==
  required: 7/7, known-hard passing: 3/7
  ```
  Identical scores (10/14); the known-hard mix differs slightly (helper passes en-exclamation-extra-space, mlx-lm passes fr-semicolon-missing-space) — the corpus documents these as state-dependent wobblers, which is why only required cases are asserted.
- [x] Bug fix regression (parent-PID tether): the first live run caught the watchdog being deinit'd immediately (`_ =` discard cancels the kqueue source). Failing before:
  ```
  error: testHelperExitsWhenParentPIDDies : Asynchronous wait failed: Exceeded timeout of 30 seconds,
  with unfulfilled expectations: "helper exited after parent death".  (81.826 seconds)
  ```
  Passing after the retain fix:
  ```
  Test Case 'testHelperExitsWhenParentPIDDies' passed (2.172 seconds).
  ```
- [x] Realtime integration untouched by this PR (voxmlx path unchanged); CI runs it as tier 1 on this branch.
- [x] UI change: Settings copy only ("Runs the bundled polishing engine on this Mac.", status row shows "Polishing engine"); no group-structure changes. Not yet hand-verified — try-pr.sh after CI builds the artifact.

## Deferred follow-ups

- Prompt-cache reuse in the helper (the fork's ~50% prefill win). `ChatSession` has first-class cache save/trim, but Qwen3.5 is hybrid linear-attention (CacheList) — verify trim/persistence before relying on it. Note the full 14-case eval runs in ~12s including model load, ~0.5s/request warm, so latency is already acceptable.
- Mechanical rename of the `mlxLM*` Swift symbols (kept, to keep this diff reviewable).
- Cleanup of the orphaned uv-installed mlx-lm tool on existing user Macs.
- Retire `.github/workflows/mlx-lm-test.yml` (mlx-lm server harness).
- `com.localvoxtral.mlxlm` eval service stays as the prompt-eval reference endpoint (runbook note added).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,13 +81,34 @@ jobs:
           VLLM_REALTIME_TEST_MODEL: T0mSIlver/Voxtral-Mini-4B-Realtime-2602-MLX-4bit
         run: swift test --filter RealtimeAPIVLLMIntegrationTests
 
+      - name: Polishing helper unit tests (PolishHelper package)
+        # Self-hosted only: hosted runners would pay a cold Cmlx C++ compile
+        # every fork PR. These tests are Metal-free (HTTP layer, router,
+        # cache locator, watchdog); the Metal path is covered below.
+        if: runner.environment == 'self-hosted'
+        run: swift test --package-path PolishHelper
+
       - name: Package app bundle
         # The self-hosted runner's keychain holds the stable localvoxtral-dev
         # signing cert, keeping TCC grants valid across builds; anywhere the
         # identity is absent (hosted runners), packaging falls back to ad-hoc.
+        # Hosted runners skip the bundled polishing helper (no MetalToolchain
+        # component + cold MLX C++ build would dominate fork-PR CI); every
+        # self-hosted lane (same-repo PRs, main, releases) builds it.
         env:
           LOCALVOXTRAL_CODESIGN_IDENTITY: localvoxtral-dev
+          LOCALVOXTRAL_SKIP_POLISHD: ${{ runner.environment == 'github-hosted' && '1' || '0' }}
         run: ./scripts/package_app.sh release
+
+      - name: Polishing helper integration (live model + eval baseline)
+        # Spawns the helper built by the packaging step with the real pinned
+        # model, drives the production polish request path, and asserts the
+        # shared eval baseline + parent-pid tether.
+        if: runner.environment == 'self-hosted'
+        env:
+          LOCALVOXTRAL_POLISHD_TEST_ENABLE: "1"
+          LOCALVOXTRAL_POLISHD_TEST_PATH: PolishHelper/.build/xcode/Build/Products/Release/localvoxtral-polishd
+        run: swift test --filter PolishHelperIntegrationTests
 
       - name: Zip app for artifact upload
         # upload-artifact mangles symlinks/exec bits in .app bundles; ditto-zip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,13 @@ jobs:
         run: |
           if ! xcrun metal --version >/dev/null 2>&1; then
             echo "Metal toolchain not active for $(whoami); installing component..."
-            xcodebuild -downloadComponent MetalToolchain
+            # The MobileAsset catalog fetch fails transiently (observed for
+            # both the builder and runner users on 2026-07-07) — retry.
+            for attempt in 1 2 3; do
+              xcodebuild -downloadComponent MetalToolchain && break
+              echo "catalog fetch failed (attempt $attempt); retrying in 10s..."
+              sleep 10
+            done
             xcrun metal --version
           fi
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,14 @@ jobs:
           retention-days: 7
 
       - name: Zip dSYM for artifact upload
-        run: ditto -c -k --keepParent dist/localvoxtral.dSYM dist/localvoxtral-dSYM.zip
+        # Includes the polishing helper's dSYM when present (self-hosted
+        # lanes) — helper crashes (MLX C++/Metal) are the likeliest field
+        # crashes, and a dSYM that never leaves the runner symbolizes nothing.
+        run: |
+          ditto -c -k --keepParent dist/localvoxtral.dSYM dist/localvoxtral-dSYM.zip
+          if [[ -d dist/localvoxtral-polishd.dSYM ]]; then
+            ditto -c -k --keepParent dist/localvoxtral-polishd.dSYM dist/localvoxtral-polishd-dSYM.zip
+          fi
 
       - name: Upload dSYM artifact
         # Longer retention than the app: field crashes surface days after the
@@ -152,7 +159,9 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: localvoxtral-dsym
-          path: dist/localvoxtral-dSYM.zip
+          path: |
+            dist/localvoxtral-dSYM.zip
+            dist/localvoxtral-polishd-dSYM.zip
           retention-days: 30
 
       - name: Smoke test packaged app launch
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,21 @@ jobs:
         if: runner.environment == 'self-hosted'
         run: swift test --package-path PolishHelper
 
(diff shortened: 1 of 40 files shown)


T0mSIlver:
**Review pass (4 parallel reviewers over the branch diff) — findings and outcomes**, fixed in 377a147 unless noted:

**Fixed — helper:**
- CRITICAL: negative/non-numeric `Content-Length` trapped in `Data.prefix(-1)` → one malformed loopback request could crash the helper into a restart loop. Rejected at parse now, + regression tests, + 64KB head cap for header-less byte streams.
- MAJOR: a listener failure *after* ready cancelled the listener but left the process alive — dead port, and the supervisor (which only watches process exit post-ready) would never restart it. The helper now exits(1) so supervision recovers; pre-ready `.cancelled` no longer suspends `start()` forever.
- Watchdog `onParentExit` fires exactly once (dead-parent probe + kqueue event could double-fire on some libdispatch versions).

**Fixed — root app & tooling:**
- `ensureReady`'s install gate is now structural (`installKind`), not installer-data-driven, with a manager-level regression test (a misbehaving installer could previously route the bundled backend into `preconditionFailure`).
- Removed the dead `LOCALVOXTRAL_POLISHD_PATH` override (unused, wrong comment).
- Field tooling was blind to the new process: `ui-smoke.sh`'s lazy-launch invariant, `mac-diag.sh`, and `mac-crashlog.yml` all still grepped `mlx_lm.server` only — patterns updated so the tier-2 invariant actually covers polishd.
- Stale copy: onboarding consent subtitle ("Downloads mlx-lm…"), diagnostics labels.

**Fixed — tests/CI:**
- `ensureModelCached` completeness now keys on `refs/main` (written last): a cancelled half-download could leave `config.json` without weights and poison the runner's HF cache into a permanently red suite. Provisioning failures are test failures, not skips.
- `xcodebuild -disableAutomaticPackageResolution`: Package.resolved drift fails loudly instead of silently re-resolving on the runner.
- Helper dSYM now ships in CI + release artifacts (it was built and never uploaded — helper crashes are the likeliest field crashes now); missing dSYM on a non-skip package run is a hard error.

**Verified clean by reviewers:** eval corpus refactor is byte-identical (no weakened assertions anywhere; the replaced BackendManager assertions are strictly stronger), gate allowlist covers all new payloads, ci.yml expressions/ordering correct, `--exclude '.build/'` rsync-survival claim confirmed, HF symlink snapshot layouts handled, `Memory.cacheLimit` math correct, no PR#66 alert-flag violations.

**Accepted / deferred (not fixed):**
- Idle half-open loopback connections are held until the peer closes (hostile-local-peer only; real clients timeout).
- Sub-ms watchdog registration window at spawn (also a production race, negligible on macOS).
- Stale-bundle pickup from a warm xcodebuild products dir if a bundle-producing dep is ever *removed* (follow-up).
- Fork-PR artifacts have no helper (by design; managed polishing fails at spawn with a visible supervisor error).
- Old uv-installed mlx-lm venv stays on upgraded user Macs (already listed as follow-up).
- CI run is now ~6 min (was ~1.5): checkout's `git clean` cold-builds the MLX C++ core twice per run. Consider `clean: false` or an actions cache for `PolishHelper/.build` — owner call.

Re-verified after fixes: helper suite 27/27, root suite 524/524, integration-polishd 2/2, eval unchanged at required 7/7 + known-hard 3/7.

PR #90, opened 2026-07-08
Secure-input session signals: error sound + menu bar warning icon

### LINKED ISSUE #89: Secure Keyboard Entry warning is invisible in practice — needs an always-visible signal (toast / icon / sound)
Field observation (owner, 2026-07-08, PR #82 hand-testing): the Secure Keyboard Entry warning lands in the status popover, but the popover is closed while dictating — so the one place the warning appears is the one place the user isn't looking. The condition it warns about (secure input blocks ALL synthetic keystrokes — e.g. Ghostty around password prompts) makes dictation silently type nothing, which is exactly when immediate feedback matters.

Options, cheapest first:
1. **Error sound at session start** when `IsSecureEventInputEnabled()` — instant, zero UI (e.g. `NSSound` Basso). User speaks zero words into the void instead of a sentence.
2. **Menu bar icon swap** for the session — the failure icon variant already exists (`MicIconTemplate@2x_failure.png`); passive but always visible.
3. **Toast/HUD** — small non-activating auto-dismissing NSPanel near the menu bar; the right UX. `DictationOverlayController` is the NSPanel reference pattern. Needs hand-verification (no automated UI tier for it yet).

Suggested: 1+2 as a small follow-up PR now, 3 as the durable fix. Keep the popover line either way — it's the explanation once the user opens the popover, and `StatusPopoverView` line-limits protect the layout.

The popover warning itself (token-scoped, session-cleared) shipped in PR #82 and stays.

T0mSIlver:
Closes #89. Two parts, both driven by owner field-testing: the original signals (sound + icon), then the behavioral split after the field observation that the app kept running "as if healthy" in an error state — mic capturing, overlay buffering, and (worse) the overlay commit *reporting success* while secure input silently swallowed every synthetic keystroke (posting succeeds; delivery doesn't).

## What

**Signals (always fire when Secure Keyboard Entry is detected at session start):**
- `NSSound("Basso")` once — you find out after zero words, not a sentence. Plays even when the Accessibility warning owns the popover line. Injectable seam for tests.
- Menu bar icon: new `MenuBarIndicatorState.secureInputWarning` (failure asset / orange `exclamationmark.triangle.fill` fallback), tracked independently of `lastError` so higher-priority popover warnings can't hide it. Outranks `.connected`. Fresh-sampled at every attempt so it can never mask a later unrelated failure.

**Split behavior:**
- **Live Auto-Paste: refuses to start.** No mic, no socket, no managed-backend boot (the preflight runs before `ensureReady`, so a cold backend never runs a long install for a doomed session). Status line: "Blocked: Secure Keyboard Entry is on." The warning persists until the next attempt re-samples. A stale failed-commit overlay panel is reset on the refused start.
- **Overlay Buffer: proceeds** — the pipeline still produces text. The warning shows *inside the overlay panel* while buffering, and the commit **re-checks secure input at commit time**: skips the doomed synthetic attempts (which would falsely report success), copies the text to the clipboard **unconditionally** (losing the words is the only alternative), and shows the explicit failure state: "Secure input blocked auto-paste — text copied, paste it manually."

## Proof

- Suite: `Executed 593 tests, with 2 tests skipped and 0 failures (0 unexpected)` — 14 new tests across `TerminalTargetDetectorTests`, `OverlayBufferSessionCoordinatorTests`, `OverlayBufferStateMachineTests`, `DictationViewModelFailFastUXTests`.
- Codex (GPT-5.5 high), four rounds — three real P2s found and fixed, each pinned by a dedicated regression test:
  1. Icon vanished when the Accessibility warning owned the popover line → independent `sessionSecureInputActive` tracking.
  2. Stale icon masked unrelated early-exit failures on the next attempt → fresh sample per attempt.
  3. Refused start left a stale failed overlay panel onscreen + refusal ran after managed-backend boot → shared preflight before `ensureReady` + coordinator reset.
  Final verdict: "No actionable correctness regressions… covered by focused regression tests across live and overlay modes."

## Hand-verification (owner, via try-pr — UI/audio change)

- [ ] Ghostty `sudo -k && sudo true` (leave the password prompt up), **hold** (live): Basso immediately, warning icon in the menu bar, **no session starts** (no mic indicator, nothing typed), status "Blocked: Secure Keyboard Entry is on."
- [ ] Same prompt, **tap** (overlay): session runs, overlay shows the warning line while buffering; on stop the overlay shows the failure state and the text is on the clipboard — paste it manually.
- [ ] Cancel the prompt, dictate live again: no sound, normal icon, everything works; stale warning cleared.

### DIFF
--- a/Sources/localvoxtral/DictationOverlayController.swift
+++ b/Sources/localvoxtral/DictationOverlayController.swift
@@ -83,7 +83,8 @@ final class DictationOverlayController {
         let initialView = DictationOverlayView(
             phase: .idle,
             text: "",
-            errorMessage: nil
+            errorMessage: nil,
+            secureInputActive: false
         )
         hostingView = TransparentHostingView(rootView: initialView)
         // Without this, NSHostingView probes the SwiftUI content at ∞×∞ and
@@ -127,7 +128,8 @@ final class DictationOverlayController {
         hostingView.rootView = DictationOverlayView(
             phase: snapshot.phase,
             text: snapshot.bufferText,
-            errorMessage: snapshot.errorMessage
+            errorMessage: snapshot.errorMessage,
+            secureInputActive: snapshot.secureInputActive
         )
 
         let contentHeight = Self.measureContentHeight(
--- a/Sources/localvoxtral/DictationOverlayView.swift
+++ b/Sources/localvoxtral/DictationOverlayView.swift
@@ -95,7 +95,12 @@ struct DictationOverlayView: View {
     private var phaseTitle: String {
         switch phase {
         case .buffering:
-            return secureInputActive ? "Listening (secure input)" : "Listening"
+            // Actionable, not just descriptive: the commit re-checks Secure
+            // Keyboard Entry at stop, so moving focus to a normal field
+            // before then gets a real insert instead of the clipboard.
+            return secureInputActive
+                ? "Secure input — select another field before finalizing"
+                : "Listening"
         case .finalizing:
             return secureInputActive ? "Finalizing (secure input)" : "Finalizing"
         case .commitFailed:
--- a/Sources/localvoxtral/DictationOverlayView.swift
+++ b/Sources/localvoxtral/DictationOverlayView.swift
@@ -80,20 +80,33 @@ struct DictationOverlayView: View {
     let phase: OverlayBufferPhase
     let text: String
     let errorMessage: String?
+    let secureInputActive: Bool
     private let cornerRadius: CGFloat = 12
     private let bodyFontSize: CGFloat = 13
+
+    /// Warning text needs explicit light/dark variants: system `.red` over
+    /// the translucent panel material washes out on light desktops.
+    static let warningColor = Color(nsColor: NSColor(name: nil) { appearance in
+        appearance.bestMatch(from: [.aqua, .darkAqua]) == .darkAqua
+            ? NSColor(srgbRed: 1.00, green: 0.48, blue: 0.44, alpha: 1.0)
+            : NSColor(srgbRed: 0.63, green: 0.07, blue: 0.05, alpha: 1.0)
+    })
 
     private var phaseTitle: String {
         switch phase {
         case .buffering:
-            return "Listening"
+            return secureInputActive ? "Listening (secure input)" : "Listening"
         case .finalizing:
-            return "Finalizing"
+            return secureInputActive ? "Finalizing (secure input)" : "Finalizing"
         case .commitFailed:
             return "Insert failed"
         case .idle:
             return "Ready"
         }
+    }
+
+    private var isSecureInputTitle: Bool {
+        secureInputActive && (phase == .buffering || phase == .finalizing)
     }
 
     private var displayText: String {
@@ -131,7 +144,7 @@ struct DictationOverlayView: View {
             HStack(alignment: .center, spacing: 6) {
                 Text(phaseTitle)
                     .font(.system(size: 11, weight: .semibold))
-                    .foregroundStyle(.secondary)
+                    .foregroundStyle(isSecureInputTitle ? Self.warningColor : Color.secondary)
                 if phase == .finalizing {
                     ProgressView()
                         .controlSize(.small)
@@ -179,7 +192,7 @@ struct DictationOverlayView: View {
             if let errorMessage, !errorMessage.trimmed.isEmpty {
                 Text(errorMessage)
                     .font(.system(size: 11))
-                    .foregroundStyle(.red)
+                    .foregroundStyle(Self.warningColor)
                     .fixedSize(horizontal: false, vertical: true)
                     .frame(maxWidth: .infinity, alignment: .leading)
             }
@@ -188,8 +201,10 @@ struct DictationOverlayView: View {
         .frame(minWidth: 400, idealWidth: 420, maxWidth: 540, alignment: .leading)
         .background(RoundedMaterialBackground(cornerRadius: cornerRadius))
         .overlay(
+            // Hairline must survive both appearances: pure white vanished
+            // against light desktops.
             RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
-                .strokeBorder(Color.white.opacity(0.25), lineWidth: 1)
+                .strokeBorder(Color.primary.opacity(0.25), lineWidth: 1)
         )
         .clipShape(RoundedRectangle(cornerRadius: cornerRadius, style: .continuous))
         .compositingGroup()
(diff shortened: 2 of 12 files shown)


T0mSIlver:
Codex review (GPT-5.5 high), two rounds:
1. Found a real P2 — the warning icon gated on `currentErrorToken`, which the Accessibility warning deliberately outranks, so the icon silently stayed `.connected` in the double-warning case. Fixed in `117be2a` (`sessionSecureInputActive` tracked independently of `lastError`), pinned by the extended `testSoundStillPlaysWhenAccessibilityWarningOwnsThePopoverLine`.
2. Re-review after the fix: *"consistently wired and covered by regression tests. No actionable defects."*

CI green (583 tests). Awaiting owner hand-verification of the sound + icon (checklist in the description) before merge.

T0mSIlver:
## Field fixes (owner hand-test round 2) + three more codex rounds

**Fixed both reported issues:**
1. **Icon stayed red after releasing the modifier** — the refused start reset the hold flags, so no release branch ever cleared it. The warning now ends when the attempt gesture ends: release for holds, immediately for tap/popover-button starts (which have no release event — two codex-found latch variants). The popover line stays as the explanation; the refusal preflight also moved before the mic-permission gate (a doomed session must not trigger a mic prompt — and per-user TCC state was why tests diverged between CI and the build host).
2. **Overlay stayed on after ending a session with buffered text** — the clipboard fallback returned `.failed`, and failed panels persist by design. It's now its own outcome: status reads **"Copied — paste it manually."** (not "Insert failed."), and the panel dismisses after a 4s readable hold anchored to the fallback message. Genuine failures keep the persistent panel — including the codex-found edge where the *pasteboard write itself* fails: then we don't claim "copied", and the panel (with the text) stays.

Proof: `Executed 600 tests, with 2 tests skipped and 0 failures` — 7 new pins (`testRefusedStartIconClearsWhenTheHoldGestureEnds`, `testRefusedModifierTapDoesNotLatchTheWarningIcon`, `testRefusedPopoverToggleStartDoesNotLatchTheWarningIcon`, `testClipboardFallbackOutcomeDismissesOverlayWithReadableHold`, `testClipboardFallbackPanelDismissesAfterReadableHold`, `testFailedCommitOutcomeStillKeepsOverlayPanel`, `testFailedClipboardWriteUnderSecureInputKeepsThePersistentPanel`). Codex (GPT-5.5 high) rounds 5–8: three more P2s found, fixed, pinned; final: *"No discrete, actionable regressions."* CI green.

T0mSIlver:
## Field fixes (round 2) — 871606b

Three owner reports on the secure-input overlay flow:

**1. Menu bar icon flashed yellow during LLM polishing after dismiss** — `stopDictation` sets `isFinalizingStop` and switches the indicator to the yellow session icon; the gesture-end clear (`clearSecureInputRefusalSignalsIfAttemptEnded`) only checked `isDictating`/`isConnecting`, so releasing the shortcut cleared `sessionSecureInputActive` mid-finalization and the icon fell from red to yellow until teardown. The clear now also guards on `!isFinalizingStop`; session teardown (which already clears the flag after the clipboard-fallback commit) owns the clear once a session actually ran.

**2. Overlay warning text too long / too punctuated** — the red buffering sentence ("Secure input is on — auto-paste is blocked; the text will be copied at stop.") is gone. The phase title carries the marker instead: **"Listening (secure input)"** / **"Finalizing (secure input)"**, tinted with the warning color. `setBufferingWarning(message)` became a `setSecureInputWarning()` flag on the state machine snapshot.

**3. Light-mode contrast** — warning text now uses an explicit light/dark pair (darker red in light mode; system `.red` washed out over the translucent panel material), and the panel hairline stroke uses `Color.primary` instead of pure white, which vanished on light desktops.

### Proof

Regression test failing before the fix (bug reproduced — icon reports the yellow session state):

```
Tests/localvoxtralTests/TerminalTargetDetectorTests.swift:404: error: -[localvoxtralTests.TerminalTargetDetectorTests testMenuBarSecureWarningSurvivesStopFinalizationAndPolish] : XCTAssertEqual failed: ("connected") is not equal to ("secureInputWarning") - the release-time clear must not drop the warning to the session icon mid-polish
```

After the fix, full unit suite:

```
Test Case '-[localvoxtralTests.TerminalTargetDetectorTests testMenuBarSecureWarningSurvivesStopFinalizationAndPolish]' passed (0.002 seconds).
Test Case '-[localvoxtralTests.OverlayBufferStateMachineTests testSecureInputMarkerSetsOnlyWhileBufferingAndResetsOnNewSession]' passed (0.000 seconds).
Test Case '-[localvoxtralTests.OverlayBufferSessionCoordinatorTests testShowSecureInputWarningRendersInsideTheOverlayWhileBuffering]' passed (0.000 seconds).
Test Suite 'localvoxtralPackageTests.xctest' passed — Executed 601 tests, with 2 tests skipped and 0 failures
```

**Hand-verification still needed (UI change):** the new title wording/tint and light-mode contrast aren't covered by automated tests — please `./scripts/try-pr.sh 90` once CI is green and check (a) icon stays red through the polish window when dismissing in a secure field, (b) the overlay title reads "Listening (secure input)" with no red sentence below, (c) warning text + panel border are readable on a light desktop.

T0mSIlver:
Follow-up on the round-2 comment: the buffering title is now the actionable hint **"Secure input — select another field before finalizing"** (instead of the bare "Listening (secure input)" marker). This is accurate because the commit re-checks Secure Keyboard Entry at stop time — focus a normal field before stopping and the text inserts there for real; stay in the secure field and it falls back to the clipboard as before. Finalizing title unchanged ("Finalizing (secure input)"). Unit suite green after the change (601 tests, 0 failures). Same hand-check applies via `try-pr.sh 90`.

T0mSIlver:
## Codex review (rounds 8–9) — clean

Full-PR read-only review by Codex (GPT-5.5 high). Two findings, both fixed with failing-before/passing-after regression tests:

- **Round 8** (`b91e7f4`): secure input turning ON during managed-backend boot made the delayed refusal fire after the initiating gesture ended — red icon + "Blocked" status wedged. The startup task now ends the refusal signals itself when no gesture is still held. Regression: `testDelayedSecureInputRefusalAfterManagedStartupDoesNotWedgeTheIcon` (failed pre-fix: icon stuck on `secureInputWarning`, status stuck on "Blocked").
- **Round 9** (`e539af5`): same wedge via the mic-permission grant continuation. Fixed at the continuation call site; added a `requestAccess` test seam so `testDelayedSecureInputRefusalAfterMicPermissionGrantDoesNotWedgeTheIcon` drives the real continuation deterministically (failed pre-fix the same way).

Final verification pass (fresh Codex run over both fixes + a sweep of every async re-entry into the refusal path — realtime events, startup permissions, Accessibility, network, mic-health, backend warmup): **NO FINDINGS**. The only wedge-capable continuations were the two above, both covered.

Unit suite: 603 tests, 0 failures. CI green on `e539af5`. (The one red CI run on `b91e7f4` was `BackendManagerTests.testSupervisorStateMirrorMarksLaterFailureAndNextEnsureDoesNotShortCircuit` — a pre-existing state-mirror race, passes 3/3 in isolation on the same host and passed on the next run; same family as the earlier crash-restart race fix #53.)

PR #100, opened 2026-07-09
Live Auto-Paste: hold back only text a rule can still match

T0mSIlver:
## What

Live Auto-Paste applies replacement-dictionary rules *before* typing, because released text can never be un-typed (there are no backspaces in the insertion path — terminals can't support them, field bug 2026-07-06). To guarantee that, `LiveHoldBackReplacementStream` held back the trailing partial word plus `maxRuleWordCount - 1` complete words.

`maxRuleWordCount` is the **global max across every rule**, so a single four-word entry anywhere in `replacements.toml` made *every* live-typed word lag three complete words behind speech — including text that could never match that rule. Live Auto-Paste is the mode where text appears as you speak, so this was directly felt.

This bounds the hold by **viable prefix** instead: hold from the earliest offset where the remaining text is still a live prefix of some rule. Text that begins no rule is released immediately.

It's cheap because `makeRegex` escapes every key with `escapedPattern` and joins the words with `\s+` — **rules are literal word lists**, not user regexes, so prefix-matching needs no regex machinery. `LiveReplacementRule` now carries its split `keyWords`.

The old N-word window is kept as a **floor**: both it and the prefix scan are lower bounds on a future correction's `startOffset`, so the safe limit is the larger of the two. Single-word dictionaries are bit-for-bit unchanged (the scan never runs).

### The subtle part

Candidate match starts are **not** whitespace-delimited word starts. The pattern's `(?<![\p{L}\p{N}])` lookbehind only forbids a preceding *letter or digit*:

- `voxtral` matches inside `foo-voxtral` (preceded by `-`)
- for a decomposed `e` + U+0301, the code point before the match is a combining mark (category `Mn`), which the lookbehind accepts

A scan keyed on word starts would release `foo-voxtral ` and then let a later `rocks ` fire a `voxtral rocks` correction reaching back into already-typed text — exactly the invariant the design exists to protect. So the scan tests every offset not preceded by a letter/digit, comparing the **last unicode scalar** of the preceding grapheme to mirror the regex rather than the grapheme as a whole.

### Behavior change in tests

Three existing tests asserted the old latency and now assert the earlier release. No assertion was weakened — each still checks the same final text, and two are strictly stronger (they now pin *when* text is released, not just that it eventually is):

- `testMultiWordRuleHoldsBackPossiblePrefixWords`: `ingest("code ")` now returns `"Claude Code "` instead of `"Claude "` + a stop-flush remainder.
- `testCorrectsMultiWordKeyWithLookbackWindow` (view model): `localvoxtral ` is typed live rather than at the stop flush.
- `testFlushedRemainderIsSanitized` / `testTerminalLikeSessionSanitizesNewlinesInFlushedRemainder`: these relied on a two-word rule holding *everything* so the newline reached the remainder flush. That setup no longer holds text the rule can't match, so they now use a rule the input stays a prefix of (`run something` / `run so`), preserving the original intent. Added `testCollapsedNewlineRunSpansAnEarlyRelease` to cover the new early-release-then-collapse interaction.

## Proof

```
Executed 618 tests, with 2 tests skipped and 0 failures (0 unexpected) in 8.121 (8.164) seconds
```

- [x] Realtime integration green — `./scripts/remote-build.sh integration`

```
Executed 7 tests, with 2 tests skipped and 0 failures (0 unexpected) in 18.537 (18.540) seconds
```

- [x] Regression tests fail before the fix, pass after

The hazard above is not hypothetical. Degrading **only** the candidate-start test to the naive word-start version (`return isWhitespace(characters[offset - 1])`) and re-running:

```
=== FAILING TESTS UNDER NAIVE WORD-START SCAN ===
LiveHoldBackReplacementStreamTests.testMatchStartAfterCombiningMarkIsHeldBack
LiveHoldBackReplacementStreamTests.testMatchStartAfterPunctuationIsHeldBack
LiveHoldBackReplacementStreamTests.testReleasedTextMatchesBatchCorrectionForAdversarialInputs
LiveHoldBackReplacementStreamTests.testReleasedTextMatchesBatchCorrectionForRandomChunkings

XCTAssertEqual failed: ("foo-cloud  Code ") is not equal to ("foo-Claude Code ")
XCTAssertEqual failed: ("évoxtral ") is not equal to ("é")
```

`foo-cloud  Code` is the corrupted-output signature of a correction rewriting text that had already been typed. With the fix, all four pass:

```
Executed 34 tests, with 0 failures (0 unexpected) in 1.837 (1.840) seconds
```

### On the property test

`testReleasedTextMatchesBatchCorrection*` asserts the invariant directly: for any chunking, the concatenated releases must equal correcting the whole transcript in one batch pass. Releasing too early can only show up as a final output that disagrees with the batch result.

Worth recording — **the first version of this property test passed under the naive scan**, i.e. it was worthless. Its corpus paired the multi-word rule `voxtral rocks` with a single-word `voxtral` rule; the shorter rule fires at the boundary after `foo-voxtral`, rewrites it to `foo-localvoxtral `, and the multi-word rule can then never reach back across a release. The hazard was unreachable by construction. The corpus now keeps `cloud code` free of any single-word rule on `cloud` (documented on `invariantDictionary`), and the adversarial inputs are enumerated explicitly rather than left to the RNG.

### Not verified by hand

No UI change. Behavior was verified by the suites above, not by dictating into a live terminal.

## Note

One unrelated flake seen once during a full-suite run: `BackendManagerTests.testSupervisorStateMirrorMarksLaterFailureAndNextEnsureDoesNotShortCircuit`. It passed 3/3 in isolation and on the next full-suite run, and touches no file in this diff. Flagging rather than ignoring — that suite was deflaked in c1b2baf and may still be racy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -211,7 +211,9 @@ Key subsystems:
   (vLLM/voxmlx) over `BaseRealtimeWebSocketClient`
 - Text merge: `TextMergingAlgorithms` (pure functions — overlap merge,
   word-boundary stabilization, punctuation spacing), `FirstChunkPreprocessor`
-- Insertion: `TextInsertionService` (AX replace → Unicode CGEvents → Cmd+V)
+- Insertion: `TextInsertionService` (AX replace → Unicode CGEvents → Cmd+V);
+  Live Auto-Paste replacements run through `LiveHoldBackReplacementStream`
+  before typing — see "Known tradeoffs" below for the latency it costs
 - Overlay: `OverlayBufferSessionCoordinator` (session + hold-before-dismiss
   timing), `OverlayBufferStateMachine`, `DictationOverlayController` (NSPanel)
 - Backends: `BackendManager` lazily bootstraps app-managed local serving on
@@ -230,6 +232,15 @@ Key subsystems:
   MLX Swift inference + a minimal loopback OpenAI server + parent-pid
   watchdog), supervised like any managed backend on port 8472. Replaced the
   uv-installed mlx-lm fork wheel (upstream mlx-lm unmaintained, 2026-07).
+
+## Known tradeoffs — deliberate, not bugs
+
+- **Live Auto-Paste holds back the tail of the transcript.** Replacements are
+  applied before typing (nothing is ever un-typed — there are no backspaces in
+  the insertion path, and terminals can't support them: field bug 2026-07-06),
+  so `LiveHoldBackReplacementStream` withholds the trailing partial word plus
+  any suffix that is still a live prefix of a dictionary rule. Nothing is lost
+  (`flushRemainder()` releases it at stop) but it costs latency of appearance.
 
 ## Conventions
 
--- a/Sources/localvoxtral/AppConfigStore.swift
+++ b/Sources/localvoxtral/AppConfigStore.swift
@@ -31,7 +31,7 @@ struct ReplacementDictionary: Equatable, Sendable {
                         regex: regex,
                         replaceWith: entry.replaceWith,
                         matchLength: normalized.count,
-                        wordCount: normalized.split(whereSeparator: \.isWhitespace).count,
+                        keyWords: normalized.split(whereSeparator: \.isWhitespace).map(String.init),
                         originalOrder: originalOrder
                     )
                 )
@@ -158,12 +158,21 @@ struct LiveReplacementRule: Comparable {
     let regex: NSRegularExpression
     let replaceWith: String
     let matchLength: Int
-    let wordCount: Int
+    /// The literal, whitespace-separated words of the match key, in order.
+    /// `makeRegex` escapes every key with `escapedPattern` and joins the words
+    /// with `\s+`, so a rule is a literal word list — no metacharacter ever
+    /// survives. `LiveHoldBackReplacementStream` prefix-matches against these
+    /// words directly to decide how little text it can hold back.
+    let keyWords: [String]
     let originalOrder: Int
+
+    var wordCount: Int {
+        keyWords.count
+    }
 
     static func == (lhs: LiveReplacementRule, rhs: LiveReplacementRule) -> Bool {
         lhs.matchLength == rhs.matchLength
-            && lhs.wordCount == rhs.wordCount
+            && lhs.keyWords == rhs.keyWords
             && lhs.originalOrder == rhs.originalOrder
             && lhs.replaceWith == rhs.replaceWith
             && lhs.regex.pattern == rhs.regex.pattern
--- a/Sources/localvoxtral/AppConfigStore.swift
+++ b/Sources/localvoxtral/AppConfigStore.swift
@@ -31,7 +31,9 @@ struct ReplacementDictionary: Equatable, Sendable {
                         regex: regex,
                         replaceWith: entry.replaceWith,
                         matchLength: normalized.count,
-                        keyWords: normalized.split(whereSeparator: \.isWhitespace).map(String.init),
+                        foldedKeyWords: normalized
+                            .split(whereSeparator: \.isWhitespace)
+                            .map { String($0).caseFoldedForMatching },
                         originalOrder: originalOrder
                     )
                 )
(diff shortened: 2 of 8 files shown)


T0mSIlver:
## Codex review: 1 critical finding, confirmed and fixed

Ran an independent read-only review (`codex exec`, gpt-5.5, high reasoning) against the diff. It found a **real violation of the never-un-type invariant**. I reproduced it before fixing.

### Critical: full case folding broke the viable-prefix check

`NSRegularExpression` full-case-folds literals under `.caseInsensitive`, so the rule `foo ßx` matches the text `"foo ssx"`. The literal prefix check asked `"ßx".hasPrefix("s")`, got `false`, judged the tail `"foo s"` dead, and released it. When `"sx "` arrived the correction started at offset 0 with 4 characters already typed.

Reproduced on the build host before any fix:

```
PROBE regex ß-pattern vs "foo ssx": matches=1
PROBE hasPrefix ßx <- s: false
PROBE after 'foo s' released="foo "
PROBE stream="foo " batch="X "
```

The stream emits `"foo "` where the batch correction gives `"X "` — wrong text *and* lost text.

**Fix** (291b02c): fold both sides with `folding(options: .caseInsensitive)` before comparing. Full folding is per-code-point, so it distributes over concatenation and a folded prefix stays a prefix — exactly what the predicate needs. Verified `ß→ss`, `ﬁ→fi`, `U+212A→k`. Three regression tests added.

### Rejected fix: the regex's own `.hitEnd` flag

The tempting fix is to stop hand-rolling the predicate and ask each rule's own regex whether more input could match, via the `.hitEnd` matching flag. I implemented it, and it is **wrong**. Measured:

```
fox-rule   "hello ":            hitEnd     <- 'h' can never begin "the quick brown fox"
fox-rule   "there wide world ": (none)
cloud-rule "hello ":            (none)
```

`hitEnd` reports whether ICU read to the end of the *input*, not whether more input could complete a match; for `the quick brown fox` it fires on any input shorter than the minimum match length. It is *safe* (over-holds) but it holds `world ` forever, silently reintroducing the exact latency bug this PR removes. It cost 3 unit-test failures to discover. Documented in `isViableRulePrefix` so nobody simplifies back to it.

### Minor (also fixed): the property test was a weak oracle

Codex correctly noted that comparing final concatenated output against a batch correction can pass vacuously — a correction that rewrites already-released text can reproduce it verbatim. Its example: `foo bar` → `foo baz`, release `"foo "`, rewrite from offset 0, final output still matches.

Rather than a cleverer oracle, the invariant is now asserted where it is defined — `LiveHoldBackReplacementStream` asserts every correction starts at or after `releasedCharacterCount`. This turns all 621 tests into invariant checks. On the pre-fix code it fires:

```
LiveHoldBackReplacementStream.swift:108: Assertion failed:
hold-back violated: correction starts at 0 but 4 characters were already typed
```

`foo bar`/`foo baz` and the folding cases are now in the adversarial corpus.

### Re-verified after the fix

```
Executed 621 tests, with 2 tests skipped and 0 failures (0 unexpected) in 8.766 (8.800) seconds
Executed 7 tests, with 2 tests skipped and 0 failures (0 unexpected) in 23.415 (23.417) seconds   (integration)
```

Codex's verdict on the original diff — *"the invariant does not hold as written"* — was correct.

T0mSIlver:
## opencode review (GLM-5.2, read-only): no critical findings

Second independent review, after the codex round. It reached **VERDICT: the invariant holds**, and independently re-derived why each of the two earlier fixes is necessary and sufficient (folding distributes over concatenation so a folded prefix stays a prefix; the last-scalar lookbehind check accepts the `Mn` combining-mark case). It traced the traps I care about — replacement reproducing its own key, punctuation-internal boundaries, empty partial word, match start after hyphen/mark, multi-space runs — and found all of them inside the bound.

### Its one substantive claim was a false positive, and I checked rather than assumed

> `applyCompletedBoundaryCorrections` infinite-loops when a rule's replacement contains the rule's own complete key, e.g. `{"foo": "big foo"}` — each `apply` regenerates a match the `while let` re-finds.

That would be an **app hang mid-dictation**, so I tested it instead of arguing from the code. It does not loop: `apply()` sets `scanOffset = correction.startOffset + correction.replacementText.count`, resuming the scan past the inserted text. Four variants, char-by-char chunking, all terminate in microseconds and match the batch correction:

```
PROBE stream foo -> big foo:            "big foo bar "        equal=true
PROBE stream abc def -> xyz abc def:    "xyz abc def tail "   equal=true
PROBE stream identity a b -> a b:       "a b c "              equal=true
PROBE stream foo -> foo foo:            "foo foo "            equal=true
PROBE completed without hanging
```

The failure mode it imagined would be a hang, not a red test, so I pinned it anyway (9870110) — `testReplacementContainingItsOwnKeyTerminates`.

### One real thing I found while waiting on the review

Probing the `\s+` vs `isWhitespace` question from the brief: **ZWSP (U+200B) is in `CharacterSet.whitespacesAndNewlines` but is not matched by regex `\s`.**

```
PROBE ws ZWSP(U+200B): regex\s=false isWhitespace=true *** DISAGREE ***
(every other whitespace scalar tested — NBSP, NEL, VT, FF, U+1680, U+2000, U+202F, U+2028/9, U+3000 — agrees)
```

The stream splits words on ZWSP; the matcher does not. This does **not** break the invariant, because `lookbackStart` uses the same `isWhitespace`, so the corrector's search window is truncated identically and no correction is ever produced. Stream and batch agree on every ZWSP case I tried.

It does mean a **pre-existing** live-vs-overlay inconsistency: `ReplacementDictionary.apply(to:)` (overlay path, raw regex over full text) replaces a key spanning a ZWSP, while the live path silently cannot. That predates this PR — `lookbackStart` has always used `CharacterSet.whitespacesAndNewlines` — so I am not fixing it here. Filing it as a separate issue.

### State

```
Executed 622 tests, with 2 tests skipped and 0 failures (0 unexpected) in 8.933 (8.964) seconds
```

Two reviewers, two rounds: codex found one real invariant violation (fixed in 291b02c), opencode found none.

PR #104, opened 2026-07-09
Overlay Buffer font size setting: scale the whole overlay panel

T0mSIlver:
## What

Adds an **Overlay Buffer font size** setting (Settings > Dictation > Overlay Buffer > Font size slider, 10–24pt, default 13pt). The whole overlay panel scales proportionally from it: body/title/error fonts, header height, panel width, and the 4-line scroll cap.

Implementation notes:

- New `OverlayLayoutMetrics` is the single source of truth for the sizing that `DictationOverlayView` (SwiftUI content) and `DictationOverlayController` (AppKit `boundingRect` panel measurement) previously duplicated by hand — the two must agree exactly or the panel clips the last line, so both now derive from one struct. At the default 13pt it reproduces the legacy constants bit-for-bit (asserted by test).
- `SettingsStore.overlayBufferFontSize` persists like `modifierOnlyHoldDelay`: raw on write, clamped on load; `OverlayLayoutMetrics` also clamps at use.
- The controller reads the setting through a `@MainActor` `fontSizeProvider` closure on every render, so a change applies to the next overlay render without restarting the app.
- Settings pane follows the owner rule: the new "Overlay Buffer" group is always present (constant group structure).

## Proof

  ```
  Executed 634 tests, with 2 tests skipped and 0 failures (0 unexpected) in 6.512 (6.544) seconds
  ```
  New coverage: `OverlayLayoutMetricsTests` (clamping, legacy-constant parity at 13pt, monotonic scaling, 4-line cap at every size, error-row height) and `SettingsStoreTests.testOverlayBufferFontSize_*` (default, persistence round-trip, clamp-on-load) — all present and passing in the run above.
- [x] Realtime integration — runs in CI on the self-hosted runner (change is UI/settings only; realtime pipeline untouched).
- [x] Codex review (`codex exec review --base origin/main`, gpt-5.5 high): "No actionable correctness issues were identified in the diff. The new overlay font-size setting is wired through settings, layout measurement, rendering, and tests without evident breakage."
- [ ] UI change: **not yet hand-verified** — I can't drive the Mac GUI from this session. The panel-height math that usually regresses (view vs. controller measurement drift) is unit-tested, including exact legacy parity at the default size. Suggested hand check via `./scripts/try-pr.sh <this PR>`: dictate in Overlay Buffer mode at 10pt, 13pt, and 24pt and confirm no clipped last line, sane panel width, and that a slider change applies to the next dictation.

### DIFF
--- a/Sources/localvoxtral/DictationOverlayController.swift
+++ b/Sources/localvoxtral/DictationOverlayController.swift
@@ -45,9 +45,10 @@ private final class NonActivatingPanel: NSPanel {
 final class DictationOverlayController {
     private let panel: NonActivatingPanel
     private let hostingView: TransparentHostingView<DictationOverlayView>
-    private let panelWidth: CGFloat = 420
-    private let maximumPanelHeight: CGFloat = 420
     private let cornerRadius: CGFloat = 12
+    /// Reads the user's overlay font size on every render so setting changes
+    /// apply to the next dictation without restarting the app.
+    private let fontSizeProvider: @MainActor () -> Double
 
     /// Locked placement state for the current session. Set on first render,
     /// cleared on hide. Prevents the panel from flipping between above/below
@@ -61,7 +62,12 @@ final class DictationOverlayController {
     private var lockedPlacement: Placement?
     private var lockedOriginX: CGFloat?
 
-    init() {
+    init(
+        fontSizeProvider: @escaping @MainActor () -> Double = {
+            OverlayLayoutMetrics.defaultBodyFontSize
+        }
+    ) {
+        self.fontSizeProvider = fontSizeProvider
         panel = NonActivatingPanel(
             contentRect: NSRect(x: 0, y: 0, width: 420, height: 120),
             styleMask: [.borderless, .nonactivatingPanel],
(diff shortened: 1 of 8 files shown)


T0mSIlver:
**Updates since the codex review:**

1. **Default bumped 13pt → 14pt** (scale-1.0 base stays 13, so the default now renders at 14/13 scale; parity tests pin the base, new tests pin the 14pt default).
2. **opencode review (GLM-5.2, read-only)** — verdict: *MERGE WITH NITS*, one LOW correctness finding, both nits taken:
   - **LOW (fixed):** the controller locks the panel's X origin per session assuming constant width, but metrics were re-read on every render — dragging the font-size slider mid-dictation would widen the panel past its locked origin, potentially off-screen. Fixed by `OverlaySessionMetricsLock`: the font size is snapshotted on a session's first render and released on `hide()`, so changes apply to the next dictation. Regression test: `testSessionLockFreezesFontSizeUntilUnlocked`.
   - **Nit (fixed):** the 4-line scroll cap no longer reuses `stackSpacing`; it has its own `bodyScrollSlack` constant.
   - **Nit (fixed):** `DictationOverlayView.metrics` is a required `let`.

Proof after fixes: `./scripts/remote-build.sh test` — `Executed 636 tests, with 2 tests skipped and 0 failures (0 unexpected)`.

PR #112, opened 2026-07-09
Repo vocabulary: index the focused terminal's repo, hint the polisher (opt-in, loopback-only)

T0mSIlver:
# Repo vocabulary from the focused terminal (opt-in, loopback-only)

Feature 5 of the polish-for-agents series (stacked on #106 ← #105 ← #103 ← #101).

When dictating into a terminal, the polisher now knows the project's actual file
names and identifiers: the app resolves the frontmost terminal's working
directory from its AX window title, walks up to the git root, indexes
`git ls-files` output (bounded subprocess, 5-min TTL cache keyed on HEAD mtime),
fuzzy-matches transcript n-grams against the vocabulary off the main actor, and
injects the top matches as extra replacement-dictionary entries into the polish
prompt. Spoken "use auth dot t s" lands as `useAuth.ts`.

## Privacy / gating

- Setting `repoVocabularyEnabled`, **default OFF** (one row inside the existing
  Polishing group — group count/identity unchanged).
- Attaches **only to loopback polishing endpoints** (same gate as clipboard
  context, #105). Also skipped entirely when the active prompt template lacks
  the `{{replacement_dictionary}}` slot.
- Logs and session records carry **counts, never content**
  (`vocab:<n>` in the provenance summary; "Repo vocabulary attached: N entries").

## Design points

- **Sanctioned rewrites** (resolves the F1↔F5 conflict): the token guard (#101)
  used to protect the STT's *wrong* filename-shaped token and deterministically
  revert the very correction the vocabulary hint requested.
  `PolishTokenGuard.verifyAndRepair` gained an append-only
  `sanctionedReplacements: [(from, to)] = []` parameter: a missing protected
  token counts as preserved iff it canonically equals a pair's `from` AND that
  pair's `to` occurs standalone in the polished text — checked *before*
  near-miss repair (the revert path). Default `[]` keeps F1–F4 byte-identical;
  `PolishTokenGuardTests` diff has **zero deleted lines** (pure additions).
- **Technical-signal gate**: a vocabulary term needs a dot, separator, or
  internal capital (camelCase/PascalCase). `Tests`/`Resources`/`docs`/`main`
  never become hints that capitalize ordinary prose; accepted loss:
  `Makefile`/`LICENSE`-style bare names (documented).
- **No main-actor stalls**: exact-index + ±1-length fuzzy buckets built at
  construction (TTL-amortized); resolve-cwd → index → match runs in one
  `Task.detached` hop; only the AX title read stays on main, with
  `AXUIElementSetMessagingTimeout(0.5)` on both elements.
- **Bounded subprocess**: `/usr/bin/git ls-files -z` with 2 s timeout, 2 MB cap,
  `POSIXPipeRead` (never `FileHandle.availableData`), terminationHandler +
  2 s-bounded exit wait, `isRunning`-guarded signals, hardened env
  (`GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null`,
  `GIT_TERMINAL_PROMPT=0`).
- Slot-in with PR #99/#100 maintained: no `LLMPolishingRequest`/`Configuration`
  shape changes, no live-replacement machinery touched, `AppConfigStore`
  append-only.

## Reviews (three rounds)

1. Self-review + build round: 2 compile fixes, then green (739).
2. Independent opencode/GLM review — 4 P3s fixed: template-without-placeholder
   gate; clipboard-read atomicity (context read hoisted back pre-Task next to
   the payload read); filename sanitization at prompt render; byte-cap path no
   longer burns a 0.5 s grace wait. (745 green.)
3. Fable deep review — 2 P2s + 4 P3s fixed: sanctioned rewrites (P2, above);
   matcher complexity + main-actor beachball (P2, above); AX/FS main-actor
   hangs; unbounded `waitUntilExit`; raw kill on possibly-reaped pid;
   common-word false hints. (754 green.)

## Proof

`./scripts/remote-build.sh` (build + full unit suite) final tail:

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-07-10 00:18:06.856.
	 Executed 754 tests, with 2 tests skipped and 0 failures (0 unexpected) in 7.078 (7.118) seconds
```

754 tests (694 → 754 across the round: +45 pure indexer/matcher/cache/parser
cases, +5 VM-level, +2 settings, +4 sanctioning quartet, +tech-signal and
scale tests), 2 skipped (env-gated Tier-1), 0 failures, 0 warnings in changed
files under Swift 6.2 strict concurrency.

Behavior-demonstrating tests (named per proof culture):

- `RepoVocabularyIndexerEndToEndTests.testHarvestsVocabularyFromRealRepo` —
  live `git init/add/commit/ls-files` through the full
  `RepoVocabularyService.entries` title→entries pipeline; asserts
  `useAuth.ts`, `UserSessionManager.swift`, branch.
- `DictationViewModelPolishTokenGuardTests.testSanctionedVocabularyRewriteSurvivesTokenGuard`
  — VM end-to-end: vocabulary entries → sanctioned pairs → stub polish applies
  the case fix → committed text keeps `useAuth.ts` (pre-fix the guard reverted
  it; quartet `testSanctionedRewriteIsNotRevertedByCaseRepair` /
  `testSanctionedFuzzyRewriteDoesNotFallBack` /
  `testWithoutSanctionedPairsOldGuardBehaviorHolds` /
  `testSanctionedToAbsentFallsBackNormally` pins each side).
- `RepoVocabularyMatcherTests.testMatcherHandlesLargeVocabulary` — 20k terms ×
  300-word transcript, correctness at scale, 0.130 s (no wall-clock assertion
  per repo rule; the index restructure is the guarantee, this is the canary).
- OFF path byte-identical: all pre-existing polish/context/macro/profile VM
  tests pass unchanged; `ClipboardPayloadMacroTests` /
  `PolishContextClipboardReaderTests` / `TerminalTargetDetectorTests` have no
  diff at all.

No `integration-polishd` / `eval-llm` run needed: no prompt-template files,
eval corpus, or request/config shapes changed.

## Needs Mac-side hand verification (will note before merge)

- AX window-title read is seam-stubbed in unit tests. Real check: setting ON,
  loopback endpoint, dictate into Terminal/iTerm2 sitting in a git repo →
  `Log.polishing` shows "Repo vocabulary attached: N entries" and a spoken
  "use auth dot t s" commits as the exact filename. Requires the existing
  Accessibility grant; silent skip without it.
- Terminal.app shows a bare folder name by default (correctly ignored);
  iTerm2/Ghostty/WezTerm `~/path` titles resolve. Worth spot-checking across
  emulators.
- New settings row visual pass via `try-pr.sh` (same pass pending for
  #103/#105/#106).

---

## 2026-07-11 field-test fixes

Two new commits after restacking on the amended #105:

**1. Clipboard entities join the sanctioned-rewrite pipeline (+ substring-canonical sanctioning).** Field log (T5): `Token guard discarded polish; 1 protected token(s) not preserved` — clipboard context ON, clipboard holding `UserSessionManager.swift`, dictated "look at user session manager dot swift"; STT emitted the guard-protected `manager.swift`, the model correctly produced `UserSessionManager.swift`, and the guard discarded the whole polish. Now: code-like entities are extracted from the (already privacy-gated) clipboard excerpt with the guard's OWN recognizer (`ClipboardVocabulary`, no second grammar), matched against transcript n-grams with the exact repo-vocabulary matcher, injected as dictionary-slot hints, and registered as sanctioned pairs. The guard's sanction check is now substring-canonical (a protected token canonically contained in a sanctioned alias whose `to` the model produced counts as preserved) — closing the previously documented "known hole". Gates unchanged: toggle ON + loopback only + concealed/transient never read; counts-only logging (`clipboard-vocab:<n>`).

Also: repo-vocab **skip reasons elevated `.debug` → `.info`** (the unified log store does not persist `.debug`, which made the field no-attach undiagnosable) and the two remaining silent skips (empty vocabulary, no transcript-relevant matches) got `.info` lines. All static strings + counts; "Repo vocabulary attached: N entries" was already `.info`.

**2. Ghostty's `..`-abbreviated window titles resolve (T6 root cause).** Ghostty elides leading path components: the owner's tab title was exactly `../Desktop/projects/supervoxtral`, the parser extracted `/Desktop/projects/supervoxtral`, the existence check failed, and the feature silently no-oped. `TerminalWorkingDirectoryResolver` gains a home-anchored FALLBACK tier (consulted only after every exact candidate fails): `../X/Y` → `$HOME/X/Y`, and a non-existent absolute `/X/Y` not under home → `$HOME/X/Y`. Exact candidates keep priority; `workingDirectoryCandidates` is untouched (existing parser table byte-identical).

Proof — new tests shown failing with the fixes behaviorally reverted (guard containment → equality, clipboard-vocab wiring off, resolver fallback off), passing after:

```
before:
  testClipboardEntityCorrectionSurvivesTokenGuard                      failed  (T5 end-to-end: guard discards)
  testClipboardEntitySanctioningWorksWithoutDictionarySlot             failed
  testSanctionedAliasContainingProtectedTokenDoesNotFallBack           failed
  testUnrelatedProtectedTokenStillGuardedUnderContainment              failed
  testGhosttyAbbreviatedTitleResolvesHomeAnchored                      failed  (T6)
  testNonExistentAbsolutePathFallsBackToHomeAnchored                   failed
  Executed 78 tests, with 10 failures (0 unexpected)
after:
  Executed 78 tests, with 0 failures (0 unexpected)   (PolishTokenGuardTests + resolver + ClipboardVocabularyTests)
  Executed 30 tests, with 0 failures (0 unexpected)   (all RepoVocabulary* classes)
```

Full unit suite on the restacked stack tip (feat/polish-revert-affordance): `Executed 827 tests, with 2 tests skipped and 0 failures (0 unexpected)`.

### Review follow-up (codex P2): fallback constrained to `../`-prefixed titles

The first cut re-anchored ANY non-existent absolute path under home — an SSH/container path like `/work/repo` or an unmounted volume would have resolved to `$HOME/work/repo` and injected wrong-repo vocabulary. The fallback now applies ONLY to `../`-prefixed runs (the title itself must signal elision); non-existent absolute paths resolve to nil, full stop. The confirmed Ghostty case is `../`-prefixed, so coverage is unchanged. Also wired the #105 leak-guard exemptions: sanctioned rewrites' `to` values are exempt from the clipboard leak scan (the T5 end-to-end test exercises this — `UserSessionManager.swift` sits exactly at the 24-char leak threshold).

Proof — over-broad re-anchoring vs constrained:

```
before: testNonExistentAbsolutePathDoesNotFallBackToHome     failed (resolved $HOME/work/repo)
        testHomeAnchoredFallbackCandidateShapes              failed (absolute run generated a fallback)
        Executed 17 tests, with 2 failures (0 unexpected)
after:  Executed 17 tests, with 0 failures; all touched suites: Executed 124 tests, with 0 failures
```

Restacked full unit suite on the stack tip: `Executed 835 tests, with 2 tests skipped and 0 failures (0 unexpected)`.

### T6 round 2 (owner re-test): the elision mark is a Unicode ellipsis

The owner's re-test with the fallback build STILL failed — because Ghostty's real elision prefix is `…/` (U+2026 HORIZONTAL ELLIPSIS), not ASCII `../` (a typed report loses the distinction; owner-confirmed). Accepted elided prefixes are now `../`, `…/`, and `‥/` (U+2025). Production wiring was verified sound (`entries()` → `resolveWorkingDirectory` with the real-filesystem default predicate; the miss was purely the regex). The canonical field case is now a named regression test, additionally run through the REAL filesystem predicate against a temp directory tree so hermetic test defaults can never mask production behavior. The no-resolution log line now includes a redacted title SHAPE (letters→a, digits→9, separators/elision marks kept, capped 60) so the next such failure is self-diagnosing without leaking paths.

```
before: testT6GhosttyEllipsisElidedTitleResolvesHomeAnchored           failed
        testT6GhosttyEllipsisElidedTitleResolvesWithRealFilesystemCheck failed
        testTwoDotLeaderElidedTitleResolvesHomeAnchored                 failed
        testEllipsisFallbackCandidateShape                              failed
        Executed 23 tests, with 4 failures (0 unexpected)
after:  Executed 23 tests, with 0 failures; all repo-vocab suites: 53/53
```

Stack rebased onto main (now carrying #116/#117/#118); full unit suite on the stack tip: `Executed 864 tests, with 2 tests skipped and 0 failures (0 unexpected)`.

### DIFF
--- a/Sources/localvoxtral/AppConfigStore.swift
+++ b/Sources/localvoxtral/AppConfigStore.swift
@@ -220,6 +220,14 @@ struct LLMPromptTemplates: Equatable, Sendable {
     private static let userPromptPlaceholderPattern = #"\{\{[a-zA-Z0-9_]+\}\}"#
     private static let splitPlaceholders = ["{{replacement_dictionary}}", "{{input_text}}"]
 
+    /// True when the user template carries the OPTIONAL dictionary placeholder.
+    /// The placeholder is documented as removable; features that ride in that
+    /// slot (repo vocabulary) must check this BEFORE doing any work, because
+    /// `renderTemplate` silently drops the section when the slot is absent.
+    var supportsReplacementDictionary: Bool {
+        userContent.contains("{{replacement_dictionary}}")
+    }
+
     func renderedUserPrompt(
         inputText: String,
         replacementDictionary: String
--- a/Sources/localvoxtral/DictationViewModel+Session.swift
+++ b/Sources/localvoxtral/DictationViewModel+Session.swift
@@ -626,59 +626,113 @@ extension DictationViewModel {
                 let capturedPolishProfile = polishProfile.rawValue
                 let promptTemplates = appConfigStore.loadLLMPromptTemplates(profile: polishProfile)
 
-                var userPrompts = promptTemplates.renderedUserPrompts(
-                    inputText: workingText,
-                    replacementDictionary: replacementDictionaryPrompt
-                )
-                // Opt-in clipboard grounding: prepend ONE reference-context
-                // block to the FINAL user message that lets the polish model
-                // fix near-miss spelling of technical terms against whatever
-                // the user just copied. When the setting is off OR the
-                // polishing endpoint is not loopback, the pasteboard is never
-                // read (privacy).
-                let capturedPolishContextSummary: String?
-                // The excerpt is retained for the post-polish leak guard: the
-                // model output must never smuggle clipboard content into the
-                // committed text.
-                let capturedClipboardExcerpt: String?
-                if let endpointURL = polishingConfig?.endpointURL,
-                   let clipboardContext = polishClipboardContextIfEnabled(endpointURL: endpointURL)
-                {
-                    // Prompt-cache safety: polishd checkpoints ALL-BUT-LAST
-                    // messages as its single-slot prefix cache, so per-request
-                    // context must ride INSIDE the last message — a separate
-                    // context message between prefix and suffix invalidated
-                    // the checkpoint on every request and a cold 4B re-prefill
-                    // blew the polish client timeout (field, 2026-07-11).
-                    // Prepended, never appended: the working text stays LAST
-                    // in the message (this model family echoes instructions
-                    // placed after the input text).
-                    let lastIndex = userPrompts.count - 1
-                    userPrompts[lastIndex] =
-                        PolishContextClipboardReader.contextMessage(
-                            excerpt: clipboardContext.excerpt
-                        ) + "\n\n" + userPrompts[lastIndex]
-                    capturedPolishContextSummary = clipboardContext.provenanceSummary
-                    capturedClipboardExcerpt = clipboardContext.excerpt
-                    Log.polishing.info(
-                        "Polish clipboard context attached: \(clipboardContext.provenanceSummary, privacy: .public)"
-                    )
-                } else {
-                    capturedPolishContextSummary = nil
-                    capturedClipboardExcerpt = nil
-                }
-
-                let polishingRequest = LLMPolishingRequest(
-                    inputText: workingText,
-                    systemPrompt: promptTemplates.systemContent,
-                    userPrompts: userPrompts
-                )
-
                 statusText = StatusStrings.polishing
                 debugLog("LLM polishing started for \(workingText.count) chars")
 
+                // Opt-in clipboard grounding is read HERE, pre-Task, right next
+                // to the payload-macro clipboard read above, so both features
+                // observe the SAME pasteboard state — the repo-vocabulary await
+                // inside the Task can take up to ~2 s, and a copy landing during
+                // that window must not make the context ground against different
+                // text than the payload macro substitutes. When the setting is
+                // off OR the polishing endpoint is not loopback, the pasteboard
+                // is never read (privacy).
+                let capturedClipboardContext: PolishClipboardContext?
+                if let endpointURL = polishingConfig?.endpointURL {
+                    capturedClipboardContext = polishClipboardContextIfEnabled(
+                        endpointURL: endpointURL
+                    )
+                } else {
+                    capturedClipboardContext = nil
+                }
+
+                // Repo vocabulary rides in the `{{replacement_dictionary}}`
+                // slot; a user template without that placeholder (removing it is
+                // explicitly supported) silently drops the section in
+                // renderTemplate, so the whole vocabulary path — AX read, git
+                // subprocess, provenance — is skipped up front when the ACTIVE
+                // template can't carry it.
+                let templateCarriesDictionarySlot =
+                    promptTemplates.supportsReplacementDictionary
+
                 polishAndCommitTask = Task { @MainActor [weak self] in
                     guard let self else { return }
+
+                    // The polish request is assembled HERE, inside the Task, so
+                    // the opt-in repo-vocabulary indexing — whose git subprocess
+                    // runs OFF the main actor with a 2 s timeout — can complete
+                    // before the request is built without stalling the commit. On
+                    // timeout / no repo / feature off it is a fast no-op and the
+                    // request is byte-identical to the no-vocabulary path.
+                    var replacementDictionarySection = replacementDictionaryPrompt
+                    var repoVocabularyCount = 0
+                    // The vocabulary corrections we ASK the model to make must
+                    // also be sanctioned through the token guard, or its
+                    // near-miss repair would deterministically revert them
+                    // (the STT's wrong filename-shaped token is what the guard
+                    // protects). (from: spoken alias, to: exact term) pairs.
+                    var sanctionedVocabularyRewrites: [(from: String, to: String)] = []
+                    if templateCarriesDictionarySlot,
+                       let endpointURL = polishingConfig?.endpointURL,
+                       let vocabularyEntries = await self.repoVocabularyEntriesIfEnabled(
+                           endpointURL: endpointURL,
+                           transcript: workingText
+                       )
+                    {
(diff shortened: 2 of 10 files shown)


T0mSIlver:
Pushed a second commit (3194657) with post-review hardening of the vocabulary pipeline, found by a Fable deep review of the F6 diff (#115) that covered this branch's code too:

- **Deadline race (new)**: the detached title→cwd→index→match pipeline is raced against a 3 s deadline via a resume-once continuation. Rationale: the git wait was already bounded, but a `fileExists` stat on a stale network mount during cwd resolution can block indefinitely — and the polish task awaited it, wedging that session's commit at "Polishing…". A task group can't express this (it awaits all children); the losing pipeline is abandoned (it only returns values; its shared side effects are the Mutex-guarded TTL cache and the gate release).
- **Single-flight gate (review P2)**: an abandoned pipeline pins one cooperative-pool thread; without a gate, every subsequent commit against the same wedged mount stacked another until the pool — and the deadline mechanism itself — starved. `RepoVocabularyFlightGate` caps the cost at one thread; while a pipeline is in flight, commits fast-skip vocabulary.
- **Priority (review P3)**: the deadline task runs `.userInitiated` (timeliness is its job); the pipeline deliberately stays `.utility`.
- New DEBUG seams (pipeline + deadline sleep, injected-sleep pattern) and two regression tests: `testVocabularyPipelineDeadlineProceedsWithoutVocabulary` and `testWedgedPipelineSingleFlightSkipsNextCommit` (deterministic via a start signal, no wall-clock).

Full suite: **761 tests, 2 skipped, 0 failures, 0 warnings** (this branch's tip builds standalone; CI will confirm).

PR #128, opened 2026-07-12
Resolve repo vocabulary from terminal processes

T0mSIlver:
## What

Fix repo vocabulary when a terminal program replaces the focused window title (Claude Code, Codex, opencode, editors, and other OSC 0 users).

The focused Accessibility window title remains tier 1 because it can disambiguate a focused tab. When it contains no usable git path, tier 2 now:

- snapshots the terminal app's complete descendant process tree with `sysctl(KERN_PROC_ALL)`;
- reads each descendant CWD with `proc_pidinfo(PROC_PIDVNODEPATHINFO)` off the main actor and inside the existing deadline/single-flight machinery;
- maps every CWD through the existing git-root walker; and
- accepts the fallback only when every descendant maps to the same canonical repo.

The fallback fails closed for multiple repos, non-repo descendants, unreadable CWDs, or process races. It is enabled only for built-in or user-configured terminal bundle IDs, so IDE/build-helper descendants cannot become a repo signal. Process snapshot and CWD reads are injectable; unit tests never inspect live processes.

The README now describes title-independent discovery and the deliberate ambiguity skip.

`[run-llm-eval]`

## Root cause

Repo discovery was hard-gated on extracting `/…`, `~/…`, or an elided path from the focused terminal title. Coding agents commonly replace that title with an app/task name and may not restore it, leaving no path both during and after the agent session.

## Proof

- [x] Mac capability probe before implementation — a non-root, ad-hoc hardened-runtime executable (`flags=0x10002(adhoc,runtime)`) successfully read a sibling same-UID process CWD: `bytes=2352 cwd=/private/var/folders/.../localvoxtral-proc-cwd-probe-...`.
- [x] Bug fix regression failed before the fix — `./scripts/remote-build.sh test --filter RepoVocabularyIndexerEndToEndTests.testTitleClobberedAgentResolvesRepoFromTerminalDescendant`:
  `XCTAssertEqual failed: ("nil") is not equal to ("Optional("useAuth.ts")")`; 1 test, 1 failure.
- [x] The same regression passes after the fix — 1 test, 0 failures in 0.197 s.
- [x] Final-base unit suite green — `./scripts/remote-build.sh`: 887 tests, 2 expected skips, 0 failures in 10.954 s.
- [x] PolishHelper unit suite green — `./scripts/remote-build.sh test --package-path PolishHelper`: 45 tests, 0 failures.
- [x] Realtime integration green — `./scripts/remote-build.sh integration`: 7 tests, 2 expected mic skips, 0 failures; live synthetic-audio accuracy test passed.
- [x] Packaging green — `./scripts/remote-build.sh package`: release app and bundled `localvoxtral-polishd` built; `codesign --verify --deep --strict` passed.
- [x] Required final-base live-model lane green — `./scripts/remote-build.sh integration-polishd`: 6 tests, 0 failures in 86.819 s. Production standard baseline: required 7/7, known-hard 10/10, technical 2/17. Agent profile: required 1/1, known-hard 6/8.
- [ ] GUI hand test with a real title-clobbered terminal — not run: the SSH build gate has no interactive GUI/AX session. The live hardened-runtime CWD capability was probed directly, and the full title-clobbered pipeline is covered end to end with injected process seams plus real git.

## Safety / tradeoffs

The terminal PID does not identify a focused tab. There is no sound generic mapping from an AX terminal window to one PTY across Terminal, iTerm2, Ghostty, Warp, and others, so the implementation intentionally does not rank descendants. Mixed repo/non-repo tabs, multiple repos, unreadable descendants, reparented multiplexer processes, and short process-exit races may produce false negatives. They cannot select one visible repo over another.


T0mSIlver:
Rebase + final proof (post #139/#142/#143 merges):

- Rebased onto main; only conflict was README.md. Resolution reconciles the repo-vocabulary bullet with #143's grounding pre-apply (the "Hints only, never automatic replacement" claim was stale on main): discovery text from this PR kept, tail now reads "High-confidence, boundary-checked matches are corrected in the working text before the polish request; everything else stays a hint". The stale token-guard bullet from this branch's original text was NOT resurrected.
- Independent review (read-only): merge-ready, no blockers; fail-closed paths verified, semantic composition with #143's grounded matcher explicitly checked (descendant-discovered git root flows into `groundedCandidateEntries`).
- Reviewer-requested regression tests added (`testCyclicPPIDDataTerminates`, `testSymlinkAliasCWDsCanonicalizeToUnique`, `testEmptySnapshotFailsClosedAtEntriesLevel`) — 11/11 focused, and the full unit suite on the rebased head: **Executed 988 tests, with 2 tests skipped and 0 failures** (remote Mac).
- CI on the rebased head: all green including tier-1 live voxmlx integration and the polishd live-model lane (auto-triggered by the `[run-llm-eval]` marker).

Note (documented tradeoff, from review): post-#143, a unanimously-wrong descendant repo resolution feeds deterministic pre-apply rather than a mere hint — kept safe by the unanimity gate + the matcher's boundary checks and abstention margins; worth remembering when calibrating fallback thresholds.

PR #129, opened 2026-07-12
Existing installs adopt updated bundled config defaults (fixes the #126 known limitation)

T0mSIlver:
## What

Fixes the known limitation surfaced in #126: `ensureConfigFilesExist` seeds bundled config files only when absent, so **existing installs never picked up improved bundled defaults** (the #126 prompt trim reached fresh installs only).

Mechanism — `AppConfigStore.reconcileBundledDefaults()`, run once at launch:

- **Unedited stale seeds refresh silently.** `BundledConfigDefaultHistory` registers the SHA-256 of every default config version ever shipped on main (all 12 historical versions, derived from git history and independently re-verified). A user file matching any of them is an unedited seed — replacing it loses nothing, so it's refreshed in place and logged.
- **Customized files are never touched without asking.** Unknown content gets one alert (after launch, and held until the onboarding wizard closes if one is due): **Update (Keep Backups)** replaces the files and saves the user's versions alongside as `<name>.toml.backup-<timestamp>` (collision-safe); **Keep Mine** records the decision in a hidden `.bundled-defaults-state.json` sidecar so the prompt never repeats for the same default version — it returns only when the defaults change again; **Show Files…** reveals the files in Finder and re-presents the alert, so Update/Keep Mine stay available while you inspect (quitting instead still re-prompts next launch).
- **The replacement dictionary is exempt.** It's the user's own rule set, not a tunable default — the format has never changed since shipping, so reconcile never refreshes it and never prompts for it (regression-tested).
- **Refreshes are once-per-version too**: a silent refresh records the bundled hash it handled, so a user who deliberately restores an old shipped default afterwards keeps it — no refresh loop.
- A hash missing from the history table degrades safely: the file is treated as customized and prompted, never clobbered. A new guard-rail test fails with the exact hash to paste whenever a bundled TOML changes without a matching history entry, and it enumerates files through a `#if DEBUG` seam on the private `ConfigFile` enum so it can't drift. AGENTS.md documents the rule; README documents the user-facing behavior.

Reviewed by a Fable subagent (adversarial pass over hash table, sidecar state machine, strict concurrency, AGENTS.md rules): all 12 historical hashes independently re-derived from git and confirmed; its three substantive findings (refresh loop on deliberate restores, alert stacking on the onboarding wizard for pre-onboarding installs, unguarded test file list) are fixed in the second commit, each with a regression test where testable.

## Proof

- **New suite `AppConfigDefaultsReconcileTests` (9 tests, all passing on the build host)**: current-bundle-registered guard rail; stale seed refreshed to bundled bytes; deliberate restore of an old default sticks after one refresh; customized file reported and untouched; fresh seed reports nothing; adopt backs up + replaces + stops prompting; adopt touches only requested files; Keep Mine suppresses re-prompt for the same version; Keep Mine re-prompts when the default changes again.
- **Tier 0 full suite**: 890 tests, 2 env-gated skips, 0 failures in the reconcile/config area. (`BackendProcessSupervisorTests.testStopHonorsTERMWithoutRestarting` flaked twice while the Record README Demo workflow was hammering the host — it races a real `/bin/sh` `trap TERM` install against supervisor shutdown, passes in isolation, and is untouched by this diff; clean full-suite rerun after the workflow finished: **890 tests, 2 env-gated skips, 0 failures** — supervisor test included.)
- **LLM lanes skipped — justification**: no prompt content, model pins, sampling, request shape, or eval corpus changes. This PR changes only *which file on disk* holds the (already-evaled, #126) bundled default content and when it's copied there; the polish request path is untouched.
- **Hand-testing (UI)**: the alert flow is new UI and has NOT been hand-tested — the reconcile/adopt/keep logic behind each button is fully unit-tested against the real bundle and real file IO, but the NSAlert itself needs one `try-pr.sh` pass on a real install (I didn't want to mutate the live config folder on the build host). Steps: edit `llm_system_prompt.toml` under `~/Library/Application Support/localvoxtral/config` (any byte), launch, and exercise each button — Update creates the `.backup-<timestamp>` file and swaps in the new default; Keep Mine writes `.bundled-defaults-state.json` and stays quiet on relaunch; Show Files reveals the files in Finder and the alert re-presents immediately with Update/Keep Mine still available. Bonus check for the silent path: an *unedited* pre-#126 install's prompt files should just get the trimmed defaults with a `Refreshed unedited default…` line in the config log.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Open **Settings** from the menu bar popover:
 - **Text Processing** — exact-match replacements, plus the LLM Polishing switch and its agent-dictation features (agent prompt profile, repo vocabulary, clipboard context, spoken clipboard paste)
 - **About** — version, link to this repository, and Export Diagnostics (writes a redacted local report to the Desktop)
 
-The shared config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`. When an update ships improved defaults, files you haven't edited are refreshed automatically; files you have edited are never touched without asking — the app offers to update them and keeps your versions as `.backup` files alongside.
+The config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`. When an update ships improved defaults, files you haven't edited are refreshed automatically; files you have edited are never touched without asking — the app offers to update them and keeps your versions as `.backup` files alongside.
 
 ## Under the hood
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Open **Settings** from the menu bar popover:
 - **Text Processing** — exact-match replacements, plus the LLM Polishing switch and its agent-dictation features (agent prompt profile, repo vocabulary, clipboard context, spoken clipboard paste)
 - **About** — version, link to this repository, and Export Diagnostics (writes a redacted local report to the Desktop)
 
-The shared config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`.
+The shared config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`. When an update ships improved defaults, files you haven't edited are refreshed automatically; files you have edited are never touched without asking — the app offers to update them and keeps your versions as `.backup` files alongside.
 
 ## Under the hood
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -101,7 +101,7 @@ Open **Settings** from the menu bar popover:
 - **Text Processing** — exact-match replacements, plus the LLM Polishing switch and its agent-dictation features (agent prompt profile, repo vocabulary, clipboard context, spoken clipboard paste)
 - **About** — version, link to this repository, and Export Diagnostics (writes a redacted local report to the Desktop)
 
-The shared config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`.
+The shared config folder at `~/Library/Application Support/localvoxtral/config` holds `replacement_dictionary.toml`, the LLM prompt TOMLs (including the agent variants), and `terminal_apps.toml`. When an update ships improved defaults, files you haven't edited are refreshed automatically; files you have edited are never touched without asking — the app offers to update them and keeps your versions as `.backup` files alongside.
```
T0mSIlver: remove "shared", it's just a config folder

T0mSIlver: Done — dropped "shared" in 8e4b9b9.

PR #133, opened 2026-07-13
Fresh installs get the tap/hold gesture as their trigger

T0mSIlver:
## What

A new user landed on the `⌥Space` keyboard shortcut and had to go find the tap/hold gesture in Settings — even though it's the app's headline interaction (and what the README demo shows). Fresh installs now get `modifierOnlyHotKeyEnabled` seeded on.

Two things worth calling out, both discovered while building this:

**The seed is written, not left to the load fallback.** The first launch persists the backend-mode keys, so by the *second* launch the install no longer looks fresh — a fallback-based seed would silently flip the gesture back off. `testFreshInstall_seedSurvivesTheSecondLaunch` pins this.

**Polishing is deliberately NOT seeded, because it's already on by default.** The onboarding wizard enables it (`polishingConsent` defaults `true`) *together with* downloading the model. Seeding `llmPolishingEnabled` would have been actively harmful: the decline path never writes `false`, it relies on the fallback — so a seeded `true` would survive as polishing-on with no model on disk, and the first overlay commit would try to polish against nothing. `testFreshInstall_leavesPolishingToTheOnboardingWizard` guards it.

Repo vocabulary and clipboard-as-context stay opt-in (they feed the user's surroundings to the polisher), so the README's privacy copy stays accurate.

**Existing installs are untouched.** The load fallback stays `false`, which is what an existing install reads, so an update never claims Right Command behind the user's back. Two guard tests cover both detection paths (completed onboarding; legacy backend-mode key).

## Proof

New suite `SettingsStoreOutOfBoxDefaultsTests` (7 tests). Before the fix — the three fresh-install tests fail, and the guards already pass (which is what makes them guards):

```
testFreshInstall_enablesTheModifierGesture                        → failed
testFreshInstall_seedSurvivesTheSecondLaunch                      → failed
testFreshInstall_neverOverwritesAChoiceTheUserAlreadyMade         → failed
testFreshInstall_leavesTheSurroundingsFeaturesOptIn               → passed
testExistingInstall_thatCompletedOnboarding_keepsTheGestureOff    → passed
testExistingInstall_detectedByLegacyBackendModeKey_keepsTheGestureOff → passed
```

After the fix, full unit suite on the Mac build host:

```
Executed 898 tests, with 2 tests skipped and 0 failures (0 unexpected) in 10.893 seconds
```

Six existing tests changed. All six were fixtures that *implicitly* relied on the gesture being off — their own comments already said "no gesture" or "Fresh defaults: keyboard shortcuts". Each now states its premise explicitly (`modifierOnlyHotKeyEnabled = false`); no assertion was removed or weakened. `testSummary_keyboardDefault_showsOverlayShortcut` is renamed to `testSummary_keyboardShortcuts_showsOverlayShortcut` — it exercises keyboard-shortcut mode, which is no longer the default.

## Hand-testing

Not yet run — the UI-visible effect is that a fresh install's Settings → Dictation shows **Single modifier key / Right Command** instead of Keyboard shortcuts. Worth confirming on a real fresh install alongside the 0.8.0 upgrade testing.

LLM lanes: not required — no prompt, model pin, sampling, or polish-request change. This does not alter what reaches the model.

### DIFF
--- a/Sources/localvoxtral/SettingsStore.swift
+++ b/Sources/localvoxtral/SettingsStore.swift
@@ -435,15 +435,22 @@ final class SettingsStore {
         // keys were *already* stored, so it must observe the untouched domain.
         let resolvedOnboardingCompleted = Self.resolveOnboardingCompleted(
             defaults: defaults, environment: environment)
+        // Seed the out-of-box defaults on a never-launched install only, and
+        // WRITE them rather than express them as a load fallback below: once
+        // the wizard completes it persists onboarding as done, this install
+        // stops looking fresh, and a fallback would silently flip the seeded
+        // values back off.
+        //
+        // The gate needs the onboarding key to be ABSENT, not merely resolved
+        // false. "Re-run Setup…" resets that flag to false on an existing
+        // install (DictationViewModel.reRunOnboarding), so a crash or force-quit
+        // before the wizard closes would otherwise leave a configured user
+        // looking fresh on the next launch — and claim Right Command from them.
+        let isNeverLaunchedInstall = defaults.object(forKey: Keys.onboardingCompleted) == nil
         onboardingCompleted = resolvedOnboardingCompleted
         defaults.set(resolvedOnboardingCompleted, forKey: Keys.onboardingCompleted)
 
-        // A fresh install is the only place the out-of-box defaults are seeded,
-        // and they are WRITTEN rather than left to the load fallbacks below:
-        // the migration a few lines down persists the backend-mode keys, so by
-        // the second launch this install no longer looks fresh and a fallback
-        // would silently flip them back off.
-        if !resolvedOnboardingCompleted {
+        if isNeverLaunchedInstall, !resolvedOnboardingCompleted {
             Self.seedFreshInstallDefaults(defaults: defaults)
         }
 
--- a/Sources/localvoxtral/SettingsStore.swift
+++ b/Sources/localvoxtral/SettingsStore.swift
@@ -438,6 +438,15 @@ final class SettingsStore {
         onboardingCompleted = resolvedOnboardingCompleted
         defaults.set(resolvedOnboardingCompleted, forKey: Keys.onboardingCompleted)
 
+        // A fresh install is the only place the out-of-box defaults are seeded,
+        // and they are WRITTEN rather than left to the load fallbacks below:
+        // the migration a few lines down persists the backend-mode keys, so by
+        // the second launch this install no longer looks fresh and a fallback
+        // would silently flip them back off.
+        if !resolvedOnboardingCompleted {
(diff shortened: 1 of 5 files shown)


T0mSIlver:
## opencode review (GLM-5.2, read-only) — one real bug found and fixed

Full verdicts: freshness heuristic **concern** (below), init ordering **sound**, polishing-not-seeded reasoning **sound**, launch-warmup interaction **sound**, test changes **sound** (no assertion weakened; the removed `XCTAssertFalse(store.modifierOnlyHotKeyEnabled)` codified the *old* default and its replacement lives in the new suite).

### The bug (fixed in 6dc86f9)

The seed gate was `!resolvedOnboardingCompleted`, which conflates *"never launched"* with *"previously completed, now re-running"*.

`DictationViewModel.reRunOnboarding()` (`DictationViewModel.swift:1707`) writes `onboardingCompleted = false` on an existing, configured install. A graceful window close completes onboarding again — but a **crash or force-quit before the wizard closes** leaves the flag present-and-false. On the next launch `resolveOnboardingCompleted` returns false, the install reads as fresh, and the seed fires. An existing user who never asked for the gesture gets Right Command claimed, and because `registerCurrentHotKeys()` is exclusive, their `⌥Space` shortcut stops being registered.

The `object(forKey:) == nil` guard doesn't save them: a user who left the gesture at its old default has the key absent, so they look identical to a fresh install.

Fix: require the onboarding key to be **absent**, not merely resolved false.

```swift
let isNeverLaunchedInstall = defaults.object(forKey: Keys.onboardingCompleted) == nil
if isNeverLaunchedInstall, !resolvedOnboardingCompleted { … }
```

### It also caught a flaw in my reasoning

My commit message claimed the backend-mode migration is what makes an install stop looking fresh on the second launch. Wrong — `resolveOnboardingCompleted` checks the onboarding key *first*, and launch 1 persists it as `false`, so a fresh install still looks fresh on launch 2. What actually flips it is `completeOnboarding()` writing `true`.

The conclusion (write the seed, don't use a load fallback) survives — a fallback would flip the gesture off the moment the wizard completes. But `testFreshInstall_seedSurvivesTheSecondLaunch` was passing by *idempotency*, not by exercising the distinction: it would have passed against a fallback-based implementation too. It now completes the wizard between the two launches, which is the case a fallback actually breaks.

## Proof

New regression test `testExistingInstall_interruptedWhileReRunningSetup_keepsTheGestureOff`, against the pre-fix gate:

```
testExistingInstall_interruptedWhileReRunningSetup_keepsTheGestureOff → failed
(Executed 8 tests, with 2 failures)
```

With the fix, full unit suite:

```
Executed 899 tests, with 2 tests skipped and 0 failures (0 unexpected) in 11.702 seconds
```

PR #135, opened 2026-07-13
mlx-lm cleanup: remove the bare console script, not just the dotted entry points

T0mSIlver:
## What

Found by hand-testing the 0.7.4 → 0.8.0 upgrade path (the one thing CI can't cover — this is exactly what it was for).

`uv tool install mlx-lm` aborted during 0.7.4's onboarding with:

```
error: Executable already exists mlx_lm (use --force)
```

The install's `backends/bin/` held a **dangling** `mlx_lm` symlink pointing into the long-deleted `tools/mlx-lm/bin/mlx_lm`:

```
lrwxr-xr-x  1 tom  staff  84 Jul  9 00:17 mlx_lm -> …/backends/tools/mlx-lm/bin/mlx_lm
lrwxr-xr-x  1 tom  staff  84 Jul  5 11:37 voxmlx -> …/backends/tools/voxmlx/bin/voxmlx
```

## Cause

The wheel installs a **bare `mlx_lm`** console script alongside the dotted ones (`mlx_lm.server`, `mlx_lm.generate`, …). `LegacyMLXLMCleanup.legacyBinPrefix` was `"mlx_lm."` — *with the dot* — so `"mlx_lm".hasPrefix("mlx_lm.")` is `false`. Cleanup removed the venv, the wheels, the marker entry and every dotted sibling, then walked straight past the one entry that remained.

Every install upgraded to 0.8.0 carries that orphan. Nothing reads it, so it is inert — but the cleanup claims to remove the install completely and doesn't, and it breaks any subsequent (re)install of mlx-lm.

**Why the tests missed it:** their fixtures only ever created `mlx_lm.server` / `mlx_lm.generate` — a `bin/` layout no real install has. The suite was green against an imaginary directory.

## Fix

Drop the dot. The undotted prefix covers both shapes. `voxmlx*` neighbours are unaffected, and there's now an assertion pinning that.

**Checked the rest of the tree:** `mlx_lm.server` also appears in `ui-smoke.sh`, `mac-diag.sh` and `scripts/mac/localvoxtral-build-gate.sh`. Those are **process-name** patterns, not filesystem prefixes, and that process is still live — the eval launchd service runs `mlx_lm.server` (`scripts/mac/README.md:109`). Correct as-is; changing them would break the eval-server diagnostics. The defect is confined to the one prefix.

## Proof

Both new/updated tests fail against the shipped code and pass with the fix.

Before (shipped `"mlx_lm."` prefix):

```
testBareConsoleScriptLeftByAnEarlierPassIsStillRemoved → failed
testRemovesOrphanedMLXLMPiecesAndKeepsEverythingElse   → failed
testDanglingBinSymlinkAloneIsStillRemoved              → passed
testMarkerWithoutLegacyEntryIsLeftByteIdentical        → passed
testSecondRunAndMissingRootAreSilentNoOps              → passed
```

After, full unit suite on the Mac build host:

```
Executed 900 tests, with 2 tests skipped and 0 failures (0 unexpected) in 11.974 seconds
```

Cleanup stays idempotent (`testSecondRunAndMissingRootAreSilentNoOps`), so an install already cleaned by 0.8.0 gets the leftover shim removed on its next launch — no user action needed.

LLM lanes: not required — no prompt, model pin, sampling, or polish-request change.

## Ships in

0.8.1.

### DIFF
--- a/Sources/localvoxtral/Backends/LegacyMLXLMCleanup.swift
+++ b/Sources/localvoxtral/Backends/LegacyMLXLMCleanup.swift
@@ -15,8 +15,14 @@ struct LegacyMLXLMCleanup {
 
     /// uv tool venv directory name (`tools/mlx-lm`) and `installed.json` key.
     private static let legacyToolID = "mlx-lm"
-    /// Entry points uv linked into `bin/` (`mlx_lm.server`, `mlx_lm.generate`, …).
-    private static let legacyBinPrefix = "mlx_lm."
+    /// Entry points uv linked into `bin/`. The wheel installs a BARE `mlx_lm`
+    /// console script alongside the dotted ones (`mlx_lm.server`,
+    /// `mlx_lm.generate`, …). An earlier `"mlx_lm."` prefix here removed the
+    /// dotted siblings and walked straight past the bare one, leaving it behind
+    /// as a dangling symlink — and making a later `uv tool install mlx-lm` fail
+    /// with "Executable already exists". Found by hand-testing the 0.7.4 → 0.8.0
+    /// upgrade; the undotted prefix covers both shapes.
+    private static let legacyBinPrefix = "mlx_lm"
     /// Wheels the installer parked in `downloads/`.
     private static let legacyWheelPrefix = "mlx_lm-"
 
--- a/Tests/localvoxtralTests/LegacyMLXLMCleanupTests.swift
+++ b/Tests/localvoxtralTests/LegacyMLXLMCleanupTests.swift
@@ -28,6 +28,16 @@ final class LegacyMLXLMCleanupTests: XCTestCase {
         try fileManager.createDirectory(at: layout.toolBin, withIntermediateDirectories: true)
         // uv links bin entries into the venv; after the venv is gone these
         // dangle, which fileExists-based checks would miss.
+        //
+        // The wheel installs a BARE `mlx_lm` console script alongside the dotted
+        // entry points. Fixturing only the dotted names is what let the bare one
+        // ship unremoved: the old `"mlx_lm."` prefix deleted its siblings and
+        // left it behind.
+        let consoleScriptLink = layout.toolBin.appendingPathComponent("mlx_lm")
+        try fileManager.createSymbolicLink(
+            at: consoleScriptLink,
+            withDestinationURL: legacyVenv.appendingPathComponent("bin/mlx_lm")
+        )
         let serverLink = layout.toolBin.appendingPathComponent("mlx_lm.server")
         try fileManager.createSymbolicLink(
             at: serverLink,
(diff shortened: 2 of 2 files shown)


T0mSIlver:
## opencode review (GLM-5.2, read-only) — no defects found

All five areas verdict **sound**; both of my claims were independently checked and confirmed.

**Prefix safety.** The match is scoped to `toolBin` under `BackendInstallLayout.root` (`BackendInstallLayout.swift:10-17`). Everything else legitimately in that directory was enumerated: `voxmlx-serve` (doesn't match), `uv` (lives in `root/uv/<version>/`, not `toolBin`), `localvoxtral-polishd` (ships in `Contents/MacOS/` of the .app, unreachable), Python (`root/python`). No managed install emits an executable starting with `mlx_lm`. Over-matching is theoretical, not real.

**Completeness.** `legacyToolID` (`mlx-lm`, hyphenated) and `legacyWheelPrefix` (`mlx_lm-`) never shared the dotted bug. `legacyBinPrefix` has exactly one reference. `BackendInstaller`, `BackendManager`, `BackendCatalog` and `HFModelDownloader` confirmed to contain no other mlx-lm filesystem cleanup.

**The script references — claim verified.** All three (`ui-smoke.sh:133`, `mac-diag.sh:124`, `localvoxtral-build-gate.sh:127`) feed `pgrep -f`, i.e. process-cmdline matches, not filesystem prefixes. The process is live: the eval launchd service runs `~/.local/share/localvoxtral-eval/mlx-lm/bin/mlx_lm.server` (`scripts/mac/README.md:109`) — a path **outside** the cleanup's layout root, so the cleanup could not touch it even if it tried. Correctly left alone.

**Idempotency for users already on 0.8.0 — claim verified.** Traced each step for the field state (venv/wheel/dotted-entries/marker already gone, bare shim surviving): `tools/mlx-lm` no-ops via `NSFileNoSuchFileError`, the `toolBin` scan finds and removes the bare shim, `downloads/` no-ops, and `dropLegacyInstalledMarkerEntry` returns false so `installed.json` is left byte-identical. The leftover is cleared on the next launch with no user action.

**Tests.** Confirmed the wheel's `[project.scripts]` declares a bare `mlx_lm` CLI *plus* the dotted subcommands — so every real install has all three, and the old dotted-only fixture was the fiction that let this ship. No assertion weakened; all changes additive or stricter (count 5 → 6, plus a new symlink-cleared assertion). Confirmed the new regression test genuinely fails under the old prefix.

**Error handling / logging.** Dangling-symlink handling correct (`contentsOfDirectory` lists by name; `removeItem` unlinks rather than resolves). `removeIfPresent` swallows only `NSFileNoSuchFileError` and logs everything else without rethrowing, so cleanup never blocks launch. Deletions logged at `.notice` per the AGENTS.md "keep backend paths loud" rule.

> Verdict: *"a textbook one-line correctness fix with the right surrounding comment, a faithful regression test for the exact field state, and stricter (not weakened) assertions. Ship it."*

T0mSIlver:
## Field evidence: the real `bin/` after a 0.7.4 install

Captured on a Mac by installing v0.7.4 and letting onboarding install the mlx-lm backend — this is exactly what every pre-0.8.0 user has:

```
mlx_lm                  mlx_lm.convert          mlx_lm.generate         mlx_lm.server           voxmlx-serve
mlx_lm.awq              mlx_lm.dwq              mlx_lm.gptq             mlx_lm.share
mlx_lm.benchmark        mlx_lm.dynamic_quant    mlx_lm.lora             mlx_lm.upload
mlx_lm.cache_prompt     mlx_lm.evaluate         mlx_lm.manage           voxmlx
mlx_lm.chat             mlx_lm.fuse             mlx_lm.perplexity       voxmlx-convert
```

**18 mlx-lm entries: 1 bare + 17 dotted.**

- Old prefix `"mlx_lm."` → matches the 17 dotted, misses the bare `mlx_lm`. That single survivor is exactly what was found dangling in `backends/bin/` on the upgraded install, and what made `uv tool install mlx-lm` abort with "Executable already exists".
- New prefix `"mlx_lm"` → matches all 18.
- `voxmlx`, `voxmlx-convert`, `voxmlx-serve` → match neither. The undotted prefix does not reach its neighbours, which is the property the new regression test asserts.

Confirms the cause, the fix, and the blast radius (every upgraded install carried one orphaned dangling symlink).

PR #141, opened 2026-07-15
Spike: Swift Voxtral engine reaches parity with Python voxmlx (do not merge)

T0mSIlver:
## What

**Spike — do not merge.** Recorded for the decision trail. Answers a single question: can a bundled Swift engine replace the managed Python `voxmlx` backend, at acceptable accuracy and realtime headroom?

**Answer: yes.** A vendored + optimized [`Blaizzy/mlx-audio-swift`](https://github.com/Blaizzy/mlx-audio-swift) `VoxtralRealtime` reaches **parity with the Python backend** — same audio, same 80 ms chunking, same harness, same GPU, back-to-back with the Python server as the contention normalizer:

| engine | RTF | vs Python (same process) | word accuracy | peak GPU |
|---|---|---|---|---|
| stock mlx-audio-swift | 1.84 | 3.35× slower | 0.804 | 5.1 GB |
| **optimized** | **0.62** | **~1.0×** | **0.804** (0.842 FR) | **3.3 GB** |

The stock upstream engine is **~3.2× slower than Python and cannot keep up with live dictation** (RTF ~1.85). The optimized engine keeps up with ~1.6× headroom, at bit-identical accuracy and 1.8 GB less memory. It also unlocks `transcriptionDelayMs` (the value the Python backend hardcodes at 480 ms), swept here from 480→80 ms.

### The bug behind the whole gap

A **float32 leak** with two sources that both had to be plugged:

1. `AdaptiveNorm` — `x * (1.0 + adaScale)` with a float32 `adaScale`, promoting the decoder hidden state to float32 in all 32 layers.
2. The manual RoPE — builds cos/sin in float32 and re-promotes even after (1) is fixed.

Consequence: the tied fp16 embedding (131072×3072, ~805 MB) was **upcast on every token**, and quantized matmuls fell off their fast path ([mlx-swift#420](https://github.com/ml-explore/mlx-swift/issues/420)). `voxmlx` casts (`t_cond.astype(x.dtype)`); this port didn't. Fixing both: logits **73.5 → 4.8 ms/token**, decoder forward 19.6 → 13.1, RTF **1.84 → 0.62**.

Fixing either one alone gives ~0% — which is why a naïve one-change-at-a-time bisect hid the cause.

### Other changes (small individually)

Fused `MLXFast.RoPE` (also fixes a `[H,L,D]` layout that hit [mlx-swift#441](https://github.com/ml-explore/mlx-swift/issues/441): Metal RoPE garbage for single-token decode with batch>1 → empty transcripts), async-eval one-step-lag decode, dtype-correct attention masks, `asLinear` LM head, and not dumping the Metal buffer pool every chunk.

### Layout

- `spikes/VoxtralSpike/` — throwaway harness; drives the online streaming session exactly as the app feeds the mic, scores with the tier-1 word-accuracy method, sweeps chunk size / delay / engine flags, and streams the same audio to the live Python `voxmlx` as a same-process baseline.
- `spikes/VoxtralSpike/Sources/VoxtralFast/` — the vendored engine (MIT) with the optimizations behind `VOXFAST_*` flags for bisecting.
- `.github/workflows/voxtral-spike.yml` — self-hosted lane (needs xcodebuild + a real 4B model; the SSH gate allowlist can't run it).

## Proof

Not a shippable behavior change — a spike. Evidence is the CI run matrix, all on the self-hosted Mac against `say`-synthesized tier-1 audio:

- Final confirmation: [run 29351575457](https://github.com/T0mSIlver/localvoxtral/actions/runs/29351575457) — optimized RTF 0.62 @ 0.804 EN / 0.842 FR, 0 delta re-emits, delay sweep 480→80 ms.
- Bisect that isolated the leak: [run 29350934524](https://github.com/T0mSIlver/localvoxtral/actions/runs/29341979759) (all-off = stock 0.804; all-on = 0.62 @ 0.804).
- RoPE numeric self-test: MATCH (max abs dev 0.001 vs the manual RoPE it replaces).

Follow-ups (next PRs): vendor the engine into the app properly with the fixes always-on + MIT provenance + unit tests (RoPE equivalence, delta-emission regression), then a bundled `speechd` helper speaking the Realtime websocket protocol to replace the `voxmlx` process. Upstream fix for the float32 leak is being prepared separately.

- [ ] N/A — spike, not for merge

### DIFF
--- a/.github/workflows/voxtral-spike.yml
+++ b/.github/workflows/voxtral-spike.yml
@@ -1,0 +1,57 @@
+name: Voxtral Spike
+
+# THROWAWAY SPIKE lane. Answers one question before we commit to replacing the
+# managed Python voxmlx backend with a bundled Swift engine: does mlx-audio-swift's
+# VoxtralRealtime streaming session transcribe accurately, keep up with realtime,
+# and emit clean deltas (no re-emits — our insertion path has no backspaces)?
+#
+# Runs on the self-hosted Mac because it needs xcodebuild (SwiftPM's CLI cannot
+# compile mlx-swift's Metal kernels) and a real 4B model on the GPU.
+# Delete this workflow, spikes/, once the go/no-go is recorded.
+
+on:
+  push:
+    branches: ['spike/**']
+  workflow_dispatch:
+    inputs:
+      model_repo:
+        description: 'HF repo for the Voxtral realtime weights'
+        default: 'mlx-community/Voxtral-Mini-4B-Realtime-6bit'
+        required: false
+
+jobs:
+  spike:
+    runs-on: [self-hosted, macOS, ARM64]
+    timeout-minutes: 45
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Run Voxtral streaming spike
+        env:
+          MODEL_REPO: ${{ github.event.inputs.model_repo || 'mlx-community/Voxtral-Mini-4B-Realtime-6bit' }}
+        run: ./spikes/run-spike.sh
+
+      - name: Summarize
+        if: always()
+        run: |
+          {
+            echo '## Voxtral Swift spike'
+            for f in spikes/out/en.json spikes/out/fr.json; do
+              [[ -f "$f" ]] || continue
+              echo "### $f"
+              echo '```json'
+              cat "$f"
+              echo '```'
+            done
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload results
+        if: always()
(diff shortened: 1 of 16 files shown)


T0mSIlver:
Closing as recorded: do-not-merge spike, kept for the decision trail. The engine + fixes landed via #142 (part 1 of the vendoring series); the branch spike/voxtral-swift stays for reference. Parity evidence: RTF 0.62 vs Python 0.62 at word accuracy 0.804 EN / 0.842 FR (run 29351575457).

PR #150, opened 2026-07-17
Add Claude Code and Ghostty context for agent dictation

T0mSIlver:
> Stacked on [#152](https://github.com/T0mSIlver/localvoxtral/pull/152), which contains persistent-runner cleanup, exact process-group teardown, and the bounded file-backed XCTest supervisor. This PR's base is temporarily `codex/ci-xctest-cleanup` so the feature diff remains separate.

## What

Adds a budgeted “agent context” pipeline for dictation aimed at Claude Code in Ghostty, including local and SSH sessions.

- Captures the full visible Ghostty screen through Accessibility when the focused pane is positively joined to a live Claude session. Screen reads are Ghostty-only, local-polish-only, sanitized, revalidated at stop, and withheld unless the broker authorizes the pane/session join.
- Ships a Claude Code marketplace with local and remote plugins. Hooks publish prompt/session/tool facts; the local plugin uses a private Unix socket and the remote plugin uses declarative HTTP over an SSH `RemoteForward` tunnel.
- Uses OSC window-title markers to identify the focused Claude session, while hooks provide the off-screen facts a status line cannot: latest user prompt, state, cwd/workspace, recent files, and bounded remote tool snippets.
- Harvests local repository context (branch, status, staged/unstaged diff, tracked paths, active files, transcript-matched files) under hard time/byte/file caps. Secret-like file contents, logs, binaries, generated/vendor content, symlinks, and out-of-repo paths are withheld; exact path vocabulary remains available for grounding.
- Attaches clipboard, screen, Claude, and repository context under one shared character/token proxy budget. Small sources are attached in full; oversized sources are transcript-matched/selected while full harvested vocabulary remains available for conservative input-side technical-term grounding.
- Adds settings/install/enrollment UX, packaged plugin/hook assets, remote-token storage and rotation, listener/session lifecycle, diagnostics/provenance, and extensive regression coverage.

SSH generalization is intentionally asymmetric: the Mac can still capture the visible Ghostty pane and receive hook facts over the tunnel, but it never treats a remote cwd as a local filesystem path. Remote repository bytes are therefore limited to bounded hook-provided snippets; full git/file harvesting remains available only when Claude Code and localvoxtral share the local filesystem.

## Why

Dictation polishing previously saw only the transcript (and a limited clipboard excerpt), so technical identifiers and the actual coding task were often missing. A status line can identify a focused session but cannot provide its off-screen prompt, files, or repo state. This combines Ghostty focus, Claude Code hooks, repository harvesting, and SSH transport while keeping every raw source explicitly authorized and budgeted.

## Proof

- [x] Full unit suite — CI run `29581373237`: `Executed 1757 tests, with 10 tests skipped and 0 failures` in 102.8s. The #152 supervisor kept the complete XCTest stream in the uploaded `unit-test-log-macOS` artifact and printed only the final summary.
- [x] Claude/agent-context suite — `./scripts/remote-build.sh test --filter Claude`: `Executed 594 tests, with 0 failures (0 unexpected)`.
- [x] Shared polish-context suite — `./scripts/remote-build.sh test --filter PolishContext`: `Executed 124 tests, with 0 failures (0 unexpected)`.
- [x] Terminal-screen suite — `./scripts/remote-build.sh test --filter TerminalScreen`: `Executed 71 tests, with 0 failures (0 unexpected)`.
- [x] LLM request-shape suite — `./scripts/remote-build.sh test --filter LLMPolish`: `Executed 30 tests, with 2 skipped and 0 failures`; asserts zero llama.cpp thinking budget, Bifrost extra-parameter passthrough, and unchanged legacy request bytes.
- [x] Release package + launch smoke — CI packaged `localvoxtral`, `localvoxtral-claude-hook`, and `localvoxtral-polishd`; signing, artifact upload, and the copied-bundle launch smoke passed.
- [x] Claude Code manifests — `claude plugin validate` passed for both `localvoxtral` and `localvoxtral-remote`.
- [x] External LLM prompt eval — `./scripts/remote-build.sh eval-llm http://192.168.1.183:8080/v1/chat/completions llamacpp/qwen35-4b`: agent required `1/1`, agent known-hard `7/8`, prompt required `7/7`, prompt known-hard `10/10`, technical informational `2/17`; 2 tests, 0 failures in 104.7s. The lane explicitly sent `thinking_budget_tokens: 0`, `chat_template_kwargs.enable_thinking: false`, and Bifrost's passthrough header so the scoreboard does not spend the response budget on hidden reasoning.
- [x] Realtime integration — live voxmlx: `Executed 7 tests, with 2 skipped and 0 failures`; the spoken synthetic-audio accuracy test passed.
- [x] Helper suites — PolishHelper `47/47`, SpeechHelper `30/30`, and real speech-helper integration `2/2` passed.
- [x] `integration-polishd` — `6 tests, 0 failures` through the packaged helper. Production sampling baseline: required `7/7`, known-hard `10/10`, technical informational `2/17`; agent profile: required `1/1`, known-hard `7/8`. Prompt warmup and alternating two-slot prefix-cache tests passed.
- [ ] Agent E2E — not run in this PR; this many-minute ASR→polish corpus is the nightly/manual tier per repository policy. The direct external prompt eval and packaged-helper production-path baseline above cover the changed context/request path.
- [ ] UI hand test — not yet performed. Verify Settings → Claude Code local install/remove, remote enrollment/rotation/removal, Ghostty focused-pane join, and one SSH RemoteForward session with a CI artifact via `./scripts/try-pr.sh <PR>`.

## Review notes

- Multiple independent Claude Opus reviews were run for plugin/SSH architecture, privacy/security, strict concurrency, and final diff review. Their actionable findings were incorporated: HMAC token migration, transactional registry writes, revoked-session eviction, symlink/secret repo defenses, bounded TERM→KILL CLI teardown, and single-build repo vocabulary indexes.
- No raw screen, clipboard, repo content, prompt, path, token, or remote body is logged; provenance is count-only.


T0mSIlver:
Hand-testing the UI proof item found the **"Install or Update" button broken against the real `claude` CLI**, two independent ways. Fixed in b963435.

**Bug 1 — `--config` on the wrong verb.** The button runs `claude plugin update … --config publisher_path=…`; the CLI only accepts `--config` on `install` (probed on 2.1.212):

```
$ claude plugin update localvoxtral@localvoxtral --config 'publisher_path=…'
error: unknown option '--config'
```

Dropping `--config` from update is not enough: an update that can't re-pin `publisher_path` strands the shim on a stale path whenever the app moves — which the try-pr loop does on every artifact (new /tmp extract each time), and the shim then fails open silently. And the first-ever press ran `update` on a not-installed plugin, which exits 1. `updatePlugin()` is now **marketplace add → best-effort uninstall → install --config**, which handles first install, upgrade, and app-moved in one flow. README manual commands updated to match.

**Bug 2 — duplicate hooks declaration.** Both manifests declared `"hooks": "./hooks/hooks.json"`. Claude Code loads that file by convention and rejects the explicit reference as a duplicate — the plugin installs but lands `✘ failed to load`:

```
Error: Hook load failed: Duplicate hooks file detected: ./hooks/hooks.json resolves to
already-loaded file …/plugins/localvoxtral/hooks/hooks.json. The standard hooks/hooks.json
is loaded automatically, so manifest.hooks should only reference additional hook files.
```

Removed the key from both manifests; both plugins now install `✔ enabled` (validated against the real CLI in an isolated `CLAUDE_CONFIG_DIR`, both `claude plugin validate` still pass).

**Why it escaped:** the unit suite drives the service through a fake runner and CI's `claude plugin validate` checks manifests only — no lane composes a command against the real CLI. The regression tests added here pin the argument flow and the manifests:

- Before fix (tests only, remote unit run): `Executed 88 tests, with 4 failures` — `testUpdateReinstallsAndNeverRunsThePluginUpdateVerb`, `testUpdateToleratesUninstallFailureSoTheFirstPressInstalls`, and both `testPluginManifestIsValidAndLeavesTheStandardHooksFileImplicit`.
- After fix: `Executed 88 tests, with 0 failures`.

Worth a follow-up (not this PR): a self-hosted lane that runs the composed commands against the runner's real `claude` in a sandboxed `CLAUDE_CONFIG_DIR`, so CLI drift like this fails in CI instead of in the field.

PR #152, opened 2026-07-17
Contain and supervise persistent XCTest runs

T0mSIlver:
## What

Contains persistent-runner XCTest orphans and prevents a verbose or genuinely stuck suite from occupying the personal Mac indefinitely, without any global process kill.

- Before self-hosted CI starts tests, reaps only current-UID, non-zombie test processes whose `cwd` or mapped executable is inside the exact Actions workspace. It excludes the cleanup process and ancestors, preserves sibling workspaces, and revalidates process identity before TERM/KILL.
- Runs allowed remote-build payloads in their own process groups, drains descendants when the leader exits, and adds a validated gate-v3 `reap work/localvoxtral-<id>` recovery verb.
- Runs the CI unit suite under a Bash-3.2-compatible supervisor with stdin closed, complete output written directly to an artifact, a 300s self-hosted / 900s hosted bound, and exact TERM→KILL process-group cleanup. Only the final summary/tail reaches the live Actions stream.
- Adds deterministic shell regressions for selector boundaries, real same-workspace `xctest`, sibling preservation, normal/nonzero status, timeout, signal cancellation, stubborn descendants, leader-first exit, gate validation, and remote-build success/failure behavior.

## Root cause

PR #150 repeatedly appeared to stall after different trivial tests, while each blamed test passed independently. Old runs also left XCTest survivors in the persistent checkout. Follow-up reproductions showed a second amplifier: the feature suite emits hundreds of KB of XCTest progress, and foreground streaming could stop mid-line while the same suite completed when run file-backed in a background process. The moving test name was therefore not evidence of a deterministic test deadlock.

This PR addresses both failure modes: scoped orphan cleanup before a run, and a bounded file-backed supervisor during the run. A genuinely stuck suite now fails in at most five minutes on the self-hosted runner and its entire process group is drained.

## Proof

Deterministic regressions:

```text
PASS: cleanup selection is user, state, name, ancestor, and workspace scoped
PASS: production cleanup terminates its xctest and preserves a sibling workspace
PASS: gate preserves status and drains signalled or leader-exited groups
PASS: gate accepts one validated workdir and denies unsafe reap commands
PASS: remote-build reaps one failed payload and leaves successful runs alone
PASS: supervisor captures output, preserves status, and drains exact groups
```

The supervisor regression proves success and exit-7 preservation, FIFO-triggered timeout status 124 without a wall-clock wait, stubborn leader/child cleanup, TERM status 143, leader-first descendant cleanup, and survival of an unrelated sibling process.

Commands:

```bash
./scripts/ci/test-cleanup-stale-test-processes.sh
./scripts/ci/test-build-gate-process-cleanup.sh
./scripts/ci/test-build-gate-reap-command.sh
./scripts/ci/test-remote-build-reap.sh
./scripts/ci/test-run-supervised-command.sh
bash -n scripts/remote-build.sh scripts/mac/localvoxtral-build-gate.sh scripts/ci/*.sh scripts/ci/fixtures/*.sh
git diff --check
```

- [x] Final self-hosted CI — run `29581331529`: all cleanup/supervisor regressions passed; the supervised full unit suite and log upload passed; coverage, live voxmlx integration, PolishHelper and SpeechHelper unit suites, packaging, signing, artifact upload, and copied-bundle launch smoke all passed.
- [x] Feature-stack proof — stacked PR #150 run `29581373237` completed `1757 tests, 10 skipped, 0 failures` in 102.8s under this supervisor, then passed every remaining integration/package/smoke lane. This is the previously failing large-suite shape.
- [x] Scope proof — the CI reaper remains current-UID + exact-workspace scoped; old SSH-builder processes under another UID are not selected.
- [x] LLM inference skipped on this infrastructure PR by the path gate: no prompt, request-shape, context, model, or eval behavior changes.

## Deployment note

CI cleanup and supervision take effect directly from this branch. Remote-build containment degrades safely through the client warning, but gate-v3 process-group supervision and the `reap` verb require installing the reviewed scripts using `scripts/mac/README.md` after merge.

PR #150 is based on this branch so the infrastructure fix stays reviewable separately from the large feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,7 @@ jobs:
           ./scripts/ci/test-build-gate-process-cleanup.sh
           ./scripts/ci/test-build-gate-reap-command.sh
           ./scripts/ci/test-remote-build-reap.sh
+          ./scripts/ci/test-run-supervised-command.sh
 
       - name: Decide LLM eval lane (path filter + [run-llm-eval] marker)
         # CI must not pay ~8 min of 4B weight-loading + live inference on
@@ -215,17 +216,34 @@ jobs:
         # old standalone `swift build` step built plain-debug products that
         # nothing downstream consumed (tests rebuild with coverage flags,
         # packaging builds release), so it was pure duplicate compilation.
+        env:
+          RUNNER_ENVIRONMENT: ${{ runner.environment }}
         run: |
           # AgentDictationE2EEvalTests is skipped explicitly (belt) on top of
           # the marker cleanup above (suspenders): it is the nightly
           # eval-e2e.yml lane, never tier 0, and a leftover enable marker in
           # this persistent workspace must not turn the unit step into a
           # many-minute live-inference run.
-          swift test \
+          timeout_seconds=900
+          if [[ "$RUNNER_ENVIRONMENT" == "self-hosted" ]]; then
+            timeout_seconds=300
+          fi
+          ./scripts/ci/run-supervised-command.sh \
+            "$timeout_seconds" "$RUNNER_TEMP/localvoxtral-unit-tests.log" -- \
+            swift test \
             --skip RealtimeAPIVLLMIntegrationTests \
             --skip SpeechHelperIntegrationTests \
             --skip AgentDictationE2EEvalTests \
             --enable-code-coverage
+
+      - name: Upload complete unit-test log
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: unit-test-log-${{ runner.os }}
+          path: ${{ runner.temp }}/localvoxtral-unit-tests.log
+          if-no-files-found: error
+          retention-days: 7
 
       - name: Coverage summary
         # Visibility, not a gate: lets a PR's Proof section cite real coverage
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,23 @@ jobs:
           rm -f .agent-eval-e2e-enable.json .llm-polish-eval-enable.json \
             .polishd-integration-enable.json .speechd-integration-enable.json
 
+      - name: Clean abandoned test processes in this workspace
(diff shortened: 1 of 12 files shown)


T0mSIlver:
Reviewed (2 independent deep passes: kill-safety/gate security + supervisor/CI/tests). **No blockers**; all five shell regression suites also pass on Linux/bash 5.2. Verified: prefix-boundary workspace scoping (no sibling false-match), UID scoping in selection *and* revalidation, lstart identity revalidation against PID reuse, `reap` verb traversal-safe under `validate_work_dir` (`localvoxtral-..` is a literal dir, not parent), `SOURCE_ONLY` cannot cross the sshd forced-command boundary, payload exit status preserved through `set -e` + case dispatch, timeout-marker-before-TERM ordering guarantees exit 124, and the new signal traps still fire the pre-existing eval-marker EXIT cleanup.

Non-blocking follow-ups:
1. `is_test_process_name`'s bare `localvoxtral*` matches the real menu-bar app and managed helpers (ucomm-truncated `localvoxtral-pol`/`-spe`); in the CI call site the reaper runs as the owner's UID, so workspace path evidence is the *only* guard. Safe today (launch smoke runs the app from a copy outside the workspace; try-pr uses a separate dir) — but worth a comment documenting that no product binary may ever run resident from a scanned tree, or narrowing the glob.
2. `if-no-files-found: error` on the unit-log upload + `if: always()` turns any pre-swift-test step failure into a second, unrelated-looking red step ("No files were found"). The supervisor creates the log unconditionally at start, so `warn` loses nothing.
3. `test-run-supervised-command.sh` / `test-build-gate-process-cleanup.sh` assert descendant death immediately after the async SIGKILL with no bounded poll-for-gone; a loaded host could momentarily still show the reparented sleep live (same family as the known flaky supervisor TERM test).
4. Residual (acknowledged-tradeoff class): revalidation is uid+ucomm+lstart without re-checking path evidence, so a same-second exact-PID-reuse xctest in another workspace is the one theoretical cross-scope kill; probability negligible.

PR #154, opened 2026-07-17
Join the focused Ghostty pane to its Claude session by TTY (title marker fallback)

T0mSIlver:
> Stacked on #150 (base is `codex/agent-context` so this diff stays separate). Merge #150 first, then retarget to main.

## What

Adds a TTY-based focus join as the primary mechanism for resolving which Claude Code session owns the focused Ghostty pane, with the existing `lvx-` title-marker join as fallback.

- `GhosttyFocusedTerminalTTYReader`: one bounded AppleScript read (`with timeout of 1 second`) of `tty of focused terminal of selected tab of front window`, Ghostty-only by bundle id, reply validated as a plausible `/dev/tty…` device path. Error codes are logged, never messages (AppleScript error strings can quote window titles).
- `ClaudeSessionRegistry.resolve(tty:)`: exact match over live LOCAL sessions, with the established abstention semantics (`.ambiguous` for two sessions on one device, `.stale` for a dead process). Remote sessions never match — an SSH host must not be able to claim a local pane by echoing its device path.
- `ClaudeSessionJoinResolver`: tty first (on a hit the title is not read at all), marker fallback unchanged. The marker remains the only join for SSH-remote sessions and pre-1.4 Ghostty. The join is still resolved once per dictation, at start; `isStillLive` revalidation is unchanged (marker-keyed).
- The tty seam defaults to ABSTAIN, unlike the AX seam's default-live pattern: an Apple event triggers the macOS Automation consent prompt, and a defaulted live reader would send real events (and hang on that prompt) from any test that forgot to inject. Only `localvoxtralApp` wires the live reader.
- Packaging: `NSAppleEventsUsageDescription` added to Info.plist.
- AGENTS.md join invariant updated to describe tty-first resolution.

## Why

Field finding while hand-testing #150 (2026-07-17): Claude Code writes its own conversation-derived window titles and clobbers the marker whenever the model is responding — the title is a fought-over, last-writer-wins channel, so the marker-only join abstains exactly when the user dictates into a busy session. The process table has no such war: the hook publisher already reports the session's controlling TTY (`ClaudeHookProcessInfo.tty`), and Ghostty ≥ 1.4 exposes the focused pane's TTY over AppleScript (ghostty-org/ghostty#11592, merged in ghostty#11922 for the 1.4.0 milestone). Equality is exact, so two sessions in one repo — the case `resolve(workspace:)` deliberately abstains on — resolve correctly by device. This is the "focus join that is deliberately not built yet" that comment was waiting for.

## Proof

- [ ] CI unit suite on this branch — pending; the build host is temporarily unreachable from the dev box (moved to VPN), so the suite has not yet run on the final commit. Do not merge before it is green.
- [ ] Local `remote-build.sh test --filter Claude` — same blocker; will be run and pasted here when the host is reachable.
- [ ] Hand test of the live tty path — requires Ghostty ≥ 1.4 (tip channel today; stable is 1.3.1 where the `tty` property does not exist, verified on the owner's Mac: AppleScript error -1700). On 1.3.x this build must behave identically to the marker join.
- New regression tests (names, per proof culture):
  - Registry: `testTTYLookupResolvesTheLocalSessionOnThatDevice`, `testTTYLookupNeverMatchesARemoteSession` (spoof refusal), `testTwoLocalSessionsOnOneTTYAbstainAsAmbiguous`, `testTTYLookupReportsStaleWhenTheOnlyMatchIsDead`, `testTTYLookupIsUnknownForAnUnseenDevice`
  - Resolver: `testTTYJoinsWhenTheTitleCarriesNoMarker` (the title-war case), `testTTYJoinDoesNotReadTheTitleAtAll` (precedence observable), `testTTYJoinWinsOverAStaleMarkerNamingAnotherSession` (recycled pane), `testUnmatchedTTYFallsBackToTheTitleMarker`, `testAmbiguousTTYFallsBackToTheTitleMarker`, `testRemoteSessionNeverJoinsViaTTY`, `testValidatedTTYAcceptsOnlyPlausibleDevicePaths`

## Review notes

- The empirical probe behind the design: `working directory of focused terminal of selected tab of front window` resolves correctly on the owner's Ghostty 1.3.1 (focus chain + Automation grant proven); `tty` errors -1700 there, which is the exact fallback case the resolver tests pin.
- These paths are in `scripts/ci/llm-lane-filter.sh` already (`Sources/localvoxtral/ClaudeContext/*`, `*ClaudeSessionRegistry*`, `*TerminalScreenClaudeJoin*`), so the LLM lanes trigger on this diff by the existing rules.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -391,11 +391,20 @@ Key subsystems:
     filesystem" is a compile error — do not add one. Its only derivations
     (`ancestor`, `descendant`) preserve that, and `ClaudeRepoCollecting` takes
     it rather than a `String` for exactly this reason.
-  - The join is resolved ONCE per dictation, at start, from the marker in the
-    PID-pinned window title (`ClaudeSessionJoinResolver`), and every consumer —
-    raw screen attachment, the session block, repo collection — shares that one
-    answer. Three resolutions could each answer honestly about a different
-    moment; that is how one session's screen ends up next to another's repo.
+  - The join is resolved ONCE per dictation, at start
+    (`ClaudeSessionJoinResolver`), and every consumer — raw screen attachment,
+    the session block, repo collection — shares that one answer. Three
+    resolutions could each answer honestly about a different moment; that is
+    how one session's screen ends up next to another's repo. Resolution is
+    TTY-first: the focused Ghostty pane's controlling TTY (AppleScript,
+    Ghostty ≥ 1.4) matched exactly against the hook-reported session TTY,
+    LOCAL sessions only — the title is a fought-over channel (Claude Code's
+    own conversation titles clobber the marker mid-turn), the process table is
+    not. Any TTY non-answer falls through to the marker in the PID-pinned
+    window title, which remains the ONLY join for SSH-remote sessions: a
+    remote TTY names another machine's device, and `resolve(tty:)` refuses
+    remote candidates so an SSH host can never claim a local pane by echoing
+    its TTY.
   - Lookups abstain rather than guess: no marker, unknown, stale, or ambiguous
     means no context. There is deliberately no sole-session or cwd heuristic —
     it is wrong precisely when it matters.
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ On first launch, a setup wizard walks you through the microphone and Accessibili
 
 ## Privacy
 
-In the default Managed local mode, nothing you say or write is sent anywhere. Audio capture, transcription, and LLM polishing all run as local processes on your Mac, and the only network traffic is the one-time engine and model download. There is no telemetry, no account, and no cloud fallback. The context-aware polishing features (repo vocabulary, clipboard context) are opt-in and only ever talk to a loopback polishing endpoint, never a remote server. If you point localvoxtral at your own External URL server instead, your data goes only where you send it.
+In the default Managed local mode, nothing you say or write is sent anywhere. Audio capture, transcription, and LLM polishing all run as local processes on your Mac, and the only network traffic is the one-time engine and model download. There is no telemetry, no account, and no cloud fallback. The context-aware polishing features (repo vocabulary, clipboard context) are opt-in and by default only ever talk to a loopback polishing endpoint — a non-local endpoint receives context only if you additionally enable the explicit trusted-endpoint opt-in (default off). If you point localvoxtral at your own External URL server instead, your data goes only where you send it.
 
 ## Shortcuts
 
--- a/Sources/ClaudeHookPublisherCore/ClaudeHookPublisher.swift
+++ b/Sources/ClaudeHookPublisherCore/ClaudeHookPublisher.swift
@@ -7,6 +7,10 @@ import Darwin
 import Glibc
 #endif
 
+#if canImport(os)
+import os
+#endif
+
 /// Reads bounded stdin once, enriches with safe process/TTY metadata, and
 /// publishes one NDJSON line. Pure logic lives here so the executable's `main`
 /// stays a three-liner and every branch below is unit-testable.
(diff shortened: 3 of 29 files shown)


T0mSIlver:
CI was red for two independent reasons, both fixed on this branch:

1. **Compile error** (`ed4b202`): `NSAppleScriptErrorNumber` is rejected by the macOS 26 SDK — renamed to `NSAppleScript.errorNumber`.

2. **Deterministic tier-0 hang** (`201bfdb`): with the compile fixed, the unit suite froze until timeout (2× CI, 3× SSH gate). A `sample` of the wedged xctest (captured via a temporary diagnostic branch, #156) showed the suite parked in `NSAlert.runModal` → `_DPSBlockUntilNextEventMatchingListInMode`:
   `testStaleSecureIconDoesNotMaskEarlyExitFailuresOnNextAttempt` → `beginDictationSession` → `handleConnectFailure(.invalidEndpoint)` → `presentConnectionFailureAlert` → real modal, waiting for a click that never comes. The `NSApp != nil` guard doesn't help once any earlier test touches `NSApplication.shared` — this branch's new suites shifted that initialization order, arming a landmine that exists on main too (a canary run of unmodified main, #155, stayed green only because its suite order never initializes NSApp first).

**Proof** (full unit suite via the SSH gate — the environment that hung 3/3 before the fix):
```
Executed 1761 tests, with 2 tests skipped and 0 failures (0 unexpected) in 64.031 (64.115) seconds
testConnectionFailureAlertNeverRunsModalEvenWithNSAppInitialized passed (0.006 seconds)   <- new regression test
testStaleSecureIconDoesNotMaskEarlyExitFailuresOnNextAttempt     passed (0.009 seconds)   <- previously the hang site
```
Before the fix, the same command wedged indefinitely at that test (log frozen 3+ hours; sample attached to the #156 run artifact).

⚠️ The base branch `codex/agent-context` (#150) will hit the same hang on its next CI run — cherry-pick `201bfdb` down (or merge this PR first).

T0mSIlver:
Trusted-endpoint toggle (`c7ce003`) proof:

- Full unit suite via the SSH gate: **1766 tests, 0 failures** (5 new: policy truth table, screen-gate accept + only-endpoint-relaxed, repo-context remote-admitted + feature-toggle-not-bypassed, fresh-install default off).
- CI green including both live-model lanes (auto-triggered by the path filter): `PolishHelperIntegrationTests` 6/6 vs the real pinned model + eval baseline, `SpeechHelperIntegrationTests` 2/2 vs real audio. Default-off means nothing reaches the model differently unless the user opts in; the lanes confirm the default path is unchanged.
- UI: one new row in the existing Text Processing group (group structure unchanged per the pane rule). Hand-verification planned via try-pr: point polishing at a LAN endpoint, confirm context features stay OFF by default and activate with the toggle.

T0mSIlver:
CI timing note from profiling this branch's runs: **`PolishContextPreparationTests` alone is 38 s of the unit step's 64 s of test execution** (run 29694638494's unit-log artifact; the other 120 suites total ~26 s). Two tests dominate:

- `testOversizedSingleLineIsTruncatedWithAMarkerWithinTheCap` — **20.9 s**
- `testOversizedLineIsSkippedSoALaterFittingLineStillRenders` — **10.3 s**

Both just prepare a ~20–40 KB *single-line* buffer against a 500-char render budget. That input size taking tens of seconds (debug build) points at something superlinear in the oversized-single-line path — plausibly the entity recognizer's nine `NSRegularExpression` passes or repeated `String` index arithmetic over the giant line. Notably the suite's own `testSelectionOverManyLinesCostsNoMoreThanOneWholeBufferExtraction` guards the *many-line* quadratic that was already fixed, but the single-oversized-line path has no such budget assertion and seems to have the same class of problem.

Two reasons to care beyond CI time:

1. **Field latency**: this is real clipboard-capture code, and the test's own doc comment names the realistic inputs — "a minified bundle, a base64 blob". A user copying one of those before dictating would eat this cost at polish time.
2. **CI**: fixing it roughly halves unit-suite wall time on every push (64 s → ~26 s).

The suite comes in via the base branch (`codex/agent-context`, #150 — commits 8ead826/f7468bf), so the fix probably belongs there rather than here. A budget assertion in the style of the existing many-line one would keep it from regressing once fixed.

PR #155, opened 2026-07-19
CI canary: host-state check (do not merge)

T0mSIlver:
Diagnostic only: runs the unit suite on unmodified main to test whether today's full-suite hang (seen on #154 and its base) is host-state. Will be closed and deleted after the run.


T0mSIlver:
Canary served its purpose: unmodified main is green on the runner today, so the unit-suite hang is specific to the agent-context lineage, not host state.

PR #156, opened 2026-07-19
Diagnostic: sample the unit-suite hang (do not merge)

T0mSIlver:
Captures thread stacks of the deterministic unit-suite hang on the agent-context lineage (#154 investigation). Will be closed after one run.


T0mSIlver:
Diagnosis complete: the sample caught the suite parked in NSAlert.runModal — presentConnectionFailureAlert runs a real modal once any earlier test initializes NSApplication.shared. Fix + regression test land on tty-session-join (#154).

PR #158, opened 2026-07-20
ui-smoke: evening lock-aware schedule (20:00 Paris anchor + retry slots)

T0mSIlver:
## What

The 04:00 ui-smoke nightly was structurally wasted: the drill needs an **unlocked GUI session**, and at 04:00 the MacBook is always locked. The failure signature is unambiguous in run [29722553773](https://github.com/T0mSIlver/localvoxtral/actions/runs/29722553773)'s log artifact — the menu bar is still AX-readable (status-item steps PASS) but no window can be presented, so "Settings did not open" and all six tab selections FAIL. Daily red noise that trains everyone to ignore the lane.

This replaces the nightly with an **evening lock-aware retry ladder**, anchored at 20:00 Paris (owner request):

- Three schedule slots: 18:00 / 19:30 / 21:00 UTC (20:00 / 21:30 / 23:00 Paris in summer; an hour earlier local in winter — the requirement is "evening, Mac likely unlocked", not a precise hour).
- Each scheduled slot runs `scripts/ci/ui-smoke-guard.sh`, which decides:
  - **skip (green)** if a successful ui-smoke run completed in the last 20 h — at most one real drill per day, later slots are ~10 s no-ops;
  - **skip (green)** if the console session is locked or absent (`CGSessionCopyCurrentDictionary` probe — the lock key is only present while locked);
  - **run** otherwise, and on any probe/API error it **fails open** (same philosophy as the CI lane filters: an uncomputable state must never silently disable a lane).
- The decision is written to the step summary (`UI smoke slot: RUNNING/SKIPPED (reason)`), like ci.yml's lane-decide steps, so a skipped slot is self-explanatory.
- `workflow_dispatch` and `needs-ui-smoke` label runs bypass the guard entirely — explicit operator intent always runs.

Also updated: eval-e2e.yml's cron comment (it no longer needs to dodge a 04:00 ui-smoke) and the AGENTS.md tier-2 row.

## Proof

- Guard decision logic is covered by `scripts/ci/test-ui-smoke-guard.sh` (both probes stubbed via the script's env seams — no gh, no swift; runs on any OS), wired into ci.yml's shell-test step so it runs per-push including hosted fork PRs. Local run:

  ```
  PASS: recent success skips regardless of lock state (successful ui-smoke run 1 h ago — today is already covered)
  PASS: success 19 h ago still counts as covered (successful ui-smoke run 19 h ago — today is already covered)
  PASS: success 21 h ago is stale — run again (screen unlocked, no recent successful run)
  PASS: locked screen skips (screen is locked — the AX window drill would false-red; the next slot retries)
  PASS: missing GUI session skips (screen is no-session — the AX window drill would false-red; the next slot retries)
  PASS: unlocked screen runs (screen unlocked, no recent successful run)
  PASS: probe error fails open into a run (lock probe unavailable (error) — failing open)
  ui-smoke-guard tests passed
  ```

- The live `swift`/`gh` probe paths run for the first time at tonight's first scheduled slot after merge; the step summary will show the decision and reason. If the probe errors on the runner, the guard fails open into a normal run — behavior is never worse than today's.
- Unit suite / realtime integration: not run locally — no Swift or app code is touched (workflow YAML, two new shell scripts, AGENTS.md); CI's tier 0/1 on this PR covers the tree.
- LLM lanes: skipped — no LLM-relevant paths (the path filters won't match, correctly).


T0mSIlver:
Adversarial review (opencode / glm-5.2) found 4 issues, all fixed in 8834a3a:

1. **HIGH — skip-runs counted as coverage.** A guard-skipped slot also concludes `success`, so the `?status=success` dedup let a locked 18:00 slot suppress the 19:30/21:00 retries; the ladder only retried across days. The guard now counts only runs whose **`Run AX UI smoke` step conclusion is `success`** (scans the last 10 success runs via the jobs API). Regression-tested through the real query path with a PATH-stubbed `gh`; before/after: the pre-fix guard returns `run=false (covered)` for a fresh skip-run with the screen unlocked, the fixed one retries.
2. **MED — fail-open hardening**: step conditions now use `run != 'false'` so a guard emitting no output can never green-skip every slot forever.
3. **LOW — fractional-second timestamps** no longer break the age parse.
4. **LOW — workflows/README.md** still described the old nightly; updated.

Full test run (10 cases): all PASS, including the exact review scenario ("guard-skip alone does not count as covered — retry runs").

PR #160, opened 2026-07-20
Fix the tier-0 hang class: XCTest modal guard on main + supervisor timeout forensics

T0mSIlver:
## What

Fixes the tier-0 unit-suite hang class (three 300 s-timeout kills in two days: runs [29690098130](https://github.com/T0mSIlver/localvoxtral/actions/runs/29690098130), [29693730846](https://github.com/T0mSIlver/localvoxtral/actions/runs/29693730846), [29770530435](https://github.com/T0mSIlver/localvoxtral/actions/runs/29770530435) attempt 1) in two parts:

1. **Cherry-pick the sample-proven modal guard to main** (`201bfdb` from `tty-session-join`, PR #154's branch): `presentConnectionFailureAlert`'s `NSApp != nil` guard is insufficient in an xctest process — once any earlier test touches `NSApplication.shared`, NSApp is non-nil for the rest of the suite, and a connect failure then runs a REAL `NSAlert.runModal()`, parking the whole run in `_DPSBlockUntilNextEventMatchingListInMode` (hand-sampled on the wedged xctest, 2026-07-19). Main has the identical unguarded `runModal` — a survey of the codebase shows it is the only unbounded main-thread parker (`localvoxtralApp.swift`'s is app-startup-only; `RepoVocabulary`'s semaphore waits are bounded). The guard on `NSClassFromString("XCTestCase")` plus the regression test come with the cherry-pick.

2. **Timeout forensics in `run-supervised-command.sh`**: the watchdog now `sample`s up to three members of the wedged process group (xctest first) into the log file — which CI already uploads as the `unit-test-log-macOS` artifact even on failure — *before* killing the group. Forensic evidence for this: the 07-20 hang on the main lineage (attempt 1 of run 29770530435) froze **entering `BackendManagerTests`** — a *different* signature from the 07-19 NSAlert hangs (which both froze right before `testStaleSecureIconDoesNotMaskEarlyExitFailures…`, the alert-driving test). I traced the 07-20 wedge's candidates (connect-timeout timer arithmetic, `ensureReady` single-flight, `isReady`, the port defense, the test fakes' continuation machinery) and ruled each out from the logs and code — its true trigger is unidentified, so the next occurrence must self-diagnose instead of costing another blind rerun.

Investigation record: the 07-19 hang logs both stop at the same alert-driving test (NSAlert signature, fix #1); the 07-20 log stops milliseconds into the first `@MainActor` test after three non-MainActor suites, at 21:06:06.910, with no view-model-owning suite having run — consistent with a parked main thread of unknown origin (fix #2 captures it next time).

## Proof

- **Modal guard, before**: PR #154's branch hung deterministically pre-fix (runs 29690098130 / 29693730846 — unit-log artifacts stop at the alert-driving test; xctest hand-sample showed `runModal` parked in `_DPSBlockUntilNextEventMatchingListInMode`) and went green immediately after `201bfdb` landed on it (run [29694638494](https://github.com/T0mSIlver/localvoxtral/actions/runs/29694638494)). That is the failing-before/passing-after pair for this exact commit; the failing mode is a 300 s hang, so it is not reproduced fresh here.
- **Modal guard, after (on main lineage)**: `./scripts/remote-build.sh test --filter TerminalTargetDetectorTests` — `Executed 38 tests, with 0 failures`, including:
  ```
  Test Case '-[localvoxtralTests.TerminalTargetDetectorTests testConnectionFailureAlertNeverRunsModalEvenWithNSAppInitialized]' passed (0.020 seconds).
  ```
- **Forensics**: `./scripts/ci/test-run-supervised-command.sh` (new case: PATH-stubbed `sample`, deterministic FIFO-triggered timeout):
  ```
  === supervisor timeout forensics: sampling process group 2986747 ===
  --- sample pid 2986747 (bash) ---
  stub-sample-of-pid-2986747
  --- sample pid 2986749 (sleep) ---
  stub-sample-of-pid-2986749
  PASS: supervisor captures output, preserves status, and drains exact groups
  ```
- Full unit suite runs on this PR's CI (tier 0).

Note for the #150/#154 stack: `201bfdb` is cherry-picked with `-x`; when those PRs merge, the identical hunk should auto-resolve.

### DIFF
--- a/Sources/localvoxtral/DictationViewModel+Session.swift
+++ b/Sources/localvoxtral/DictationViewModel+Session.swift
@@ -1668,6 +1668,15 @@ extension DictationViewModel {
             Log.dictation.error("connection-failure alert skipped: no NSApplication in this process")
             return
         }
+        // The nil guard above is NOT sufficient in a test process: any earlier
+        // test that touches NSApplication.shared initializes NSApp for the rest
+        // of the suite, and runModal() then stops the whole run dead waiting
+        // for a click that never comes (xctest sample 2026-07-19: this frame
+        // parked in _DPSBlockUntilNextEventMatchingListInMode mid-suite).
+        guard NSClassFromString("XCTestCase") == nil else {
+            Log.dictation.error("connection-failure alert skipped: XCTest process")
+            return
+        }
         NSApp.activate(ignoringOtherApps: true)
         let alert = NSAlert()
         alert.alertStyle = .warning
--- a/Tests/localvoxtralTests/TerminalTargetDetectorTests.swift
+++ b/Tests/localvoxtralTests/TerminalTargetDetectorTests.swift
@@ -1,3 +1,4 @@
+import AppKit
 import Carbon
 import Foundation
 import XCTest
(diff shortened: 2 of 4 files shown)


T0mSIlver:
Adversarial review (opencode / glm-5.2): the Swift side came back clean — reviewer verified `NSClassFromString("XCTestCase")` matches the existing `TerminalTargetDetector.isRunningUnderXCTest` precedent, cannot fire in the shipped app (XCTest.framework not linked), and skipping the modal is state-machine-safe because `handleConnectFailure` sets `lastError`/status surfaces *before* the alert call; the regression test provably pins the hang (invalid-endpoint reaches the alert synchronously, before `scheduleConnectTimeout`).

Four findings in the forensics shell code, all fixed in e670e49:
1. **HIGH — unbounded `sample` could postpone the kill indefinitely** (a sampler stuck on a wedged task = the exact blind hang this feature diagnoses). Each sample now runs in its own background group with a hard cap (default 8 s), group-killed on expiry.
2. **MED — 124-mislabel race widened to seconds**: the timeout marker was written before sampling, so a command finishing naturally during forensics reported 124 with a green log. Marker now written only if the group is still alive after sampling.
3. `seen_pids` not `local`. 4. Missing `pgrep` now logs an explicit skip instead of silently disabling forensics.

Both behavioral fixes have deterministic regression cases (hung stub sampler → killed at 300 ms test cap, 124 + full group drain preserved; flag-released natural finish during forensics → real status 0 kept). Full suite: 7 cases, all PASS in 1.3 s.

PR #161, opened 2026-07-20
speechd-bench: optional cache-limit-mb argument

T0mSIlver:
## What

`./scripts/remote-build.sh speechd-bench [seconds] [cadence-ms] [cache-limit-mb]` — the new optional third argument rides the bench marker JSON into the packaged helper's existing `--cache-limit-mb` flag. Omitted, the marker is byte-identical to before.

## Why

The speechd perf investigation (2026-07-20) needed a pool-cap A/B probe (4 GB vs 2 GB) to test whether buffer-pool hoarding was degrading the lm-head gemv; without this knob that requires editing and re-syncing test code per run. (Result, for the record: cap size made no measurable difference — the phases are bandwidth-bound.)

## Proof

Exercised live on the Mac: `speechd-bench 60 100 2048` ran the packaged helper with the cap applied — `BENCH mark=5s … cache_mb=2048.0`, `mark=15s … cache_mb=2047.2`, `mark=30s … cache_mb=2059.5`, `mark=60s … cache_mb=2068.9` (vs 4095–4115 on the default runs bracketing it in the same session). Default two-argument invocations before and after the change produced identical marker JSON and normal runs (baseline + 5 candidate benches in the same session). Tier-0 unaffected: the lane is marker-gated and skips without the marker file; the decoder-side `cacheLimitMB` field is optional, so existing markers parse unchanged.

### DIFF
--- a/Tests/localvoxtralTests/SpeechdStreamingBenchTests.swift
+++ b/Tests/localvoxtralTests/SpeechdStreamingBenchTests.swift
@@ -16,6 +16,7 @@ final class SpeechdStreamingBenchTests: XCTestCase {
         let seconds: Int
         let cadenceMilliseconds: Int
         let wavPath: String?
+        let cacheLimitMB: Int?
     }
 
     private var repoRoot: URL {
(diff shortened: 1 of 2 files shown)


T0mSIlver:
opencode (GLM-5.2) review for the 0.9.0 release pass: **VERDICT: PASS**, one Low finding — the `speechd-bench` header comment in `remote-build.sh` didn't mention the new third `cache-limit-mb` argument. Fixed in the follow-up commit on this branch. Full plumbing verified end-to-end (script → marker JSON → test → `--cache-limit-mb` → `Memory.cacheLimit`); default path byte-identical to pre-PR.

PR #162, opened 2026-07-20
scripts/mac: quantized-tied-head checkpoint fork tool

T0mSIlver:
## What

`./scripts/mac/convert-qhead-checkpoint.py --out DIR [--upload T0mSIlver/...-qhead]` — forks the pinned `mlx-community/Voxtral-Mini-4B-Realtime-2602-4bit@fdebf7b2` snapshot with only the tied embedding quantized (4-bit/g64, the checkpoint's own scheme), verifies the round-trip, rewrites the shard + index, and optionally uploads under your HF account. uv script header self-provisions Python + mlx.

## Why

Companion to mlx-audio-swift fork PR #11 (load-time head quantization, ~30→3 ms lm-head, eval-e2e clean) and `perf/quantized-head-checkpoint` (loader accepts checkpoint-shipped embed scales). Shipping the quantization in the checkpoint saves ~540 MiB of download/disk per install and removes the load-time conversion; the load-time path remains as the fallback for fp16-head checkpoints. Quantization is deterministic, so the produced tensors are bit-identical to what the engine computes at load — PR #11's evidence transfers.

## Proof

Owner-runbook tooling, no app behavior change, tier-0 unaffected (nothing imports it). The gate blocks running it remotely by design (`uv` is not allowlisted; verified — `remote-build.sh exec uv --version` → "denied command"), and the upload needs owner HF credentials, so live execution is deliberately an owner step. After upload, the per-model gate is one command: `./scripts/remote-build.sh integration-speechd T0mSIlver/<repo>` (self-provisions weights), followed by the SpeechModelCatalog pin-switch PR carrying that evidence.

### DIFF
--- a/SpeechHelper/Package.resolved
+++ b/SpeechHelper/Package.resolved
@@ -23,7 +23,7 @@
       "kind": "remoteSourceControl",
       "location": "https://github.com/T0mSIlver/mlx-audio-swift.git",
       "state": {
-        "revision": "23c2c3b3357fbd454b9beb778a66022cb7f86963"
+        "revision": "03890317975a2371fe0a0a9b13ad6a790f929814"
       }
     },
     {
--- a/SpeechHelper/Package.resolved
+++ b/SpeechHelper/Package.resolved
@@ -23,7 +23,7 @@
       "kind": "remoteSourceControl",
       "location": "https://github.com/T0mSIlver/mlx-audio-swift.git",
       "state": {
-        "revision": "03890317975a2371fe0a0a9b13ad6a790f929814"
+        "revision": "23c2c3b3357fbd454b9beb778a66022cb7f86963"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


T0mSIlver:
opencode (GLM-5.2) review for the 0.9.0 release pass: **VERDICT: PASS** with findings, all now addressed on the branch:
- The pin bump to fork commit `23c2c3b` (+ stale DEPENDENCY.md) was **superseded by #169** (pins `791ab87` = upstream main + Blaizzy#232, docs rewritten) — dropped here; this PR is now the conversion tool only, which also moots the speechd-lane requirement the reviewer flagged.
- Script fixes: `.gitattributes` is now copied (the 'same snapshot' promise), `index.setdefault("metadata", {})` guards sources whose index legally omits it, and the docstring no longer claims the engine has no load-time quantize path (it exists; an already-quantized head just never reaches it).

PR #163, opened 2026-07-21
Build-host disk GC: gate v4 gc/disk verbs, per-use stamps, auto-gc after every remote run

T0mSIlver:
## What

`/Users/builder/work` on the build host fills the disk: every ephemeral Linux worktree mints its own `work/localvoxtral-<slug>-<hash>` dir (deliberate parallel-agent isolation), each keeps a multi-GB SwiftPM/xcodebuild tree alive across runs (deliberate incremental-build cache), and nothing ever deleted them. This adds the missing garbage collection plus disk visibility:

- **Gate v4** (`scripts/mac/localvoxtral-build-gate.sh`):
  - Every gated use of a work dir (`mkdir` / rsync / build payload) touches a `.lv-last-used` stamp, so staleness never depends on rsync-preserved source mtimes (`rsync -a` copies the *sender's* old mtimes).
  - New `gc` verb: deletes `work/localvoxtral-*` dirs with no entry modified in `LV_GC_MAX_AGE_DAYS` (default 14, overridable in `~/.localvoxtral-gate.conf`). Three independent keep-checks, each failing toward "keep": recent activity (stamp or fresh build products, for pre-stamp dirs), live processes rooted in the dir (lsof cwd/mapped-text evidence, same as `reap`), and **`EvalRecordings/` presence, which downgrades deletion to a prune that preserves the recordings in place** — private human WAVs may exist only in that remote copy.
  - New `disk` verb (df + per-dir du + last-used ages), a cheap Disk section in `diag`, and its `Data volume free: N GiB` line.
- **remote-build.sh** fires `gc` backgrounded/best-effort after every run — the disk only fills while agents build, which is exactly when it runs; no cron/LaunchDaemon. Both client rsyncs gain `--filter='P .lv-last-used'`. New `disk`/`gc` pass-through verbs.
- **mac-health.sh** surfaces the free-space line and warns below `LV_MIN_FREE_GIB` (default 25 GiB).
- Portability bug fixed along the way: GNU `stat -f %m` *succeeds* with a mount point instead of an mtime, so `file_mtime` tries `-c` first (BSD fails `-c` cleanly).
- Docs: `scripts/mac/README.md` v4 section (+ HF-cache / TTS-cache notes), AGENTS.md.

**Owner action to activate**: reinstall the gate script on the Mac (same flow as v3, `scripts/mac/README.md` "Installing / upgrading the gate"). Until then the auto-`gc` is denied and logged (`DENY gc` once per run) — verified fail-open below.

## Proof

- [x] New regression tests (run on this Linux box; also wired into both ci.yml shell-test blocks):

```
$ ./scripts/ci/test-build-gate-gc-command.sh
PASS: gate stamps work dirs on use, gc reclaims only stale idle dirs, recordings survive
$ ./scripts/ci/test-remote-build-gc.sh
PASS: remote-build fires one background gc per run and shields the stamp from --delete
```

Covers: stamp created by `mkdir` verb and refreshed by the `cd`+payload verb; stale dir deleted; fresh-stamp dir kept; old-stamp-but-fresh-build-product dir kept (pre-stamp fallback); stale dir with `EvalRecordings/` pruned around the intact recordings; stale dir with live processes (stubbed lsof) kept; non-`localvoxtral-*` dirs untouched; `gc`/`disk` argument injection denied with exit 126; `disk` prints a numeric last-used age (regression-pins the GNU-stat fix).

- [x] Existing gate/remote-build tests still green:

```
PASS: remote-build reaps one failed payload and leaves successful runs alone
PASS: gate accepts one validated workdir and denies unsafe reap commands
PASS: gate preserves status and drains signalled or leader-exited groups
```

(+ test-cleanup-stale-test-processes, test-run-supervised-command, test-ui-smoke-guard, test-clean-stale-outputs all PASS.)

- [x] Live against the **deployed v3 gate** — the added rsync protect filter does not disturb the pinned server-argv check, and auto-`gc` fails open:

```
$ ./scripts/remote-build.sh test --filter TextMergingAlgorithmsTests
Test Suite 'TextMergingAlgorithmsTests' passed at 2026-07-21 09:41:25.598.
	 Executed 84 tests, with 0 failures (0 unexpected) in 0.009 (0.012) seconds
$ cat .build/last-gc.log
localvoxtral build gate: denied command
$ ./scripts/mac-health.sh    # still parses old-gate diag, no disk line, no warning
OK: SSH gate reachable (gate v2)
```

- LLM lanes: skipped — build-infrastructure/shell-only change, nothing that reaches the model.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,9 @@ jobs:
           ./scripts/ci/test-cleanup-stale-test-processes.sh
           ./scripts/ci/test-build-gate-process-cleanup.sh
           ./scripts/ci/test-build-gate-reap-command.sh
+          ./scripts/ci/test-build-gate-gc-command.sh
           ./scripts/ci/test-remote-build-reap.sh
+          ./scripts/ci/test-remote-build-gc.sh
           ./scripts/ci/test-run-supervised-command.sh
           ./scripts/ci/test-ui-smoke-guard.sh
           ./scripts/ci/test-clean-stale-outputs.sh
@@ -78,7 +80,9 @@ jobs:
           ./scripts/ci/test-cleanup-stale-test-processes.sh
           ./scripts/ci/test-build-gate-process-cleanup.sh
           ./scripts/ci/test-build-gate-reap-command.sh
+          ./scripts/ci/test-build-gate-gc-command.sh
           ./scripts/ci/test-remote-build-reap.sh
+          ./scripts/ci/test-remote-build-gc.sh
           ./scripts/ci/test-run-supervised-command.sh
 
       - name: Decide LLM eval lane (path filter + [run-llm-eval] marker)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,7 +50,11 @@ letting rsync hang or CI queue forever.
 
 Parallel agents are isolated automatically: the remote dir defaults to a
 per-worktree name derived from the local checkout path. `LV_BUILD_DIR`
-still overrides it when you need a specific dir.
+still overrides it when you need a specific dir. Abandoned remote dirs are
+garbage-collected by the gate's `gc` verb (fired best-effort after every
+run; 14-day unused window, live dirs and `EvalRecordings/` always kept), so
+minting a fresh dir is cheap — never hand-clean `~/work` on the Mac.
+`./scripts/remote-build.sh disk` shows per-dir sizes and last-used ages.
 
 - An interrupted remote run can leave a stale SwiftPM lock in its remote dir —
   don't debug it, switch to a fresh `LV_BUILD_DIR`.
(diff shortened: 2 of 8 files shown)


T0mSIlver:
**Second-opinion review** (opencode / GLM-5.2, read-only agent): verdict was *merge after fixes* — 1 major, 4 minor, 7 nits, all with verified repros. Fixed in 61b6982:

- **M1 (fixed)**: `workdir_in_use` failed **open** when lsof exists but errors (empty output → awk exit 1 → "not in use" → delete). Now: lsof not running cleanly = evidence unavailable = keep. Test-pinned with an erroring lsof stub.
- **m1 (fixed)**: activity `find` was `-maxdepth 4`; an overwritten xcodebuild product at depth 9 bumps no ancestor dir mtime, so a pre-stamp dir with only `package` activity could be reclaimed. Now depth 9, test-pinned with a deep-overwrite fixture.
- **m2 (fixed)**: the new unwrapped `df` line could abort all of `diag` under `set -e`, which mac-health would misread as "host unreachable". Now `|| true`.
- **m3/m4 (fixed)**: `rm -rf`/`du` failures no longer abort the whole gc pass.
- **n1 (fixed)**: already-pruned EvalRecordings skeletons are skipped silently instead of re-reported as "pruned" every run (test-pinned).
- **n6 (fixed)**: stamp protect filter root-anchored (`P /.lv-last-used`).
- **n4/n5/n7 (accepted, not changed)**: the gc-vs-marker-cleanup-rsync interaction is safe because the payload's `cd` verb stamps the dir before any cleanup rsync (reviewer's own analysis concurs); the sub-ms mkdir→stamp race can at worst delete an empty dir the next rsync recreates; the `disk` age display vs gc decision mismatch is cosmetic.

Proof after fixes: all shell regression tests pass locally (`test-build-gate-gc-command.sh`, `test-remote-build-gc.sh`, plus the existing reap/process-cleanup/remote-reap suites), and a live `remote-build.sh test --filter TextMergingAlgorithmsTests` (84/84) confirms the anchored filter still clears the deployed v3 gate's pinned rsync server-argv check.

PR #183, opened 2026-07-22
Claude context: herdr pane.read screen context + iTerm2/Terminal.app joins with AppleScript screen route

T0mSIlver:
> **Stacked on #181** (`herdr-first-class-join`). Merge #181 first; then retarget this PR to `main` BEFORE deleting the base branch (GitHub auto-closes children on base-branch delete). GitHub's native Stacked PRs preview is not enabled for this repo (waitlist), so this is a classic base-branch stack; `gh stack submit` can link it natively later if the repo is enrolled.

[run-llm-eval]

## What

Two features that close the screen-context gaps left by #181, implemented as two commits:

**1. herdr joins get pane-exact screen context via `pane.read`** (`68ea4b9`)

A herdr-joined session previously got no screen excerpt at all: the join authorizer refuses the composite AX capture (neighboring panes must not ride into this session's prompt), so herdr dictations were vocabulary-only. This builds the planned follow-up: fetch the JOINED pane's clean visible text over herdr's JSON socket and use it as the screen context — feature parity with a raw Claude-in-Ghostty session.

- Wire shape pinned against herdr source `c234f221`: request `{"id","method":"pane.read","params":{"pane_id","source":"visible","format":"text","strip_ansi":true}}`; the client validates the response's `pane_id` against the request. herdr remains arm's-length AGPL IPC — nothing vendored.
- `pane.read` fires only downstream of a resolved `.herdrPane` join and only for that join's pane (`ClaudeHerdrPaneBinding` captured at resolution — there is no other place a pane id enters the path).
- The composite AX capture still never attaches for herdr joins; the refusal and its log line are untouched. The pane text replaces the screen decision in the shared `TerminalScreenContext.reconcile` truth table (vocab grounding, render demand, rendered block, provenance) behind the same consent gate and sanitize/24k-cap pipeline as an AX read.
- Any `pane.read` failure/timeout/garbage falls back loudly to the pre-existing behavior (composite AX text, vocabulary-only, nothing attached). Consent withdrawal at stop drops the pane text rather than downgrading.

**2. iTerm2 + Terminal.app become first-class join targets with AppleScript screen context** (`cf835b9`)

Deliberately relaxes the Ghostty-only invariant (owner decision 2026-07-22). `TerminalScreenAllowlist` is split by capture ROUTE:

- `axCaptureBundleIDs` — Ghostty only (its single-`AXTextArea` grid is the only verified one); raw AX capture remains impossible for the new terminals.
- `appleScriptCaptureBundleIDs` — iTerm2 + Terminal.app: focused-TTY joins via per-terminal AppleScript readers (`AppleScriptTerminalTTYReader`, generalized from the Ghostty reader: same `/dev/tty` validation, abstain on every failure class, per-terminal one-shot Automation consent prewarm), and screen context ONLY from the AppleScript `contents` of the focused session/tab (`TerminalScreenAppleScriptReader` — visible screen, never `history`; answered by the terminal process itself, same trust class as the TTY read; per-pane clean by construction). Oversized raw replies abstain rather than truncate.
- Both routes share one downstream pipeline (sanitization, caps, start/stop reconcile, vocab-always / raw-excerpt-only-after-authorized-join). The herdr probe needs only a surface TTY string, so the herdr arm now works on all three terminals; a herdr join still never attaches surface text on any terminal.
- Terminal.app's undocumented scripting properties verified from its sdef on the build host, verbatim: tab `tty` ("The tab's TTY device.", code `ttty`), tab `contents` ("The currently visible contents of the tab.", code `pcnt`), vs `history` (entire scrollback — not used).

Also: AGENTS.md invariants updated for both features; `scripts/ci/llm-lane-filter.sh` gains `*TerminalScreenAppleScriptReader*`, `*TerminalFocusedTTYReader*` (previously unmatched despite deciding the join), and `Sources/localvoxtral/DictationViewModel.swift` — this diff self-triggers the lane (`run=true` verified).

## Proof

- [x] Full unit suite green on the merged tree (Mac build host):
  `Executed 1990 tests, with 2 tests skipped and 0 failures (0 unexpected) in 68.440 (68.535) seconds` (skips are the pre-existing env-gated pair)
- [x] Tier-1 realtime integration green on the merged tree:
  `Executed 7 tests, with 2 tests skipped and 0 failures (0 unexpected) in 19.874 (19.876) seconds`
- New-behavior tests (all green in the run above):
  - `HerdrPaneScreenContextTests` (11): `testHerdrJoinRendersPaneReadExcerptAtCommit`, `testPaneReadRequestsExactlyTheJoinedPaneID`, `testCompositeAXNeverAttachesForHerdrJoinEvenWhenPaneReadFails`, `testStopReadFailureFallsBackToCompositeVocabularyOnly`, `testPaneTextIsControlStrippedAndCapped`, `testWhitespaceOnlyPaneTextIsGarbageAndFallsBack`, `testChangedPaneTextDegradesToPaneVocabularyOnly`, `testDeadSessionAtStopWithholdsRenderKeepsPaneVocabulary`, `testPolicyRejectedAtStopDropsPaneText`, `testNonHerdrJoinNeverTriggersPaneRead`, `testCaptureGateRejectionMakesNoSocketRequest`
  - `HerdrSocketClientTests` (+5, real AF_UNIX one-shot server): wire shape, mismatched `pane_id` / wrong result type / error envelope abstentions, injected-deadline expiry
  - `TerminalScreenAppleScriptCaptureTests` (16): allowlist truth table, exactly-one-capture-route-per-bundle, visible-screen-never-history script chains, AppleScript-never-AX for new terminals (start and stop), AX sanitization-path parity, oversized-reply abstention, gate-precedes-script, unchanged-screen render authorization, TTY arm + herdr probe + pane join for new bundles, unsupported-bundle refusal before any seam
  - `AppleScriptTerminalTTYReaderTests` (7): per-terminal script chains, valid-TTY parse, abstention on every failure code and implausible replies, never-executes-for-unsupported-bundles
  - `TerminalAutomationConsentPrewarmTests`: prior behaviors generalized + per-terminal one-shot state
- LLM lanes: the diff matches the path filter (ClaudeContext/**, screen-context readers, DictationViewModel.swift), so the polishd lane runs per-PR; `[run-llm-eval]` is also set above. No prompt/model/sampling change — the diff changes *which* clean excerpt attaches, not how it is polished.
- Not yet verified by hand (needs the owner's Mac):
  - iTerm2 end-to-end (iTerm.app is not installed on the build host): TTY join, `contents` capture, first-run Automation prompt. The chain is encoded defensively — any AppleScript error abstains.
  - herdr `pane.read` live (same publisher-override recipe as the #181 hand-test; expect a pane excerpt to attach instead of `screen-vocab-only`).
  - New one-time TCC Automation prompts for iTerm2/Terminal.app on first prewarm.


T0mSIlver:
**Hand-test results (owner, 2026-07-22) + one field bug fixed in `609a240`:**

- ✅ herdr: pane.read screen context confirmed working live
- ✅ Terminal.app: confirmed working live
- 🐛 iTerm2: the Automation consent prompt appeared and **vanished before it could be answered**. Root cause: macOS dismisses the TCC consent sheet when the pending Apple event times out, and the pre-warm sent the per-dictation read script (`with timeout of 1 second`) — so the sheet lived one second. Ghostty/Terminal.app never showed this because their grants already existed on this machine; iTerm2 was the first sheet that actually had to be answered.

Fix in `609a240`: the consent probe now sends the identical focused-pane question under `with timeout of 600 seconds` on its own per-terminal executor (a parked sheet can never queue the shared read executor); per-dictation reads keep the 1 s bound. The probe also logs how consent settled (granted/declined) instead of discarding it.

Proof:
- Regression test failing before the fix (compile error — the probe API did not exist; the old probe was, structurally, the 1 s read): `error: generic parameter 'T' could not be inferred … AppleScriptTerminalTTYReaderTests.swift:65` against `cf835b9`
- Passing after: `AppleScriptTerminalTTYReaderTests` — `Executed 9 tests, with 0 failures`
- Full unit suite on the fixed tree: `Executed 1992 tests, with 2 tests skipped and 0 failures (0 unexpected) in 68.888 (68.983) seconds`

Remaining hand-test: iTerm2 end-to-end after re-granting — the sheet should now stay up; if it was already auto-denied by the earlier timeout, toggle localvoxtral under System Settings → Privacy & Security → Automation → iTerm2 (or remove the entry) before retrying with the new build.

PR #187, opened 2026-07-24
CI: skip scheduled self-hosted lanes when the MacBook is on battery

T0mSIlver:
## What

The self-hosted runner is the owner's personal MacBook: scheduled lanes must not run while it is unplugged (owner request, 2026-07-24). Manual dispatches (and the `needs-ui-smoke` label) are explicit operator intent and always run.

- New `scripts/ci/ac-power-guard.sh`: probes `pmset -g ps` and emits `run=`/`reason=` GitHub-output lines. Battery (or UPS) power skips; AC runs; a missing/failed/unrecognized probe **fails open** into a run, same philosophy as the existing lane filters — a broken probe must never silently disable a lane. Test seam: `AC_POWER_GUARD_STATE=ac|battery|error`.
- `ui-smoke.yml` (evening AX-drill ladder): no workflow logic change — `ui-smoke-guard.sh` consults the power guard as its first rule, before the gh coverage query and the lock probe (cheapest probe first; a battery skip makes no API calls).
- `eval-e2e.yml` (nightly): new guard step under `if: github.event_name == 'schedule'`, writing `RUNNING (…)`/`SKIPPED (…)` to the step summary; every later step is gated with the fail-open `!= 'false'` comparison copied from ui-smoke.yml. The `always()` steps (scoreboard, artifact, leak check) are also gated so a battery-skip doesn't emit a misleading "no eval log produced".
- These are the only two workflows with `schedule:` triggers; everything else is manual or push/PR driven and unaffected.
- Tests: new `scripts/ci/test-ac-power-guard.sh` (decision logic + the real `pmset` parse via a PATH stub: multi-line battery output, UPS, pmset failure, unknown wording) wired into ci.yml's shell-test step; `test-ui-smoke-guard.sh` gains battery/power-error cases and pins `AC_POWER_GUARD_STATE=ac` for the pre-existing cases so the suite itself cannot flip when run on an unplugged Mac.
- AGENTS.md tier table updated for both lanes.

## Proof

Both guard suites pass (pure shell, stubbed probes — run on the Linux dev box; the same scripts run in ci.yml's shell-test step on both runners):

```
$ ./scripts/ci/test-ac-power-guard.sh
PASS: AC power runs (runner is on AC power)
PASS: battery power skips (runner is on battery power — scheduled lanes wait for AC)
PASS: probe error fails open into a run (power probe unavailable (error) — failing open)
PASS: pmset AC Power parses as ac (runner is on AC power)
PASS: pmset Battery Power (with detail lines) parses as battery (runner is on battery power — scheduled lanes wait for AC)
PASS: pmset UPS Power parses as battery (runner is on battery power — scheduled lanes wait for AC)
PASS: pmset failure fails open (power probe unavailable (error) — failing open)
PASS: unrecognized pmset wording fails open (power probe unavailable (error) — failing open)
ac-power-guard tests passed

$ ./scripts/ci/test-ui-smoke-guard.sh
PASS: battery power skips regardless of coverage and lock state (runner is on battery power — scheduled lanes wait for AC)
PASS: power probe error falls through to an unlocked run (screen unlocked, no recent successful run)
PASS: power probe error still respects a locked screen (screen is locked — the AX window drill would false-red; the next slot retries)
PASS: recent success skips regardless of lock state (drill ran and passed 1 h ago — today is already covered)
... (all 13 cases pass)
ui-smoke-guard tests passed
```

All three touched workflows parse as valid YAML (`python3 -c "import yaml; yaml.safe_load(...)"` — OK for eval-e2e.yml, ui-smoke.yml, ci.yml).

Live-validated on real macOS (the self-hosted Mac was offline, so the probe ran on GitHub-hosted runners via a throwaway branch, since deleted — runs 30118823335 / 30118881291). Both **macOS 26.4** (the runner's major version) and **macOS 15**:

```
$ pmset -g ps
Now drawing from 'AC Power'      # exact wording the parser matches; cat -v clean, rc=0

$ ./scripts/ci/ac-power-guard.sh   # unstubbed, AC_POWER_GUARD_STATE unset
run=true
reason=runner is on AC power
```

Both guard suites also pass there under `/bin/bash` 3.2.57 with BSD `sed`/`head` — the macOS shell/tooling combination the Linux dev-box runs could not cover.

Still not testable without the Mac: the literal `'Battery Power'` first line from a genuinely unplugged machine (hosted runners have no battery). It is the same pmset formatter that produced the AC string verbatim, and a wording miss falls OPEN into a run — the failure mode is "battery saving doesn't engage", never "the nightly silently dies". The next scheduled run's step summary shows the RUNNING/SKIPPED decision either way.

- [x] Unit suite: no Swift code touched (workflows + `scripts/ci/**` only); the shell-test step in this PR's own CI run executes both new/updated guard suites.
- [x] LLM lanes: skipped — CI/scheduling change only, nothing that reaches the model.


T0mSIlver:
## opencode review (GLM-5.2, read-only reviewer agent)

Verdict: **APPROVE** — no blockers/majors. It traced every event path and confirmed: manual dispatch/label runs are never blocked (all gates short-circuit on `github.event_name != 'schedule'` before reading the guard output), battery scheduled runs are fully suppressed including the `always()` steps, the `!= 'false'` fail-open idiom is applied consistently, and a battery skip cannot poison the ui-smoke ladder dedup (it keys on the drill step's conclusion, which is `skipped`).

Two minor findings, both folded into 52de5bd:

1. **`pipefail | head` clobber (latent)** — `first_line="$(pmset -g ps 2>/dev/null | head -n 1)" || first_line=""`: if pmset's output outgrew the 64 KB pipe buffer, the writer dies of SIGPIPE, the pipeline exits non-zero under `pipefail`, and the recovery clobbers a *successfully captured battery line* into a fail-open run — the one wrong fail direction for this guard. Doesn't manifest with real `pmset -g ps` (~120 bytes), but fixed via whole-capture + parameter expansion, with a regression case.

   Failing before the fix (fe2d691):
   ```
   FAIL: pmset battery line survives 80 KB of trailing output: expected run=false, got run=true (power probe unavailable (error) — failing open)
   ```
   Passing after (52de5bd):
   ```
   PASS: pmset battery line survives 80 KB of trailing output (runner is on battery power — scheduled lanes wait for AC)
   ```
2. **Doc drift** — `.github/workflows/README.md`'s ui-smoke guard sentence didn't mention the new battery rule. Updated.

Plus two hardening nits it suggested: the pmset-parse test cases now pin `AC_POWER_GUARD_STATE=` empty (an exported seam in a dev shell can no longer false-green them — verified by running the suite with `AC_POWER_GUARD_STATE=battery` exported), and the guard's header now states explicitly that fail-open means a scheduled lane may still run unplugged when the probe is broken ("never on battery" is best-effort by design, traded for "never silently dead").

Full suites after the fixes: 9/9 ac-power-guard, 13/13 ui-smoke-guard.

---
Note on CI: the first run's `build-test` failed on 4 pre-existing locked-screen host-state flakes (`DictationViewModelFailFastUXTests`/`TerminalTargetDetectorTests` Secure-Input/Accessibility UX tests — hotkey-monitor install fails while the runner Mac's screen is locked; same family as #122). Unrelated to this diff — the shell-test step running both new guard suites passed. Needs a green run once the Mac is unlocked.

PR #194, opened 2026-07-26
Enrollment sheet: confirm and report inside the section that asked

T0mSIlver:
## What

Field report (owner, 2026-07-26): tried the ssh enrollment flow, confirmed the insert, and believed nothing was written to `~/.ssh/config`. Two things were actually true:

1. **The insertion worked.** Diagnosed via `mac-crashlog.yml` first (per the runbook): the app log shows `Claude remote ssh config insertion completed` at 20:18:18 **and** 20:19:17 — the owner confirmed twice because there was no visible feedback. A new redacted ssh-config probe (added to `mac-crashlog.yml` in this PR) then proved the block on disk: markers `# BEGIN/END localvoxtral claude context (h9b61d528)` at lines 105–115 of `~/.ssh/config`, `RemoteForward 8473` present, file mtime `Jul 26 20:19:17` = the second insertion. The block lands at the END of a long config, and the sole feedback was a small caption pooled below step 2 — invisible in practice.
2. **The confirmation UX was misplaced by design.** Both confirmations and all step results rendered under step 2 regardless of which section's button requested them.

Changes:

- The confirmation now renders inside the section whose button requested it: the action button swaps to `Cancel` / `Confirm Insert` (or `Confirm Run`) right where the user clicked, with the exact text highlighted. The second-explicit-confirmation invariant holds — the displayed body IS the confirmation preview (plan snippet for step 1, token-redacted commands for step 2; asserted equal by existing tests).
- Step results render inside the same section. The model tags results with the action that produced them (`enrollmentResultsAction`), cleared when a new confirmation starts, so a stale step-1 result can never render while step 2 is being confirmed.
- Success copy names the file: `Inserted this host's block into ~/.ssh/config.`
- `mac-crashlog.yml`: new "Claude remote ssh config state" step — lstat + marker line numbers + `RemoteForward 8473` count + leftover atomic-write temp files. Never prints config content; Host aliases/hostnames stay private (public Actions log).

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh test`:
  ```
  Executed 1998 tests, with 2 tests skipped and 0 failures (0 unexpected) in 74.461 (74.556) seconds
  ```
- [x] Regression test added — `testStepResultsAreTaggedWithTheActionThatProducedThem` plus tagging assertions in the three existing enrollment tests. A literal failing-before run is impossible for the model half (the test exercises `enrollmentResultsAction`, which did not exist — it fails by compile error against the old model); the field evidence above is the failing "before": two confirmed insertions with the outcome rendered where the owner never saw it.
  ```
  Test Case 'testStepResultsAreTaggedWithTheActionThatProducedThem' passed (0.000 seconds).
  Test Suite 'ClaudeIntegrationSettingsModelTests' passed — Executed 24 tests, with 0 failures
  ```
- [x] Diagnostic probe verified live — dispatched `mac-crashlog.yml` from this branch (run 30214904868):
  ```
  ~/.ssh/config: type=Regular File perms=0644 owner=tom mtime=Jul 26 20:19:17 2026 size=2689
  105:# BEGIN localvoxtral claude context (h9b61d528)
  115:# END localvoxtral claude context (h9b61d528)
  --- RemoteForward 8473 present: 1
  ```
- [ ] UI change: hand-verification pending — I have no GUI access; owner loop is `./scripts/try-pr.sh <this PR#>` → Settings → Claude → enroll a host → click "Insert into ~/.ssh/config": Cancel/Confirm should appear under step 1 with the snippet highlighted, and the result line should appear there too (same for step 2's "Run on SSH host").

LLM lanes: skipped — settings-sheet UI + model state tagging + a diagnostic workflow; nothing on the `llm-lane-filter.sh` paths, nothing that reaches the model input.

### DIFF
--- a/.github/workflows/mac-crashlog.yml
+++ b/.github/workflows/mac-crashlog.yml
@@ -113,6 +113,25 @@ jobs:
           probe_model "polishd" "$polish_repo" "$polish_revision"
           exit 0
 
+      - name: Claude remote ssh config state (markers only)
+        run: |
+          # Field-debugging the enrollment insertion: report only what proves
+          # or disproves that the app's marked block is present. Never print
+          # the config's content — Host aliases and hostnames are private and
+          # this log is public. Marker lines carry only the random host id.
+          cfg="$HOME/.ssh/config"
+          echo "--- lstat ~/.ssh and ~/.ssh/config:"
+          /usr/bin/stat -f '%N: type=%HT perms=%Mp%Lp owner=%Su mtime=%Sm size=%z' \
+            "$HOME/.ssh" "$cfg" 2>&1 | sed "s|$HOME|~|g"
+          echo "--- localvoxtral marker lines (line number + marker only):"
+          grep -n '^# \(BEGIN\|END\) localvoxtral claude context' "$cfg" 2>/dev/null \
+            || echo "<no localvoxtral markers>"
+          echo "--- RemoteForward 8473 present:"
+          grep -c 'RemoteForward 8473' "$cfg" 2>/dev/null || echo "0"
+          echo "--- leftover atomic-write temp files:"
+          ls "$HOME/.ssh/".config.localvoxtral.* 2>/dev/null | sed "s|$HOME|~|g" \
+            || echo "<none>"
+
       - name: List running localvoxtral instances
         run: |
           pids=$(pgrep -x localvoxtral || true)
--- a/.github/workflows/mac-crashlog.yml
+++ b/.github/workflows/mac-crashlog.yml
@@ -116,21 +116,42 @@ jobs:
       - name: Claude remote ssh config state (markers only)
         run: |
           # Field-debugging the enrollment insertion: report only what proves
-          # or disproves that the app's marked block is present. Never print
-          # the config's content — Host aliases and hostnames are private and
-          # this log is public. Marker lines carry only the random host id.
+          # or disproves that the app's marked block is present. This log is
+          # PUBLIC: never print config content, filenames, aliases, hostnames,
+          # or account names. Markers print only when they match the app's
+          # exact grammar (grep -o keeps the matched text alone, so a
+          # marker-like line carrying extra private text is counted, never
+          # shown). A diagnostic must not go red on the states it exists to
+          # characterize (missing file, symlink), hence set +e.
+          set +e
           cfg="$HOME/.ssh/config"
-          echo "--- lstat ~/.ssh and ~/.ssh/config:"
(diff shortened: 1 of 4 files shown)


T0mSIlver:
**Two independent reviews** (owner-requested): codex `gpt-5.6-sol` (high reasoning effort) and opencode `glm-5.2` (opencode has no OpenAI auth, so gpt-5.6-sol could not run through it — glm-5.2 is its only real model). Both read-only over `origin/main...HEAD`. Findings and dispositions, fixed in a47f56e:

**codex/gpt-5.6-sol:**
1. *major, confirmed* — late-result guard compared only `host.id`, which rotation **reuses**: a setup still in flight when the sheet went away could publish its old-token outcome under the rotation sheet. Fixed: guard requires whole-presentation equality (rotation mints a new token, so value equality distinguishes generations). Regression test `testALateResultFromBeforeARotationNeverSurfacesUnderTheNewToken` — **failing before** the fix:
   ```
   error: … XCTAssertTrue failed - an old-token result must not render under the rotation sheet
   error: … XCTAssertNil failed: "runRemoteSetup"
   ```
   passing after (suite now 1999 tests, 0 failures).
2. *blocker, confirmed* — the probe printed whole marker-like lines to a public log; a malformed marker carrying private text would leak. Fixed: strict-grammar `grep -o` prints only exact matches; marker-like-but-malformed lines are counted, never shown.
3. *major, confirmed* — `ls` printed temp filenames verbatim. Fixed: count only.

**opencode/glm-5.2:**
1. *refuted premise* — claimed the `stat` pipeline aborts under `bash -eo pipefail`; GitHub's default for `run:` is `bash -e` **without** pipefail (visible in the run's own log header), and the pipeline exits with sed's status. Hardened anyway with `set +e` — a diagnostic must not go red on the states it exists to characterize.
2. *minor, confirmed* — `stat %Su` printed the bare macOS account name past the `$HOME` path redaction (visible in the first live probe run). Fixed: owner-is-current-user yes/no instead (the actually diagnostic bit — it is what the `sshDirectoryNotTrusted` gate checks).
3. *nit, confirmed* — `grep -c … || echo 0` double-printed `0` on no-match. Fixed.

Rewritten probe verified live on the runner (run 30218550225): strict markers at 105/115, `marker-like lines: 2 (well-formed: 2)`, counts only, no username, step green.

CI note: the first run of a47f56e failed with exit 124 — xctest wedged in `BackendManagerTests testStopAllDropsSupervisorsSoNextEnsureRebuildsArguments` (my suites never got to run; the host was simultaneously running a remote-build suite and a queued workflow). Same family as the documented supervisor-stop under-load flake; rerun on the quiet host is green.

PR #195, opened 2026-07-26
Remote setup: resolve claude under sshd's minimal PATH, drop setup forwarding

T0mSIlver:
## What

Field report (owner, 2026-07-26, second round of hand-testing): "Run on SSH host" failed with `/bin/sh: 2: claude: not found` although claude is installed on the host, plus a scary `Warning: remote port forwarding failed for listen port 8473` in the error output.

Diagnosis: the error format is **dash's** (Debian-family `/bin/sh`), and `ssh <alias> /bin/sh -s` runs under sshd's minimal PATH with no login rc — so user-local install dirs (`~/.local/bin`, where claude actually lives on the reporting host) are absent. The forwarding warning was noise: setup's own ssh connection tried the 8473 RemoteForward from the inserted config block while the user's real session already held it — setup never needed the tunnel at all.

- `remoteScript` now prepends a POSIX resolver for claude-invoking commands: probes `~/.claude/local`, `~/.local/bin`, `~/bin`, `/opt/homebrew/bin`, `/usr/local/bin`, `~/.nvm/versions/node/*/bin` (mirroring `ClaudePluginInstallService.claudeCLICandidates`), then fails with an actionable message + exit 127 instead of dash's bare one. Token-free by construction, same class as the already-injected `set -eu`; a plan note discloses it.
- Setup connections pass `-o ClearAllForwardings=yes` — no tunnel competition with the user's live session, no misleading warning inside setup errors.
- Non-claude commands stay unguarded (`testRemoteScriptLeavesNonClaudeCommandsUnguarded`).

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh test`:
  ```
  Executed 2009 tests, with 2 tests skipped and 0 failures (0 unexpected) in 70.853 (70.956) seconds
  ```
- [x] Regression tests added — `testRemoteScriptResolvesClaudeFromUserLocalInstallLocations` + `testRemoteScriptLeavesNonClaudeCommandsUnguarded`; existing argv/stdin equality tests updated to pin the new invocation. (The "failing before" state is the field error itself: the old script was literally `set -eu\n<command>` with no resolution.)
- [x] Resolver verified under real dash (Debian `/bin/sh -> dash`) with `PATH=/usr/bin:/bin`:
  ```
  === claude in ~/.local/bin:  FAKE-CLAUDE plugin list / exit=0
  === claude absent:           localvoxtral: 'claude' was not found ... / exit=127
  ```
  and on the reporting host with the real binary under `env -i HOME=$HOME PATH=/usr/bin:/bin dash -s`:
  ```
  2.1.220 (Claude Code)
  exit=0
  ```
- [ ] Hand-verify: `./scripts/try-pr.sh <this PR#>` → enroll/rotate → "Run on SSH host" against the real host should now pass step 1 (or fail with the actionable message, never dash's).

LLM lanes: skipped — remote-enrollment execution path only; nothing on `llm-lane-filter.sh` paths, nothing reaches the model.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteEnrollmentService.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteEnrollmentService.swift
@@ -286,6 +286,10 @@ public struct ClaudeRemoteEnrollmentService: Sendable {
             "A second concurrent SSH session to the same host will fail to bind \(port) on the "
                 + "remote and — because ExitOnForwardFailure is `no` — will connect anyway with no "
                 + "tunnel. The first session keeps the forward.",
+            "One-click setup connects with forwarding disabled (the tunnel belongs to your real "
+                + "sessions, not setup) and first resolves `claude` from common install locations — "
+                + "non-interactive SSH shells often lack the user-local PATH entries an interactive "
+                + "login has.",
         ]
     }
 
@@ -410,7 +414,8 @@ public struct ClaudeRemoteEnrollmentService: Sendable {
     ///
     /// Throws `.executionNotConfigured` when no runner was supplied. Each plan
     /// command is sent to `/bin/sh -s` over SSH stdin. The only spawned argv is
-    /// `ssh -o BatchMode=yes <alias> /bin/sh -s`, which contains no token.
+    /// `ssh -o BatchMode=yes -o ClearAllForwardings=yes <alias> /bin/sh -s`,
+    /// which contains no token.
     ///
     /// - Parameter token: the plaintext, used ONLY to redact it back out of any
     ///   failure. Nothing here logs or stores it.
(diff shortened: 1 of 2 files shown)


T0mSIlver:
Second field round folded in (767742c): steps 1–2 now succeed via the resolver, and the Verify block was the remaining confusion —

- `remote port forwarding failed` from the probe sessions is **expected** while the user's live session holds the 8473 forward (by design, first session keeps it). The verify output never said so.
- `zsh:1: command not found: claude` — plain `ssh <host> 'cmd'` runs the login shell non-interactively (no rc files): the setup-path disease in its user-run form. Reproduced and fixed with a PATH prefix, proven against the reporting host's real claude: bare non-interactive zsh with `PATH=/usr/bin:/bin` → exit 127; prefixed → `2.1.220 (Claude Code)`.
- `401` from the tunnel curl is the **success signal** and is now labeled as such.

Verify commands now carry their interpretation as `#` lines (uninstall block precedent) and the plugin-list probe prefixes the common install dirs. `testVerifyCommandsProbeTheTunnelAndThePlugin` pins all three. Suite green per-commit in CI.

PR #196, opened 2026-07-27
Remote plugin: convert http hooks to a curl command shim (token actually reaches the wire)

T0mSIlver:
## What

The `localvoxtral-remote` plugin's http hooks have **never authenticated**: Claude Code expands http-hook header `${VAR}` references from the actual process environment only and never injects plugin userConfig options there (verified empirically on 2.1.220 with a capture server — the header arrives as literally `Bearer `). Every remote hook was 401'd forever; users saw `UserPromptSubmit hook error: HTTP 401` on every prompt. Upstream issue: anthropics/claude-code#81742.

Command hooks ARE the surface that receives `CLAUDE_PLUGIN_OPTION_TOKEN`, so this converts the plugin's five hooks to a bundled POSIX-sh shim (`hooks/post.sh`) that curls the event JSON to the same `127.0.0.1:8473/v1/hook/<Event>` endpoints:

- **Token never in argv** (`/proc/<pid>/cmdline` is world-readable): header reaches curl via a 0600 tempfile (`--header @file`), written by a redirected heredoc (external `printf`/`echo` would leak into argv), cleaned by EXIT/HUP/INT/TERM traps.
- **Fail-open everywhere**: missing curl, unset/empty token, tunnel down, timeout, non-200 → silent exit 0, curl `--max-time 1` (same ceiling as the old http hooks). Dictation context being unavailable must never block or warn a Claude turn.
- **Response passthrough**: on 200 the listener's allowlisted `ClaudeHookOutput` body (marker `terminalSequence` / `suppressOutput`) prints to stdout untouched; anything else prints nothing (a UserPromptSubmit hook's non-JSON stdout would be appended to the user's prompt).
- Plugin 1.0.0 → 1.1.0; name and `token` userConfig key unchanged, so enrollment's generated `claude plugin install … --config 'token=…'` keeps working as-is. Docs (AGENTS.md, READMEs, enrollment caveats, listener/wire doc comments) updated honestly: still no localvoxtral binary and no jq/nc/Node on the remote host, but `sh` + `curl` are now required (fail-open when absent).

`ClaudeRemotePluginManifestTests` rewritten (26 tests): all prior guarantees re-pinned against the shim (loopback URL + per-event paths asserted against `ClaudeRemoteHTTPCodec` constants, 8471/8472 collision, no literal credential), plus new pins — every hook is `command` with no http hook anywhere, shim is the only executable, `#!/bin/sh`, token-out-of-argv source pins (`--header @`, no `Bearer $` on a curl line, `umask 077`), `--max-time 1`, quoted `${CLAUDE_PLUGIN_ROOT}`, and runtime fail-open tests executing the shim under `/bin/sh` on network-free paths.

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh` (full build + unit):
  ```
  Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-07-27 21:15:54.332.
       Executed 2012 tests, with 2 tests skipped and 0 failures (0 unexpected) in 68.037 (68.152) seconds
  ```
- [x] Bug fix: regression tests added — failing before the fix, passing after. Ran the new `ClaudeRemotePluginManifestTests` against the pre-fix plugin (old `hooks.json`/`plugin.json` checked out, shim removed):
  ```
  Test Suite 'ClaudeRemotePluginManifestTests' failed at 2026-07-27 21:16:32.737.
       Executed 26 tests, with 19 failures (7 unexpected) in 0.339 (0.341) seconds
  ClaudeRemotePluginManifestTests.swift:176: XCTAssertEqual failed: ("Optional("http")") is not equal to ("Optional("command")") - UserPromptSubmit must be a command hook
  ClaudeRemotePluginManifestTests.swift:177: XCTAssertNil failed: "http://127.0.0.1:8473/v1/hook/UserPromptSubmit" - UserPromptSubmit: an http-hook url would never authenticate
  ```
  Passing after (part of the full-suite run above):
  ```
  Test Suite 'ClaudeRemotePluginManifestTests' passed at 2026-07-27 21:15:00.279.
  ```
- [x] **Live end-to-end against the production listener** (Linux remote host, real SSH `RemoteForward` tunnel, real enrolled token — the exact setup that 401'd before):
  ```
  $ echo '{"hook_event_name":"UserPromptSubmit","session_id":"post-sh-live-test",...}' | CLAUDE_PLUGIN_OPTION_TOKEN=<real> sh post.sh UserPromptSubmit
  {"suppressOutput":true,"terminalSequence":"]2;lvx-441e1124"}   # exit 0
  $ … CLAUDE_PLUGIN_OPTION_TOKEN=<wrong> sh post.sh UserPromptSubmit
  (no output)                                                                # exit 0 — fail-open
  ```
- [x] **Claude Code 2.1.220 harness verification** (throwaway plugin + capture server on 127.0.0.1:8474, Linux):
  - Root cause reproduced first: identical http hook with the option set arrives as `Authorization: 'Bearer'` (empty); a process-env control expands, proving header expansion is env-only.
  - Converted plugin installed via `claude plugin install … --config 'token=cmdtest-secret-42'`; `claude -p` fired UserPromptSubmit/Stop/SessionEnd, each captured with `Authorization: Bearer cmdtest-secret-42` and the full event JSON; the 200 body `{"suppressOutput":true}` passed through with clean model output — proves 2.1.220 injects the `sensitive` option into command-hook env.
  - Argv scan while the server stalled 3 s/request: 26 sweeps of `ps -eo args` + `/proc/*/cmdline`, 10 in-flight curl sightings (positive control), **0 token leaks**.
  - Fail-open matrix under sh/dash/bash: server down, token unset (no `--config` install), curl missing, connection refused — all exit 0, zero stdout/stderr, and with no token the shim bails before dialing (zero POSTs logged).
- [x] LLM lanes: not required in substance — the `Sources/**` ClaudeContext changes are doc-comment-only (no change to what reaches the model); the path filter may still run the lane, which is fine.

## Reviewer notes (Mac-side)

- Worth one packaged-app check that `package_app.sh`'s copy preserves the shim's executable bit into the bundled marketplace.
- Remaining hand-test: marker `terminalSequence` riding back into the terminal and a real remote dictation join after `claude plugin update` on an enrolled host (the wire-level 401→200 half is proven above).
- Post-merge, enrolled hosts pick the fix up with `claude plugin update localvoxtral-remote@localvoxtral` (or uninstall/reinstall via the enrollment plan) and re-enabling the plugin.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "Off-screen Claude Code session context for the localvoxtral dictation app. Add this marketplace from a REMOTE host, where there is no app bundle to register a local directory from: claude plugin marketplace add T0mSIlver/localvoxtral",
-    "version": "1.1.0"
+    "version": "1.2.0"
   },
   "plugins": [
     {
@@ -17,7 +17,7 @@
     {
       "name": "localvoxtral-remote",
       "source": "./integrations/claude-code/plugins/localvoxtral-remote",
-      "description": "Publishes Claude Code session context from a remote host to localvoxtral on your Mac over an SSH RemoteForward tunnel. Install this on the REMOTE host, not the Mac. Declarative HTTP hooks only — no binary, no shell shim, no dependencies. Fail-open: does nothing without a tunnel."
+      "description": "Publishes Claude Code session context from a remote host to localvoxtral on your Mac over an SSH RemoteForward tunnel. Install this on the REMOTE host, not the Mac. Command hooks running a bundled POSIX-sh shim that POSTs via curl — no localvoxtral binary, needs only sh and curl. Fail-open: does nothing without a tunnel, a token, or curl."
     }
   ]
 }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -389,8 +389,15 @@ Key subsystems:
     plugin userConfig options there (verified on 2.1.220), so every hook
     authenticated as `Bearer ` and was 401'd — command hooks are the only
     surface that receives `CLAUDE_PLUGIN_OPTION_TOKEN`. The shim keeps the
-    token out of every argv (`curl --header @tempfile`, 0600, heredoc-written)
-    and prints the listener's 200 body untouched. `ClaudeRemoteContextListener`
+    token out of every argv (`curl --header @tempfile`, 0600, heredoc-written),
+    and its stdout FAILS CLOSED — the mirror image of delivery failing open:
+    it prints a 200 body only when it matches exactly the one grammar the
+    listener can emit (`markerResponseBody` — `suppressOutput:true` plus an
+    optional lvx-marker `terminalSequence`), one line, size-capped; anything
+    else prints nothing. Command-hook stdout is appended to the user's prompt
+    when it is not control JSON (and `additionalContext` when it is), so
+    whatever answers on 8473 must never be able to put a byte into the prompt
+    (owner rule 2026-07-27). `ClaudeRemoteContextListener`
     (loopback-bound POSIX, dedicated port 8473; 8471/8472 remain the managed
     backends) authenticates the Bearer token *before retaining a body* against
     `ClaudeRemoteHostRegistry` (0600 atomic file, token hashes only,
@@ -554,7 +561,11 @@ Key subsystems:
   live), so a second enrollment cannot drop the first host's tunnel.
   A bind conflict is reported, never routed around onto another port: a
   squatter on 8473 receives the remote's bearer token before anything rejects
-  it, so the user must learn it is there. Note also what is NOT defensible: a
+  it, so the user must learn it is there. What the squatter does NOT get is a
+  path into the prompt: the remote shim's stdout gate (post.sh) rejects any
+  200 body that is not exactly the listener's allowlisted control JSON, and a
+  forged well-formed marker is inert because markers are broker-allocated —
+  an unknown one joins nothing. Note also what is NOT defensible: a
   malicious process running as the user on the REMOTE host can still read
   `~/.claude/` and therefore the plugin's token no matter what we do. Say so
   rather than implying the token bounds it.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -379,10 +379,18 @@ Key subsystems:
     `ClaudeContextBroker` verifies peer UID *before reading*, and only ever
     unlinks a socket it has PROVED stale by connect-probe — a second live
     instance owns its socket legitimately.
-  - **Remote** (`localvoxtral-remote`, installed on the REMOTE host): declares
-    `type: "http"` hooks, so Claude Code itself POSTs to
-    `127.0.0.1:8473/v1/hook/<Event>` through an OpenSSH `RemoteForward` — no
-    binary, no shim, no `jq`/`nc`/Node on that host. `ClaudeRemoteContextListener`
+  - **Remote** (`localvoxtral-remote`, installed on the REMOTE host): command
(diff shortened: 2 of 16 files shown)


T0mSIlver:
## Review (opencode/GLM-5.2) + fixes — `5e0c79f`

An independent GLM-5.2 review of this PR confirmed all six hard requirements on the legitimate path (token never in argv, fail-open incl. the no-curl/no-cat host under dash, POSIX-clean, enrollment-compatible, endpoints unchanged) and raised four findings. Disposition:

**Finding 1 (medium) — 200 body reached stdout unvalidated → prompt-injection channel for an 8473 squatter.** Fixed, fail-closed: the shim now prints a 200 body only when it is a single, ≤256-byte line matching exactly the grammar of `ClaudeRemoteHTTPCodec.markerResponseBody` (`suppressOutput:true` + optional `lvx-` marker `terminalSequence`); anything else prints nothing. `--max-filesize 1024` added as a belt. Owner rule recorded in AGENTS.md: absolutely nothing may be inserted into any user prompt — the squatter paragraph now states what a squatter does and does not get.

**Finding 2 (low/medium) — exact `Authorization: Bearer` line lost its positive pin.** Fixed: `testShimWritesTheExactBearerHeaderTheListenerParses` pins the heredoc line verbatim, restoring the guarantee the old `XCTAssertEqual(authorization, "Bearer ${…}")` carried.

**Finding 3 (low) — tempdir leak under SIGKILL.** Accepted and documented in the shim (0700/0600, private to the user, bounded by hook kills).

**Finding 4 (informational) — 1s→3s timeout semantics.** Already pinned by `testHooksAreSynchronousSoTheMarkerResponseReachesClaudeCode` (`--max-time 1` network ceiling, `timeout: 3` backstop); no change.

**Bonus bug caught during live re-verification** (not in the review): with a reset tunnel — curl 200-less, body file never written — the shell's own `cannot open` redirection error escaped a trailing `2>/dev/null` (input redirection fails before the stderr one is applied) and leaked to stderr on exactly the must-be-silent path. Fixed with `[ -r body ]` + `{ …; } 2>/dev/null` grouping, same ordering fix on the header heredoc, covered by the body-never-written stub case.

### Proof

Failing before, on the pre-fix shim (new tests, old `post.sh`):

```
Test Suite 'ClaudeRemotePluginManifestTests' failed at 2026-07-27 22:00:26.255.
     Executed 29 tests, with 8 failures (0 unexpected) in 4.307 (4.311) seconds
(all 8: testShimStdoutFailsClosedOnAnythingButTheExactAllowlistedBody — hostile bodies reached stdout)
```

Passing after, full suite:

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-07-27 22:05:57.098.
     Executed 2015 tests, with 2 tests skipped and 0 failures (0 unexpected) in 71.469 (71.568) seconds
```

Hostile-body matrix (stub curl, run against the real shim under sh and dash, plus the same cases as XCTest): legit marker body and no-marker body pass through **byte-for-byte** (fixtures generated by the real codec, so encoder drift fails the test); plain-text injection, `additionalContext` smuggle, valid-line-plus-rider, raw ESC bytes, extra key on an otherwise perfect body, `suppressOutput:false`, non-allowlist marker, oversized, empty, non-200-with-legit-body, and body-never-written are all **silent** (exit 0, no stdout, no stderr).

Live: the wire-level 401→200 fix was verified against the production listener through the real tunnel earlier today (marker `terminalSequence` returned); during the fix pass the tunnel happened to be down, and the final shim handled the real reset-tunnel condition with a silent exit 0 — the exact live path the stderr-leak fix is for. LLM lanes: still doc-comment-only in `Sources/**`; no change to what reaches the model.

PR #201, opened 2026-07-27
Settings: rows read like macOS System Settings rows

T0mSIlver:
## What

**Stacked on #199 — this PR targets `settings-sidebar-chrome`, not `main`.**
Do not delete #199's branch before this one is retargeted; GitHub auto-closes
children when a base branch disappears. Review #199 first.

Presentation-only rewrite of the shared row/group primitives in
`SettingsView.swift`. Same controls, same bindings, same behavior — rendered in
the row idiom the rest of macOS uses: **label leading, control trailing,
explanation on its own full-width line underneath.**

The old row put the label in a fixed 128pt column with everything else to its
right. Long labels ("Use Claude Code project files as polish context") wrapped
inside that column while short ones ("Enable") left a gutter, and every help
sentence started at the column's edge — so each card had a second, ragged left
margin. Now the label takes the leftover width (`layoutPriority(0)`, so controls
keep their intrinsic size) and the explanation is a row of its own, aligned to
the label.

Also:

- Page-level `.toggleStyle(.switch)`, so 12 call sites stop repeating it.
- `SettingsGroup`: corner radius 18 → 8, `.quinary` fill with a `.quaternary`
  border, `.headline` header (down from 17pt semibold), and a `Divider()` between
  rows.
- The last divider is removed **geometrically** — the row container is laid out
  1pt short and clipped, so the final divider falls outside the clip. The
  alternative was teaching the card to enumerate its children, which are
  heterogeneous and arrive wrapped in `Group`/`if` branches.

### Deviations from the brief (with reasons)

1. **Help text became a `help:` parameter — the one call-site change.** The brief
   asked for no call-site changes, but most call sites nest the help text inside
   their own `VStack`, and a row cannot pull a nested view out of its control
   column — which is exactly where this text must not be. The zero-call-site
   alternative is hoisting the string with a `PreferenceKey`, which puts every
   row's layout one frame behind its content and fights Swift 6's `@Sendable`
   `onPreferenceChange` (the closure would capture a non-`Sendable` generic
   `View`). The migration is mechanical, and it *shrinks* the call sites — the
   wrapper `VStack`s disappear.
2. **No page-level `.controlSize(.mini)`.** Applied at page level it shrinks every
   button, picker, and text field on the page, not just switches — and macOS
   System Settings uses regular-size switches. Says so here rather than shipping
   a pane of miniature buttons; happy to reverse if the owner wants the smaller
   switch specifically.
3. **Added `layout: .stacked` for composite rows** (see below). Not in the brief,
   but it is the fix for the exact rows the brief asked me to hand-check.
4. Row label/control alignment stays `.top` rather than centering: several rows
   have tall controls, and a centered short label beside those reads as
   misaligned.

### The composite rows the brief flagged — what I found and did

Hand-check here = careful code reading plus a clean build; none of it has been
seen on screen (see below).

| Row | Under trailing alignment | What I did |
|---|---|---|
| `ShortcutRecorderField` + Reset/Clear button (`:578-632`) | Fine — fixed 132pt field + button is a bounded, intrinsically-sized HStack. The `SettingsInlineMessage` ("Not set. Record one to enable.") now sits under the control, trailing-aligned, instead of under the label. | Left as-is; validation text belongs next to the control it validates. Flagged for the owner's eye. |
| Sliders (`:558`, `:651`) | **Broken.** A `Slider` reports no intrinsic width, so in a trailing column it collapses to nothing. | Both get an explicit `SettingsLayout.sliderWidth` (190pt) track. |
| `ManagedBackendStatusRow` (`:386`) | Fine — dot + optional progress + short caption. | Unchanged. |
| `ClaudePluginSettingsRow` | **Broken.** Two buttons + spinner + result line ≈ 400pt of control; the label would be squeezed to a wrapped stub. | `layout: .stacked`. |
| `ClaudeRemoteHostsSettingsRow` | **Broken**, worse — host rows (3 buttons each) plus an enrollment form of two text fields and a button. | `layout: .stacked`. |
| `SettingsFileNotes` ("Included files") | Full-width list of file names, not a control. | `layout: .stacked`. |
| `PermissionRowsView` | Shared verbatim with the onboarding wizard, and the card no longer supplies inner padding. | Wrapped in `SettingsGroupRow` at the call site rather than modified — onboarding is untouched. |
| `SettingsAvailabilityCard` | Same padding issue. | Renders inside `SettingsGroupRow`; radius 16 → 8. |

Every consent-critical help string (the polish-context toggles, the non-local
endpoint toggle) is byte-identical, and the rationale comments that explain each
one moved with them.

## Proof

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-07-28 00:17:59.858.
	 Executed 2021 tests, with 2 tests skipped and 0 failures (0 unexpected) in 71.535 (71.637) seconds
```

No new warnings (`.build/last-remote.log` has only the pre-existing
`ClaudePluginInstallService.swift:226` and `ClaudeHookPublisherTests` ones).

CI `build-test` on this branch: pass (4m38s). #199 picked up review hardening
after this branch was cut (`da5d653`, drill scripts only — no app code), so this
branch is one commit behind its base; nothing in it touches the same files.

- **AX drill**: the drill's needles are group titles ("Permissions", "Polishing",
  "Start dictation with", …) and field labels ("Endpoint", "API key"), all
  unchanged by this PR, and #199's scoped selectors are unaffected. The drill
  could not run tonight — see #199's Proof for the two runs and why (an existing
  app instance on the runner's GUI session would not quit; the guard was not
  weakened).
- **LLM lanes**: not run, and not required — SwiftUI layout of existing controls.
  Nothing here changes prompts, model pins, sampling, or anything that reaches
  the model; `scripts/ci/llm-lane-filter.sh` matches nothing in the diff.

### UI change — what was verified, and what was not

Verified: clean build under Swift 6.2 strict concurrency, full unit suite, and
the row-by-row reading in the table above.

**Not verified — needs the owner's eyes** (`./scripts/try-pr.sh <this PR>`), in
priority order:

1. The **Text Processing** pane, which has the longest labels and the two stacked
   composite rows: do the toggles line up on the right, and does the explanation
   under each one read as belonging to it?
2. The **Dictation** pane's two shortcut rows and the 190pt sliders — the two
   spots where I changed geometry on a guess.
3. That the group cards show a divider between every row and **no** stray line
   above the bottom edge (the 1pt clip trick).
4. That `.quinary` on `.windowBackgroundColor` is still visible as a card in both
   light and dark appearance.


T0mSIlver:
## Manual UI review — CI artifact

Tested the exact prebuilt artifact with `./scripts/try-pr.sh 201` (CI run [30310509726](https://github.com/T0mSIlver/localvoxtral/actions/runs/30310509726)), using Computer Use against the native window.

Coverage: all five panes, managed and both External URL endpoint states, the full Text Processing scroll depth/composite rows, sliders, light/dark appearance, and card/divider edges.

### Blocking regression: External URL text-field rows collapse incorrectly

Switching Dictation to **External URL** makes the `Endpoint`, `Model`, and `API key` labels disappear visually and turns each field row into a very tall empty band. Switching Polishing to External URL does the same for its `Endpoint`, `API key`, and `Model` rows.

The AX tree still contains the labels, but the rendered window does not. The parent #199 artifact renders the exact same state compactly with all labels visible, so this is introduced by the row rewrite.

The likely pressure point is an `.inline` `SettingsFieldRow` whose trailing content is a horizontally flexible `TextField`: unlike an intrinsic-width picker/switch, it competes for the whole HStack width. I would either make these six text-field rows `.stacked`, or give the trailing field column an explicit bounded width. Please re-check both External URL modes after the fix.

### What passed

- Text Processing’s longest labels remain readable and the switches align cleanly at the trailing edge.
- The Claude plugin/remote-host composites and Included files list work well in stacked layout.
- Both 190pt sliders render at a usable size.
- Dividers appear between rows with no stray final divider at the bottom edge.
- The quinary cards remain visible in light and dark appearance.
- The colored purple pill at the extreme top-left of a few captures is macOS’s Computer Use/screen-recording indicator, **not localvoxtral UI**.

<details open>
<summary><strong>Bug evidence and parent comparison</strong></summary>

<table>
<tr>
<td><strong>#201 — Dictation External URL (broken)</strong><br></td>
<td><strong>#199 — same state (compact/labels visible)</strong><br></td>
</tr>
<tr>
<td colspan="2"><strong>#201 — both backends External URL</strong><br></td>
</tr>
</table>
</details>

<details>
<summary><strong>Full screenshot pass (11)</strong></summary>

<table>
<tr>
<td><strong>General</strong><br></td>
<td><strong>Dictation/sliders</strong><br></td>
</tr>
<tr>
<td><strong>Endpoints — managed/top</strong><br></td>
<td><strong>Endpoints — managed/bottom</strong><br></td>
</tr>
<tr>
<td><strong>Text Processing — top</strong><br></td>
<td><strong>Text Processing — middle</strong><br></td>
</tr>
<tr>
<td><strong>Composite Claude rows</strong><br></td>
<td><strong>Included files/bottom</strong><br></td>
</tr>
<tr>
<td><strong>Text Processing — dark/top</strong><br></td>
<td><strong>Text Processing — dark/bottom</strong><br></td>
</tr>
<tr>
<td><strong>About</strong><br></td>
<td></td>
</tr>
</table>
</details>

T0mSIlver:
## SSH enrollment flow — artifact hand-test

I repeated the complete SSH enrollment review in the CI-built #201 artifact via `try-pr.sh`: empty/partial/valid form states, one-time token, config and remote confirmation previews, verification, uninstall, and notes.

### UI findings

- **The stacked treatment is a substantial improvement over #199.** The section title gets a full line, the existing host and its actions are easy to parse, and the Name / SSH alias / Enroll row has usable widths. This is a strong concrete example of why the new row idiom helps.
- **The confirmation behavior is good:** both mutating setup actions have an explicit inline Cancel/Confirm stage. I opened both and cancelled them.
- **Remaining UX concern (shared sheet):** **Enroll…** commits the new host/token immediately, before either setup step is performed, but the sheet exits with a generic **Done**. A user can finish with a permanently enrolled-but-unconfigured host without realizing the credential was already created. “Create enrollment…” / “Finish later” or explicit persisted-state copy would clarify this.
- **The Verify / Uninstall / Notes area remains hard to scan.** It is a long wall of small gray text in a fixed-height sheet. Short subsections or disclosures would preserve the detail without making the primary flow feel like a runbook.
- This flow did **not** reproduce the disappearing-label/giant-empty-row blocker from my earlier #201 comment; that remains specific to the External URL endpoint rows.

No SSH config was inserted and no remote command was run. The temporary host registry was restored byte-for-byte and both test tokens were invalidated **before** these screenshots were uploaded.

### Screenshots

**Row states: empty → partial → ready**

SSH enrollment overview
Name entered; Enroll remains disabled
Valid form; Enroll enabled

**Enrollment sheet and confirmations**

One-time token and setup steps
SSH config confirmation preview
Remote setup confirmation preview

**Verification, uninstall, and notes**

Verification commands
Uninstall and notes
Final notes

T0mSIlver:
Fixed the blocking External URL regression from the manual UI review in f118ec6: text fields now get an explicit 280pt trailing column (same treatment as the sliders, which have the same no-intrinsic-width problem). Full unit suite green (2021 tests, 0 failures). Both External URL modes need a visual re-check — the AX tree was already correct, so only eyes can confirm the render.

T0mSIlver:
Merged the updated #199 (`settings-sidebar-chrome`, which now carries main's 21 commits: the remote-context campaign #216–#221). f118ec6 is kept.

**Conflict**: one hunk in `SettingsView.swift` — the "Use Claude Code project files as polish context" row, where main rewrote the help sentence and this branch moved that text to the `help:` parameter. Kept the row-idiom form and took main's new wording ("…a Claude Code session in a supported terminal, or a Remote Control session in the focused browser tab…").

**Ported into the row idiom** (main wrote these against the old fixed-column row, so they arrived as nested `VStack + SettingsHelpText` and were converted):
- **cmux join toggle** → inline row with `help:`, in the existing Polishing group (no new group).
- **`ClaudeCmuxPasswordSettingsRow`** → inline row with `help:`; the `SecureField` is bounded to `SettingsLayout.textFieldWidth` per f118ec6 (unbounded it starves the label), the Save button sits beside it, and the status line moved to its own line under the field rather than being a fourth item in a bar next to a 280pt field.
- **About → Build row** (dogfood variant) → inline row, with `DogfoodBuildStatus.detail` passed straight through as `help:` (it is already `String?`, which is what `help:` takes).

Untouched by the idiom, correctly: everything main added *inside* the `.stacked` composites — the per-host "Keep the tunnel open" toggle + status + Retry, `pluginUpdatePanel`, `ClaudeEnrollmentStepResults`, the rejection hint, the enrollment sheet's "Update later" section and no-alias copy-only path. A composite row's internals are its own layout; the row idiom governs the row.

**Proof**: `./scripts/remote-build.sh` — build clean, `Executed 2593 tests, with 2 tests skipped and 0 failures`. Pushed as 7d19203.

T0mSIlver:
Fixed the three findings from the final row-idiom review, all at the row-API level rather than at the call sites — pushed as 62d1e13.

**M1 — inline labels top-aligned next to short controls.** `SettingsFieldRow` now takes `controlAlignment: VerticalAlignment = .center`, and `inlineRow` uses it instead of a hardcoded `.top`. The ~10 rows whose control is a lone switch or picker now center the label against the control's centerline. The two shortcut-recorder rows (Overlay Buffer / Live Auto-Paste) — a 24pt bordered field plus a button, the tallest inline control in the pane — opt back in with `controlAlignment: .top`. Audited every other inline row: after M2 no other one has a multi-element or multi-line control (the slider rows, status rows, and the cmux field+button row are all single-line), so nothing else needed the opt-in. The comment on `inlineRow` states why the default flipped.

**M2 — status/validation text trapped in the trailing column.** `SettingsFieldRow` grows an optional `footer:` `@ViewBuilder` slot, rendered on its own full-width, **leading**-aligned line between the control row and the help text (status describes the control right above it; the help text explains the row as a whole and closes it). Rows built without a footer are `Footer == EmptyView` via a second init and a stored `hasFooter` flag, so their layout is byte-for-byte what it was. Three call sites moved into footers:

- Overlay Buffer shortcut row — "Not set. Record one to enable." / red validation error
- Live Auto-Paste shortcut row — same pair
- `ClaudeCmuxPasswordSettingsRow` — the "Password saved." status line

The `help:` doc comment now says explicitly that `help:` is the static explanation and `footer:` is the dynamic per-row status.

**M3 — fixed 280pt fields cannot shrink.** All seven call sites (six External URL fields + the cmux `SecureField`) moved from `.frame(width:)` to `.frame(maxWidth: SettingsLayout.textFieldWidth)`. Behaviour at full width is unchanged: the trailing column has `layoutPriority(1)` and a `TextField` is greedy, so it still reports and fills the full 280pt cap wherever the card is wide enough — the cap only ever binds when the card gets narrower, and then the field yields instead of the label. The constant's comment now says it is a cap and must be applied with `maxWidth`.

**Window-width verdict:** the Settings window is **fixed-width by design**, not resizable-with-no-minimum — `localvoxtralApp.swift` puts `.frame(width: 780)` plus `.frame(minHeight: 480, idealHeight: 560, maxHeight: .infinity)` on the scene content, with a documented rationale (fixed 208pt sidebar, horizontal resize only makes the panes worse) and a deliberate note against `.windowResizability(.contentSize)`. So there is no missing `minSize` and I changed nothing there. M3 stands on its own as robustness for the layout, not as a fix for a reachable state today.

**L1** — added a comment at `SettingsLayout.rowVerticalPadding` cross-referencing `SettingsGroup`'s `-1pt + clip` last-divider trick, so a future padding at or under 4pt does not silently clip the last row's focus ring.

**Proof:** `./scripts/remote-build.sh` — **2593 tests, 2 skipped, 0 failures**. No warnings introduced (the only warnings in the log are the pre-existing `try?`-unused ones in `ClaudePluginInstallService.swift`). No test weakened and none needed updating: the row types are `private` to `SettingsView.swift`, and `CmuxSocketPasswordStoreTests` pins `model.cmuxStatusText`'s STRING, which is unchanged — only where it renders moved.

Visual re-verification on the Mac is still pending; these are layout changes and the alignment/footer calls should be eyeballed in the real window before merge.

T0mSIlver:
Closing: superseded by #274 (the refreshed stack of #199/#201/#223/#224, merged 2026-08) and the later Settings work in #275/#276/#282/#284.

PR #203, opened 2026-07-27
Wire v2: agent-tagged records, per-agent session namespacing, pane focus declarations

T0mSIlver:
## What

Groundwork for the opencode integration (localvoxtral's second coding agent): the wire and registry generalization, pure Swift, no plugin yet (that is the stacked child PR).

- **Wire v2** (`ClaudeHookWire.version = 2`): records gain an `agent` field (`claude` | `opencode`). It is omitted on the wire for `.claude`, so every Claude record keeps its v1 byte shape; v1 lines still decode and are Claude by construction (`readableVersions = {1, 2}`). Unknown agent names are dropped precisely (`.unknownAgent`), never defaulted — an agent this build has never heard of must not inherit Claude's channel rules (title markers, bare id namespace). `ClaudeBrokerResponse.decodeLine` accepts any readable version so a publisher one release apart from the app degrades to "no marker", not a dead channel.
- **Per-agent session namespacing** (`ClaudeAgentSessionScope`): opencode ids are scoped `opencode:ses_…` by the RECEIVER (`ClaudeSessionRegistry.ingest`), recomputed from the agent tag on every ingest — mirroring `ClaudeRemoteSessionScope`'s host namespacing. The agent is fixed at first sight like the origin; a record spelling another agent's scoped id is dropped whole.
- **`FocusChanged` (v2) + focus arbitration in `resolve(tty:)`**: one opencode process hosts several sessions on one TTY (the TUI shows one at a time), so per-session TTY claims structurally cannot disambiguate them. A pane may now declare which session it displays; the declaration is freshness-bounded (2 min TTL, table bounded and pruned), pid-cross-checked against the focused session's registered process, LOCAL transport only, and never mints a session. `resolve(tty:)` arbitrates: fresh focus resolves (and can disambiguate competing per-session claims); stale focus is ABSENT information; a fresh declaration for a dead/unknown/unverifiable session abstains outright; a live per-session claim conflicting with a fresh focus declaration abstains `.ambiguous`. Claude behavior with no focus entry is byte-for-byte the pre-v2 rules.
- The reducer never absorbs a focus record's process block into the session — pane state (the declarer's tty/pid) must not become a per-session TTY claim (caught by the new truth-table tests, fixed before first commit).
- herdr pane claim cross-check is agent-scoped (`scopedSessionID(agent:sessionID:)`) so the existing herdr arm keeps confirming — never redirecting — when the pane hosts opencode.

Downstream (join resolver arms, repo collector, context blocks, budget) is unchanged: opencode sessions flow through the same snapshot fields, and for CLAUDE sessions context-attachment semantics are untouched — every new code path only adds a second agent's records as grounding inputs through the existing pipeline.

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh` (build + unit tests):

```
Executed 2040 tests, with 2 tests skipped and 0 failures (0 unexpected) in 71.840 (71.934) seconds
```

- [x] New focus truth-table coverage (`ClaudeSessionFocusResolutionTests`, 16 tests): single+focus resolved / single-no-focus abstain (no sole-session fallback) / multiple+fresh-focus resolved to declared / multiple+stale-focus abstain / multiple-no-focus abstain / focus-for-dead-session abstain / dead-declarer-pid abstain / pid-mismatch abstain / claim-vs-focus conflict abstain / claude single-claim unchanged / remote focus refused / bounded focus table / namespacing + cross-agent forgery drop / opencode lifecycle through the reducer.
- [x] Wire v1/v2 round-trips: v1 golden line still decodes as Claude; opencode golden line round-trips; unknown agent (`"aider"`, non-string) rejected precisely; `agent` key absent from Claude records (`ClaudeHookWireCodecTests`).
- [x] Regression shape called out: first test run caught the reducer absorbing focus-record process info into the session (3 truth-table failures — resolved-instead-of-abstain); fixed by excluding `.focusChanged` from process absorption, rerun green (`.build/last-remote.log` history in the PR discussion).

LLM lanes: these paths are in `scripts/ci/llm-lane-filter.sh` and the lane will run. Context-attachment semantics for Claude sessions are unchanged; opencode sessions only add grounding inputs through the existing pipeline (no prompt/model/request-shape change), so no eval scoreboard beyond the lane itself.

### DIFF
--- a/Sources/ClaudeContextWire/ClaudeHookWire.swift
+++ b/Sources/ClaudeContextWire/ClaudeHookWire.swift
@@ -12,7 +12,59 @@ import Foundation
 /// dropped, never partially trusted.
 public enum ClaudeHookWire {
     /// Current wire version. Bump on any incompatible field change.
-    public static let version = 1
+    ///
+    /// v2 added the `agent` field (absent = `.claude`) and the `FocusChanged`
+    /// event for the opencode integration.
+    public static let version = 2
+
+    /// Versions this build still reads. v1 is the pre-agent wire: every v1
+    /// record is, by definition, a Claude Code record, so v1 lines decode with
+    /// `agent == .claude` rather than being dropped — the publisher binary and
+    /// the app usually ship together, but a stale installed plugin pointing at
+    /// an old publisher must not silently lose all context after an update.
+    public static let readableVersions: Set<Int> = [1, 2]
+}
+
+/// Which coding agent published a record. Content, not trust: trust stays a
+/// property of the transport (`ClaudeTransportOrigin`), and every local agent
+/// publishes over the same peer-authenticated socket. What the agent tag
+/// decides is session-id namespacing (`ClaudeAgentSessionScope`) and
+/// per-agent channel rules — e.g. the broker never returns a title marker to
+/// an opencode session, because opencode owns its window title the way herdr
+/// owns its pane titles.
+///
+/// Decoding an unknown agent name is a drop, mirroring unknown events: a
+/// newer plugin talking to an older app degrades to "ignored", and an agent
+/// whose channel rules we do not know must not inherit another's.
+public enum ClaudeHookAgent: String, Sendable, Equatable, CaseIterable, Codable {
+    case claude
+    case opencode
+}
+
+/// Namespacing for per-agent session ids, mirroring `ClaudeRemoteSessionScope`.
+///
+/// Two agents pick their own session ids and cannot coordinate — Claude Code
+/// uses bare UUIDs, opencode uses `ses_…` — so a bare id is a claim, not a
+/// key. Scoping opencode ids under a prefix no Claude-published UUID can carry
+/// makes cross-agent collision structurally impossible. Applied by the
+/// RECEIVER (`ClaudeSessionRegistry.ingest`), never trusted from the wire, so
+/// a publisher cannot pre-scope itself into another namespace being honest or
+/// otherwise — the registry recomputes the key from the agent tag every time.
+/// (A local peer lying about its agent tag is a same-UID process and already
+/// outside the threat model, same as the remote scope's note on host ids.)
+public enum ClaudeAgentSessionScope {
+    /// Distinct from `ClaudeRemoteSessionScope.prefix` ("remote:") — the two
+    /// namespaces must never alias.
+    public static let opencodePrefix = "opencode:"
+
+    public static func scopedSessionID(agent: ClaudeHookAgent, sessionID: String) -> String {
+        switch agent {
+        case .claude:
+            return sessionID
+        case .opencode:
+            return opencodePrefix + sessionID
+        }
+    }
 }
 
 /// Hook events the plugin publishes. Cases map 1:1 to Claude Code hook names.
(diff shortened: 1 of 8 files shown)


T0mSIlver:
## Review response (wire/registry half) — commit 6a0b579

Addresses the codex CHANGES-REQUIRED items that live on this branch; the plugin/broker half is on #204 (`4664dfb`).

- **C2b (real ordering bug — thanks, this was the sharpest catch):** focus state was written *before* the record was validated, so a declaration naming an unknown/mismatched session suppressed a live Claude per-session TTY claim into `.stale` for the whole focus TTL even though ingest returned nil. Declarations are now fully validated first (opencode agent only, local transport, TTY present, target session known + same agent + local + registered to the declaring pid) and only then recorded. Regression: `testInvalidFocusRecordCannotSuppressAClaudeTTYClaim` (fails on the previous commit, passes now).
- **C2a:** took the reserved-prefix rejection rather than structural `(agent, rawID)` keying — the string key is load-bearing across snapshots, marker index, joins, and herdr claims, and the aliasing hole closes completely at the door: a LOCAL Claude record whose raw id spells `opencode:`/`remote:` is dropped whole (the remote listener's pre-scoped ids arrive over a remote origin and are exempt). Tests cover collide-with-existing, no-existing, and the `remote:` spelling.
- **C1 (registry half):** `FocusCleared` (wire v2) retracts a pane's declaration the moment the TUI leaves its session view; TTL tightened 120 s → 45 s (one lost 20 s heartbeat + jitter — with switches and retractions now event-shaped on the plugin side, TTL only covers lost writes). Clear refusals tested (`.claude` agent, remote transport).
- **C3:** fresh, pid-verified declarations now arbitrate multi-session herdr panes exactly like a TTY (every opencode session in one TUI shares pane id and pid, so the herdr arm went permanently `.ambiguous` at two sessions). Tests: two-session resolve-through-focus, retraction restores abstention, declaration expiry, single-session unchanged.
- **Hardening (coordinator ruling on codex finding 2):** `.claude` focus records (declare or clear) are refused as inconsistent. The kernel peer-pid binding is implemented on #204 (broker half).
- **F5:** a v1 line carrying an explicit `agent` key is `.malformed` — no v1 writer ever emitted it.
- **F3:** `registeredClaudeIsForeground` → `registeredAgentIsForeground`, abstention wording names the session's actual agent.

Truth-table additions: refused-at-ingest pid mismatch, resolve-time pid drift after a valid declaration, claude-focus refusal, C2b suppression regression, clear/retract + refusals, herdr ×3.

Proof (this branch):
```
./scripts/remote-build.sh test --filter Claude   Executed 731 tests, with 0 failures
./scripts/remote-build.sh                        Executed 2049 tests, with 2 tests skipped and 0 failures (73.556s)
```

T0mSIlver:
Regression-test evidence for C2b, both runs per proof culture (the fix landed in 6a0b579; the failing run is the same test injected into pre-fix 33e07ae):

**Before (33e07ae + the new test only):**
```
ClaudeSessionFocusResolutionTests.swift:359: error: testInvalidFocusRecordCannotSuppressAClaudeTTYClaim :
XCTAssertEqual failed: ("nil") is not equal to ("Optional(\"cl\")") - an invalid declaration must leave the per-session TTY claim untouched
Test Case ... failed (0.156 seconds).
```
(the ghost focus declaration turned the live Claude session's TTY resolution into an abstention)

**After (6a0b579):** same test passes in the suite run above (`--filter Claude`: 731 tests, 0 failures).

PR #204, opened 2026-07-27
opencode integration: dual-realm plugin, broker marker gating, contract tests

T0mSIlver:
## What

The opencode plugin itself, stacked on the wire/registry generalization (parent PR). One dependency-free JS file, `integrations/opencode/localvoxtral.js`, carrying BOTH plugin halves, plus README, broker marker gating, and contract tests.

**Why the file exports one half per JS realm.** opencode's loaders reject a default export that carries both `server()` and `tui()` (`packages/opencode/src/plugin/shared.ts`, `readV1Plugin` — verified against the v1.17.12 source), and the local TUI runs its server in a Bun **Worker thread** (`cli/cmd/tui.ts`, `new Worker(file)`): the server loader executes in the worker realm, the TUI loader in the main realm, same process, different module graphs. So a cross-realm in-process flag is impossible — and unnecessary: `isMainThread` IS the pane-ownership gate. The worker realm (never a TUI) exports the server half; the main realm exports the TUI half; headless main realms (`opencode serve`/`run`) find the wrong shape and load nothing (log-only), which is the safe outcome — a serve process must never publish a TTY.

- **Server half**: `session.created/deleted/idle` + `chat.message` + `tool.execute.after` → `SessionStart/SessionEnd/Stop/UserPromptSubmit/PostToolUse`, agent-tagged, cwd cached from `session.created` (`chat.message` carries none; missing cwd drops the FIELD, not the record), subagent (parentID) sessions filtered, herdr pane env forwarded so the existing herdr arm applies unchanged. Never a TTY.
- **TUI half**: reads its own TTY child-free (fd 0; on macOS the `/dev/ttysNNN` candidate's `rdev` must equal fd 0's or it abstains; `/proc/self/fd/0` on Linux) and declares the displayed session (`FocusChanged`) by sampling `api.route.current` on an `unref()`ed **500 ms** timer plus an immediate resample on every TuiEventBus event (opencode 1.17.x publishes no route-change event — checked the TUI event bus types and runtime source), with an explicit `FocusCleared` retraction on leaving the session view and on dispose, a 20 s heartbeat, and publish state that advances only on an accepted write — against the registry's 45 s focus TTL. (Post-review shape; see the review-response comment.)
- **Blast radius** (the plugin lives inside the user's agent process): zero `await`s, one lazily reconnected `unref()`ed AF_UNIX socket, fire-and-forget writes, replies read and discarded, every handler try/caught, every field bounded to the Swift wire limits before encoding, zero terminal output.
- **Broker**: marker emission gated per agent — an opencode session never receives a title marker (opencode rewrites its own OSC titles mid-turn; the herdr rule generalized). Allocation unchanged: the registry marker stays every join's liveness handle.
- **Install** is manual for now (Settings row is a follow-up): copy the file into `~/.config/opencode/plugins/` AND list it in `~/.config/opencode/tui.json`. Empirically the TUI loader has **no** plugin auto-discovery on 1.17.x — only the server loader scans `plugins/*.js` — so the research premise "no config edit needed" held for the server half only. The README explains the cost of skipping the tui.json step (vocabulary-only).

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh`:

```
Executed 2055 tests, with 2 tests skipped and 0 failures (0 unexpected) in 76.654 (76.746) seconds
```

  including the new `OpencodePluginManifestTests` (14 tests: wire-constant pins against `ClaudeHookWire`/`ClaudeHookLimits`/`ClaudeHookSocketPath`, forbidden-pattern gates incl. zero `await`s and zero terminal writes, and a JavaScriptCore evaluation — Bun's own engine — asserting the per-realm export shapes) and the broker gating test (`testBrokerWithholdsTheMarkerForAnOpencodeRecord`).

- [x] **Empirical plugin transcript** (strongest proof available off-Mac): on the Linux dev box, opencode 1.17.12, ISOLATED config (`XDG_CONFIG_HOME` scratch dir — the real `~/.config/opencode` untouched), `LOCALVOXTRAL_CLAUDE_SOCKET` pointed at a fake AF_UNIX broker that logs lines and replies `{"marker":null,"v":2}`. Drove the real TUI in a pty (typed a prompt, real GLM-5.2 turn, read tool ran). Everything the plugin published (cwd paths shortened to `…/oc-test/proj` for width; otherwise verbatim):

```
{"v":2,"event":"SessionStart","agent":"opencode","session_id":"ses_05a3d3459ffeFtEYWjY8GxWRYU","ts":1785192434.603,"process":{"hook_pid":2621112,"claude_pid":2621112,"herdr_pane_id":"w3:p2","herdr_socket_path":"/home/dev/.config/herdr/herdr.sock"},"cwd":"…/oc-test/proj"}
{"v":2,"event":"UserPromptSubmit","agent":"opencode","session_id":"ses_05a3d3459ffeFtEYWjY8GxWRYU","ts":1785192434.617,"process":{…same…},"cwd":"…/oc-test/proj","prompt":"Read notes.txt and tell me the demo word."}
{"v":2,"event":"FocusChanged","agent":"opencode","session_id":"ses_05a3d3459ffeFtEYWjY8GxWRYU","ts":1785192435.38,"process":{"hook_pid":2621112,"claude_pid":2621112,"tty":"/dev/pts/4","herdr_pane_id":"w3:p2","herdr_socket_path":"/home/dev/.config/herdr/herdr.sock"}}
{"v":2,"event":"PostToolUse","agent":"opencode","session_id":"ses_05a3d3459ffeFtEYWjY8GxWRYU","ts":1785192437.799,"process":{…same, no tty…},"cwd":"…/oc-test/proj","tool_name":"read","files":[{"path":"…/oc-test/proj/notes.txt","kind":"read"}]}
{"v":2,"event":"Stop","agent":"opencode","session_id":"ses_05a3d3459ffeFtEYWjY8GxWRYU","ts":1785192439.586,"process":{…same, no tty…},"cwd":"…/oc-test/proj"}
{"v":2,"event":"FocusChanged", …ts":1785192455.395, "tty":"/dev/pts/4" …}   <- 20 s heartbeat
{"v":2,"event":"FocusChanged", …ts":1785192475.416, "tty":"/dev/pts/4" …}   <- 20 s heartbeat
```

  Note the invariants visible in the bytes: **only** `FocusChanged` carries `tty`; session records never do; `hook_pid == claude_pid ==` the one opencode process in both realms (so the registry's pid cross-check binds focus to sessions); herdr pane identity forwarded (this run really was inside a herdr pane, `w3:p2`).

- [x] **Negative proof — headless publishes nothing**: same isolated config, `opencode run "Say the word hello and nothing else."` completed a real model turn (`hello`) and the fake broker received **zero** records (no log file created). Exactly the design: no pane, no TTY, no records.

- [x] An instrumented debug run (scratch copy only) confirmed both halves load in ONE process, two realms: `module eval isMainThread=false … worker.js` → ServerHalf; `module eval isMainThread=true … index.js` (same pid) → TuiHalf, `ownTTY=/dev/pts/4`, route polls `{"name":"home"}` → `{"name":"session","params":{"sessionID":…}}` on submit.

LLM lanes: paths under `Sources/localvoxtral/ClaudeContext` are lane-relevant and the lane will run. Context-attachment semantics for CLAUDE sessions are unchanged; opencode sessions only add grounding inputs through the existing pipeline — no prompt, model, or request-shape change.

## Deferred

- Settings row / bundled install service (parallel Settings redesign in flight).
- Remote/SSH phase for opencode.
- `opencode serve`+`attach` session content (attach panes get focus declarations, but serve publishes nothing — documented in the README).

### DIFF
--- a/Sources/ClaudeContextWire/ClaudeHookWire.swift
+++ b/Sources/ClaudeContextWire/ClaudeHookWire.swift
@@ -12,7 +12,59 @@ import Foundation
 /// dropped, never partially trusted.
 public enum ClaudeHookWire {
     /// Current wire version. Bump on any incompatible field change.
-    public static let version = 1
+    ///
+    /// v2 added the `agent` field (absent = `.claude`) and the `FocusChanged`
+    /// event for the opencode integration.
+    public static let version = 2
+
+    /// Versions this build still reads. v1 is the pre-agent wire: every v1
+    /// record is, by definition, a Claude Code record, so v1 lines decode with
+    /// `agent == .claude` rather than being dropped — the publisher binary and
+    /// the app usually ship together, but a stale installed plugin pointing at
+    /// an old publisher must not silently lose all context after an update.
+    public static let readableVersions: Set<Int> = [1, 2]
+}
+
+/// Which coding agent published a record. Content, not trust: trust stays a
+/// property of the transport (`ClaudeTransportOrigin`), and every local agent
+/// publishes over the same peer-authenticated socket. What the agent tag
+/// decides is session-id namespacing (`ClaudeAgentSessionScope`) and
+/// per-agent channel rules — e.g. the broker never returns a title marker to
+/// an opencode session, because opencode owns its window title the way herdr
+/// owns its pane titles.
+///
+/// Decoding an unknown agent name is a drop, mirroring unknown events: a
+/// newer plugin talking to an older app degrades to "ignored", and an agent
+/// whose channel rules we do not know must not inherit another's.
+public enum ClaudeHookAgent: String, Sendable, Equatable, CaseIterable, Codable {
+    case claude
+    case opencode
+}
+
+/// Namespacing for per-agent session ids, mirroring `ClaudeRemoteSessionScope`.
+///
+/// Two agents pick their own session ids and cannot coordinate — Claude Code
+/// uses bare UUIDs, opencode uses `ses_…` — so a bare id is a claim, not a
+/// key. Scoping opencode ids under a prefix no Claude-published UUID can carry
+/// makes cross-agent collision structurally impossible. Applied by the
+/// RECEIVER (`ClaudeSessionRegistry.ingest`), never trusted from the wire, so
+/// a publisher cannot pre-scope itself into another namespace being honest or
+/// otherwise — the registry recomputes the key from the agent tag every time.
+/// (A local peer lying about its agent tag is a same-UID process and already
+/// outside the threat model, same as the remote scope's note on host ids.)
+public enum ClaudeAgentSessionScope {
+    /// Distinct from `ClaudeRemoteSessionScope.prefix` ("remote:") — the two
+    /// namespaces must never alias.
+    public static let opencodePrefix = "opencode:"
+
+    public static func scopedSessionID(agent: ClaudeHookAgent, sessionID: String) -> String {
+        switch agent {
+        case .claude:
+            return sessionID
+        case .opencode:
+            return opencodePrefix + sessionID
+        }
+    }
 }
 
 /// Hook events the plugin publishes. Cases map 1:1 to Claude Code hook names.
@@ -36,6 +88,13 @@ public enum ClaudeHookEvent: String, Sendable, Equatable, CaseIterable, Codable 
     case fileChanged = "FileChanged"
     case stop = "Stop"
     case sessionEnd = "SessionEnd"
+    /// Published by the opencode plugin's TUI half only (v2): "the pane on
+    /// this TTY currently displays this session". One opencode process hosts
+    /// several sessions on one terminal, so per-session TTY claims cannot
+    /// disambiguate them — an explicit, freshness-bounded focus declaration
+    /// can. Claude Code never sends this; nothing in its hook set knows what a
+    /// pane displays.
+    case focusChanged = "FocusChanged"
 }
 
 /// Hard bounds applied at BOTH ends of the wire.
@@ -155,6 +214,10 @@ public struct ClaudeFileTouch: Sendable, Equatable, Codable {
 public struct ClaudeHookRecord: Sendable, Equatable {
     public var version: Int
     public var event: ClaudeHookEvent
+    /// Which coding agent published this. Encoded only when it is not the
+    /// default `.claude`, so v1 readers' "absent field" and v2's default agree
+    /// byte-for-byte on every Claude record.
+    public var agent: ClaudeHookAgent
     public var sessionID: String
     /// Seconds since the UNIX epoch, as stamped by the publisher.
     public var timestamp: Double
@@ -170,6 +233,7 @@ public struct ClaudeHookRecord: Sendable, Equatable {
     public init(
         version: Int = ClaudeHookWire.version,
         event: ClaudeHookEvent,
+        agent: ClaudeHookAgent = .claude,
         sessionID: String,
         timestamp: Double,
         rawCwd: String? = nil,
@@ -180,6 +244,7 @@ public struct ClaudeHookRecord: Sendable, Equatable {
     ) {
         self.version = version
         self.event = event
+        self.agent = agent
         self.sessionID = sessionID
         self.timestamp = timestamp
         self.rawCwd = rawCwd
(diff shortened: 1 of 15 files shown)


T0mSIlver:
## Review response (plugin/broker half) — commit 4664dfb (parent fixes merged as 43335e3)

- **C1 (focus immediacy):** one factual correction to the review premise first: **no route-change event exists** in opencode 1.17.x's TUI plugin API — `api.route.current` is a plain getter, `TuiEventBus` carries server events only (verified against `@opencode-ai/plugin/tui` types and `packages/opencode/src/plugin/tui/runtime.ts`; the poll was never a choice against an available event). What the fix does instead, from both ends: sample cadence 2 s → **500 ms** (switch latency now ≤ 0.5 s, an in-memory getter read), plus an **immediate resample on every TuiEventBus event**, so any switch coinciding with session activity is caught event-fast — live revalidation shows `FocusChanged` landing **69 ms** after `UserPromptSubmit` (was ~0.5–2 s). Leaving the session view publishes an explicit **`FocusCleared`** retraction (also on dispose, flushed via `end()`), so the home route no longer retains the old session for the TTL. And `lastSentAt`/`lastSession` advance **only on an accepted write**, so a lost declaration or retraction retries at the next 500 ms sample instead of being suppressed for a heartbeat. Registry-side TTL is 45 s (parent branch).
- **C5 (backpressure):** every write is gated by `socket.writableLength > MAX_QUEUED_BYTES` (64 KB): past the watermark the connection is reset and the record dropped — never a wait, never unbounded queueing inside the agent process. Contract-pinned (`testBackpressureIsBoundedByADropNeverAWait`).
- **C4 (mixed-version compat):** the broker now echoes the **request's** wire version in its reply. Compat matrix (bundled publisher stamps `ClaudeHookWire.version`, i.e. v2 in this build):
  | publisher | broker | record | reply | outcome |
  |---|---|---|---|---|
  | v1 (stale install) | new | v1, accepted (`readableVersions`) | **v1-shaped** | context + marker keep working — E2E-tested |
  | v2 | new | v2 | v2-shaped | normal path |
  | v2 | old (v1 app) | rejected (`unsupportedVersion`) | v1-shaped, no marker | fail-open, no context until app updates (ships in the same bundle, so transient) |

  End-to-end test: `testBrokerEchoesAV1RequestsVersionSoAStalePublisherKeepsItsMarker` (raw v1 line in → `"v":1` marker reply out; v2 in → v2 out).
- **Peer-pid binding (coordinator ruling on codex finding 2):** implemented — the plumbing was clean. `ClaudeSocketGuard.peerPID` reads `LOCAL_PEERPID` (kernel-recorded at connect), and the broker rejects any **opencode** record whose claimed pid differs from the socket peer's: the plugin runs *inside* the agent process and dials from it, so the kernel's answer must match. The Claude hook path structurally cannot be held to this (transient child publisher ≠ session process — that is why `claude_pid` rides in the record at all); the asymmetry is documented at both the guard and the broker check, and the residual same-user surface for that path remains the accepted, documented trust model. Test: `testOpencodeRecordClaimingAForeignPidIsRejectedByThePeerPidCheck`.
- **C6 (+opencode F2):** contract tests now forbid `writeSync` (raw fd-1/2 writes bypassed the blocklist) and mechanically pin that the **server half's entire span never contains the token `tty`** — `record()`'s tty parameter is lexically unreachable from it. New pins: backpressure watermark, 500 ms sample, `FocusCleared`, heartbeat-vs-TTL relation.
- **F6:** plugin local renamed `scopedID` → `boundedID` (namespacing is receiver-side by design; the plugin only bounds).
- **F7:** the plugin-load-directory cwd fallback is **gone** — an instance-less server hosts sessions from many directories and a wrong cwd grounds dictation on the wrong repo. Unknown directory ⇒ record without `cwd`.

Revalidation (real TUI, isolated config, fake broker — same rig as the original Proof):
```
UserPromptSubmit  ts 1785195198.675
FocusChanged      ts 1785195198.744   (+69 ms, event-triggered resample; tty only here)
PostToolUse / Stop …
FocusChanged      +20 s, +40 s        (heartbeats)
FocusCleared      ts 1785195239.505   (teardown retraction, flushed)
```
`opencode run` still publishes zero records.

Proof (this branch, after the merge from #203):
```
./scripts/remote-build.sh test --filter Claude   Executed 736 tests, with 0 failures
./scripts/remote-build.sh                        Executed 2068 tests, with 2 tests skipped and 0 failures (73.779s)
```

PR #206, opened 2026-07-28
Runbook: repin speechd test service to the qhead catalog default

T0mSIlver:
## What

The owner runbook (`scripts/mac/README.md`) still pinned speechd's test service to the retired `mlx-community/Voxtral-Mini-4B-Realtime-2602-4bit @ fdebf7b2` in both the `hf download` step and the embedded launchd plist — but #169 moved `SpeechModelCatalog.defaultOption` to `T0mSIlver/Voxtral-Mini-4B-Realtime-2602-4bit-qhead @ 247f2eec`. Anyone installing from the runbook verbatim (as the runbook's own "keep these pins in sync" rule warns) would silently reintroduce the old model, and tier-1 would measure a backend production no longer ships. Flagged during the 2026-07-27/28 test-services migration (the live service was installed with the correct qhead pin; only the doc drifted).

Both pins updated, and — since the prose sync rule drifted within days of being written — `MacTestServerRunbookPinTests` now pins the runbook to the catalogs: every plist `--model`/`--model-revision` in the doc must equal a current `SpeechModelCatalog`/`PolishModelCatalog` default (set equality, so a stale pin AND a missing service section both fail), and the `hf download` commands must name the same repo+revision.

## Proof

- [x] Regression test added — failing before the fix (stale runbook, new tests):
  ```
  MacTestServerRunbookPinTests.swift:43: XCTAssertEqual failed: ("["mlx-community/Voxtral-Mini-4B-Realtime-2602-4bit", …]") is not equal to ("["T0mSIlver/Voxtral-Mini-4B-Realtime-2602-4bit-qhead", …]")
       Executed 2 tests, with 4 failures (0 unexpected) in 0.200 (0.200) seconds
  ```
  Passing after:
  ```
  Test Suite 'MacTestServerRunbookPinTests' passed at 2026-07-28 11:29:08.081.
  ```
- [x] Tier-1 against the live migrated testspeechd (new gate, Swift helper, qhead model), run from the Linux box via the gate's `ensure`:
  ```
  Executed 7 tests, with 2 tests skipped and 0 failures (0 unexpected) in 19.090 (19.091) seconds
  speechd integration: word accuracy 0.759
  ```
- [x] LLM lanes: not required — docs + a doc-pinning test; nothing that reaches a model changed.

### DIFF
--- a/Tests/localvoxtralTests/MacTestServerRunbookPinTests.swift
+++ b/Tests/localvoxtralTests/MacTestServerRunbookPinTests.swift
@@ -1,0 +1,74 @@
+import Foundation
+import XCTest
+@testable import localvoxtral
+
+/// `scripts/mac/README.md` is the owner runbook for the on-demand test
+/// services, and it embeds complete launchd plists plus `hf download` commands
+/// with model pins. Its own rule — "keep these pins in sync with
+/// `SpeechModelCatalog.defaultOption` and `PolishModelCatalog.defaultOption`" —
+/// drifted almost immediately: #186 moved the speech default to the qhead
+/// repo while the runbook kept the retired mlx-community pin, so anyone
+/// installing from the runbook verbatim would silently reintroduce the old
+/// model and tier-1 would measure a backend production no longer ships.
+/// Prose sync rules drift; these assertions do not.
+final class MacTestServerRunbookPinTests: XCTestCase {
+    private func runbook() throws -> String {
+        let root = URL(fileURLWithPath: #filePath)
+            .deletingLastPathComponent() // localvoxtralTests
+            .deletingLastPathComponent() // Tests
+            .deletingLastPathComponent() // repo root
+        return try String(
+            contentsOf: root.appendingPathComponent("scripts/mac/README.md"),
+            encoding: .utf8
+        )
+    }
+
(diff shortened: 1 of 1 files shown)


T0mSIlver:
## codex (GPT-5.6 Sol) review — findings fixed in `b5fb4f9`

1. **High — set-based pins accept swapped revisions.** Fixed: pins are now per-service `(repo, revision)` pairs — each plist assertion is scoped to that binary's `ProgramArguments` window, and the `hf download` mapping must equal the catalog defaults exactly. Mutation proof: with the two revisions swapped in the runbook (which the old test passed), the rewrite fails 3 assertions; restored runbook passes `Executed 2 tests, with 0 failures`.
2. **Low — one-line adjacency brittleness.** Fixed: value may land on the following line (`\s*` join), with exactly-one-occurrence asserted per flag per window.
3. **Low — wrong history.** Fixed in the doc comment and PR body: **#169** moved the catalog default to qhead (#186 only returned the mlx-audio pin upstream); #176 installed the runbook pin that went stale.

Unverifiable-from-tree items (tier-1 0.759 run, live-host qhead state) are runtime facts recorded in the PR body's Proof section, not tree content.

PR #211, opened 2026-08-03
Remote shim: back off after transport failures so ssh stops spraying connect_to noise over the terminal

T0mSIlver:
## What

When the `localvoxtral-remote` plugin runs in Claude Code over SSH while localvoxtral is **not** running on the Mac, the terminal fills with connection-failure noise — especially painful inside herdr panes.

The shim itself was already fully silent; the noise is **ssh on the Mac**: with the enrollment's `RemoteForward`, every hook curl makes sshd open a channel back to the Mac, the local ssh client fails to connect, and OpenSSH prints `connect_to 127.0.0.1 port 8473: failed.` (exact format string verified in OpenSSH channels.c) straight onto the PTY — once per hook, dozens of times per turn via `PostToolUse`. That stderr is another process on another machine: no plugin-side redirect can reach it, and `LogLevel QUIET` in the ssh config would also swallow host-key-changed warnings, which is not a trade the app should auto-insert.

So the shim stops dialing a tunnel it just proved dead:

- A transport-level curl failure arms a 5-minute epoch stamp (`$XDG_RUNTIME_DIR/localvoxtral/hook-backoff`, else `~/.cache/...`); while fresh, every event **except `UserPromptSubmit`** skips curl entirely — no dial, no ssh line.
- `UserPromptSubmit` always dials, deliberately: it is user-paced (one line per submitted prompt while the app is down — an honest, bounded signal), it carries the join marker and the prompt, so the **first prompt after the app comes back is grounded immediately** — and its completed exchange clears the backoff for everything else.
- Any completed HTTP exchange (even a 401) clears the stamp. Every odd stamp state — no usable dir, no `date`, garbage content, clock jumped backwards, oversized value — fails toward dialing, i.e. exactly the pre-backoff behavior. Fail-open is untouched.

Also: plugin `1.1.0 → 1.2.0` + marketplace metadata `1.2.0 → 1.3.0` (enrolled hosts pick this up via the existing `marketplace update` + `plugin update` pair); README section naming the exact ssh message, AGENTS.md, and a new enrollment-plan note (pinned by tests) so the symptom stops reading as a plugin bug.

`integrations/claude-code/*` is in `scripts/ci/llm-lane-filter.sh`, so the polishd lane runs on this PR by path match. Nothing here changes what reaches the model — the change is delivery-transport backoff plus docs.

## Proof

- [x] Full unit suite green — `./scripts/remote-build.sh` (build + unit):
  ```
  Executed 2164 tests, with 2 tests skipped and 0 failures (0 unexpected) in 73.778 (73.883) seconds
  ```
- [x] Regression tests added (4, they RUN the shim against the stub curl) — failing before the fix, passing after.

  **Red** (shim reverted to main, new tests in place, on the Mac):
  ```
  error: testTransportFailureArmsTheBackoffAndLaterEventsSkipTheDialEntirely : failed: caught error: "... hook-backoff couldn't be opened because there is no such file."
  error: testUserPromptSubmitDialsThroughAnArmedBackoffAndSuccessClearsIt : XCTAssertFalse failed - a completed exchange must clear the stamp
  error: testEvenA401ClearsTheBackoffBecauseTheExchangeCompleted : XCTAssertFalse failed - any completed HTTP exchange must clear the stamp
  Executed 34 tests, with 3 failures (1 unexpected)
  ```
  (`testExpiredCorruptAndFutureStampsAllFailTowardDialing` passes on main by design — it pins that odd stamp states behave exactly like the pre-backoff shim.)

  **Green** (this branch, on the Mac):
  ```
  Test Suite 'ClaudeRemotePluginManifestTests' passed
  Executed 34 tests, with 0 failures (0 unexpected) in 7.659 (7.666) seconds
  ```
  Plus `ClaudeRemoteEnrollmentServiceTests`: `Executed 51 tests, with 0 failures`.
- [x] POSIX-sh portability: the same scenarios replicated in a dash harness on Linux (macOS runs bash-as-sh; remote hosts are typically dash) — all 10 checks pass on the new shim; same harness on main's shim fails the 4 backoff checks and passes the 6 invariant ones.
- [x] Exact ssh message verified against OpenSSH: `strings /usr/bin/ssh | grep connect_to` → `connect_to %.100s port %d: failed.` — quoted verbatim in README/comments/notes so users can search for it.

Not run: `remote-build.sh integration` (no realtime/ASR path touched). The polishd lane runs in CI via the path filter.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -394,7 +394,13 @@ Key subsystems:
     hooks running the bundled POSIX-sh shim `hooks/post.sh`, which curls the
     event JSON to `127.0.0.1:8473/v1/hook/<Event>` through an OpenSSH
     `RemoteForward` — no localvoxtral binary and no `jq`/`nc`/Node on that
-    host, but it does need `sh` and `curl` (fail-open when absent). It was
+    host, but it does need `sh` and `curl` (fail-open when absent). After a
+    transport-level failure the shim backs off for 5 minutes (epoch stamp
+    under `$XDG_RUNTIME_DIR`/`~/.cache`) for every event except
+    `UserPromptSubmit`: each dial against a live forward with no app behind
+    it makes the Mac-side ssh client print `connect_to …: failed.` onto the
+    user's terminal — stderr the remote side can never redirect — and any
+    completed HTTP exchange (even a 401) clears the backoff. It was
     `type: "http"` hooks until 2026-07-27: Claude Code expands http-hook
     header `${VAR}`s from the process environment only and never injects
     plugin userConfig options there (verified on 2.1.220), so every hook
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteEnrollmentService.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteEnrollmentService.swift
@@ -360,6 +360,11 @@ public struct ClaudeRemoteEnrollmentService: Sendable {
             "The plugin needs only POSIX `sh` and `curl` on the remote host — no localvoxtral binary. "
                 + "Its hooks fail open silently when either is missing, the tunnel is down, or the app "
                 + "is not answering: you simply get no context, never a blocked Claude turn.",
+            "While a session holds the tunnel and localvoxtral is not running, ssh itself — on this "
+                + "Mac — prints `connect_to 127.0.0.1 port \(port): failed.` into the remote terminal on "
+                + "every dial; the plugin cannot silence another process. So after a failed dial its "
+                + "hooks back off for five minutes; prompt submits still try, so context returns with "
+                + "your first prompt once the app is back.",
             "A second concurrent SSH session to the same host will fail to bind \(port) on the "
                 + "remote and — because ExitOnForwardFailure is `no` — will connect anyway with no "
                 + "tunnel. The first session keeps the forward.",
--- a/Tests/localvoxtralTests/ClaudeRemoteEnrollmentServiceTests.swift
+++ b/Tests/localvoxtralTests/ClaudeRemoteEnrollmentServiceTests.swift
@@ -465,6 +465,12 @@ final class ClaudeRemoteEnrollmentServiceTests: XCTestCase {
             notes.contains("curl") && notes.contains("fail open"),
             "the host dependency (sh + curl) and its fail-open behavior must be stated honestly"
         )
+        XCTAssertTrue(
+            notes.contains("connect_to") && notes.contains("back off"),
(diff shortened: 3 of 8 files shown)


T0mSIlver:
## opencode (GLM-5.2) review — verdict: MERGEABLE-AS-IS, both findings fixed anyway

The reviewer ran the shim under dash against a stub curl, exercising adversarial stamp states (corrupt / future / expired / oversized / partial-write / stamp-is-a-directory), 20-way concurrent arms/reads, and the stdout gate. Core logic confirmed sound; no bashisms; no new path to stdout. Two non-blocking hardening findings, both fixed in 8777bd1:

1. **Minor — symlink-following stamp write** (`post.sh`): `echo "$NOW" >"$STAMP"` wrote through a pre-planted symlink at the stamp path (reviewer reproduced a victim-file clobber). Same-user scope only, but inconsistent with the shim's own hygiene bar. Fixed: the stamp lands via a private tempfile + `mv` — rename(2) replaces the symlink instead of following it, and as a bonus a concurrent reader can no longer observe a half-written value.
2. **Nit — loose pre-existing stamp dir**: `mkdir -p` leaves an existing dir's mode alone, so a 0777 `localvoxtral/` under `XDG_RUNTIME_DIR`/`~/.cache` stayed loose. Fixed: arming `chmod 700`s the dir; any failure skips the arm, i.e. fails toward dialing.

**Proof** (regression tests run the shim against the stub curl):

Red — shim reverted to 1848d62, new tests in place, on the Mac:
```
error: testArmingReplacesAPrePlantedSymlinkInsteadOfWritingThroughIt : XCTAssertEqual failed ... "precious user data" — arming the backoff must never write through a symlink at the stamp path
error: testArmingReplacesAPrePlantedSymlinkInsteadOfWritingThroughIt : ... NSFileTypeSymbolicLink is not equal to NSFileTypeRegular — the symlink must have been replaced by a regular stamp file
error: testArmingTightensAPreExistingLooseStampDirectory : XCTAssertEqual failed: ("Optional(511)") is not equal to ("Optional(448)") — arming must tighten a pre-existing loose stamp directory to 0700
Executed 36 tests, with 3 failures (0 unexpected)
```

Green — this branch, on the Mac:
```
Test Suite 'ClaudeRemotePluginManifestTests' passed
Executed 36 tests, with 0 failures (0 unexpected) in 8.098 (8.102) seconds
```

POSIX portability re-verified in a dash harness on Linux: 8/8 checks pass on the fixed shim (symlink victim untouched, stamp replaces the link as a regular file, no stray tempfile, dir tightened to 700, arm/skip/UserPromptSubmit-dials-through all still hold); the pre-fix shim fails the 3 new checks.

Issue #215, opened 2026-08-03
Remote hook delivery: duplicate RemoteForward 8473 silently routes a second Mac's events (and token) to the first Mac


T0mSIlver:
## Summary

The remote hook delivery channel has a silent cross-Mac routing failure: when two ssh connections to the same enrolled host both request `RemoteForward 8473`, **the first connection's forward wins and keeps winning**. The second connection stays fully connected (our generated ssh-config block deliberately sets `ExitOnForwardFailure no`) and prints only a stderr warning — and from that point every hook event published on that host is delivered to the *first* Mac's listener.

Proven empirically on OpenSSH 10.0p2 (2026-08-03, as part of the remote-context research campaign):

- Second `ssh -R 9999:...` gets `Warning: remote port forwarding failed for listen port 9999` on stderr and continues without the forward.
- Traffic on the remote port routes to the **first** connection's backend, every time (probed repeatedly).
- With `ExitOnForwardFailure yes` the second *connection* dies instead — worse UX (the victim can't even open a second terminal), and the failure still lands on the victim, not the squatter.

## Impact

Concretely, with two Macs (say a desktop and a laptop, both running localvoxtral) enrolled against the same remote host:

1. **Silent blackhole**: the second Mac's dictation sessions never receive context from that host — hook events flow to the other Mac. Nothing surfaces this on either side.
2. **Token exposure**: the shim's requests — including the `Authorization: Bearer` header for Mac B's enrollment — are delivered to Mac A's listener. Mac A's listener 401s them (different token), which the shim treats as a completed HTTP exchange (it clears backoff, by design), so there is no distinguishable failure signal at all. This is the same trust class as the documented local port-squatter scenario, but reachable by accident between a user's own machines.

The single-Mac case can also hit this transiently: after a network blip, a half-dead ssh connection can hold the remote bind while a fresh connection silently loses it.

## Agreed fix shape (owner decision 2026-08-03)

**Detection + per-Mac port**:

- Add an optional `port` to the `localvoxtral-remote` plugin `userConfig` (default 8473). The shim posts to `127.0.0.1:${CLAUDE_PLUGIN_OPTION_PORT:-8473}` after validation.
- Enrollment allocates a stable per-Mac remote listen port (derived from the Mac's identity) and generates `RemoteForward <port> 127.0.0.1:8473` plus the matching plugin `--config`, so two enrolled Macs can never contend for the same remote bind. The Mac-side listener stays on 8473.
[...]

PR #226, opened 2026-08-05
Reap orphaned remote-forward ssh children at launch (fixes stale "Port held")

T0mSIlver:
## What

Field report (2026-08-05): quitting and reopening the app sometimes lands the SSH settings pane on **"Port held — close ssh sessions to that host."** with a Retry that can only fail. The holder was this Mac's own orphan: an app run that ends without `applicationWillTerminate` (crash, force-quit) — or whose SIGTERM→SIGKILL teardown outruns the bounded 3 s quit drain — leaves the app-held `ssh -N -R` reparented to launchd, alive indefinitely (BatchMode + ServerAlive keepalives), still holding the remote port bind. The next launch's fresh forward is refused its own port, and the supervisor's bind refusal is terminal by design.

The fix, following the `LegacyVoxmlxPortDefense` precedent:

- **`ClaudeRemoteForwardPidLedger`** — every spawned forward ssh is recorded (pid + kernel start time + resolved executable path) in a 0600 file beside the host registry, reusing `ClaudeRemoteHostFileStoreIO`'s hardened atomic writes. Written before the supervisor takes the process, so it survives any death the app doesn't see coming. A corrupt/unreadable ledger reads as EMPTY (loudly): it is a cleanup aid, and the safe reading of "cannot tell what we spawned" is "kill nothing".
- **`ClaudeRemoteForwardOrphanReaper`** — at launch, verifies each record against the pid's *current* kernel identity and kills matches with SIGTERM → bounded wait → SIGKILL on the injected clock. All three identity fields must match, so a reused pid can never get an innocent process signalled — a mismatch just retires the record. A SIGKILL survivor keeps its record so the next launch retries.
- **`ClaudeRemoteForwardCoordinator`** — gates the first forward start behind the reap. The gate sits *after* the listener-bound check on purpose: a second app instance cannot bind 8473 while the first lives, so it can never reap the first instance's healthy tunnels. The reap runs exactly once per launch; records written afterwards belong to this instance's own live supervisors.

Not fixed (stated honestly): "Port held" can still appear legitimately when the user's own interactive ssh session holds the enrollment `RemoteForward`, or when the server side holds a half-open session after a network drop (only sshd's own keepalive clears that). This PR removes the case where the holder was the app's own orphan — the one where Retry was hopeless.

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh` (full suite):
  ```
  Executed 2602 tests, with 2 tests skipped and 0 failures (0 unexpected) in 96.649 (96.783) seconds
  ```
- [x] Bug fix: regression test added — failing before the fix, passing after.

  Red (reap gate neutered to pre-fix behavior — forwards start immediately):
  ```
  ClaudeRemoteForwardCoordinatorTests.swift:282: error: -[localvoxtralTests.ClaudeRemoteForwardCoordinatorTests testForwardsWaitForTheOrphanReapBeforeStarting] : XCTAssertTrue failed - an orphan may still hold the port until the reap finishes
  Test Case '-[localvoxtralTests.ClaudeRemoteForwardCoordinatorTests testForwardsWaitForTheOrphanReapBeforeStarting]' failed (0.185 seconds).
  ClaudeRemoteForwardCoordinatorTests.swift:313: error: -[localvoxtralTests.ClaudeRemoteForwardCoordinatorTests testTheOrphanReapRunsExactlyOnce] : XCTAssertEqual failed: ("0") is not equal to ("1")
  ```
  Green (fix restored, `./scripts/remote-build.sh test --filter ClaudeRemoteForward`):
  ```
  Test Suite 'ClaudeRemoteForwardCoordinatorTests' passed at 2026-08-05 23:07:39.628.
  Test Suite 'ClaudeRemoteForwardOrphanReaperTests' passed at 2026-08-05 23:07:39.630.
  Test Suite 'ClaudeRemoteForwardPidLedgerTests' passed at 2026-08-05 23:07:39.631.
  	 Executed 60 tests, with 0 failures (0 unexpected) in 0.045 (0.049) seconds
  ```
  New coverage: the pid-reuse safety property (`testAReusedPidIsNeverSignalled`), the SIGTERM→SIGKILL escalation ladder, SIGKILL-survivor record retention, ledger round-trip/pid-scoped-forget/corrupt-reads-as-empty, and the coordinator gate (waits for reap, runs once, never reaps while the listener is unbound).
- LLM lanes skipped: backend-supervision/lifecycle paths only — nothing on this diff changes what reaches any model.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardCoordinator.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardCoordinator.swift
@@ -49,21 +49,49 @@ public final class ClaudeRemoteForwardCoordinator {
     @ObservationIgnored private let makeSupervisor: MakeSupervisor
     @ObservationIgnored private var supervisors: [String: any ClaudeRemoteForwarding] = [:]
 
+    /// Whether the launch-time orphan reap has run yet. Forwards may not start
+    /// before it finishes: the orphan holds exactly the remote port the new
+    /// forward is about to dial, so starting early reproduces the "Port held"
+    /// state the reap exists to clear.
+    private enum OrphanReapPhase { case pending, running, done }
+    @ObservationIgnored private var orphanReap: OrphanReapPhase
+    @ObservationIgnored private let reapOrphans: (@Sendable () async -> Void)?
+
     public init(
         hosts: ClaudeRemoteHostRegistry,
         remoteForwardPort: UInt16,
         listenerPort: UInt16 = ClaudeRemoteListenerLimits.default.port,
         isListenerBound: @escaping @MainActor () -> Bool,
-        makeSupervisor: MakeSupervisor? = nil
+        makeSupervisor: MakeSupervisor? = nil,
+        pidLedger: ClaudeRemoteForwardPidLedger? = nil,
+        reapOrphans: (@Sendable () async -> Void)? = nil
     ) {
         self.hosts = hosts
         self.remoteForwardPort = remoteForwardPort
         self.listenerPort = listenerPort
         self.isListenerBound = isListenerBound
+        self.reapOrphans = reapOrphans
+        // No reaper means nothing to wait for — the seam's absence must not
+        // stall every forward forever.
+        self.orphanReap = reapOrphans == nil ? .done : .pending
         self.makeSupervisor = makeSupervisor ?? { configuration in
             ClaudeRemoteForwardSupervisor(
                 configuration: configuration,
-                launch: { try ClaudeRemoteForwardLiveProcess(argv: $0.argv) }
+                launch: { config in
+                    let process = try ClaudeRemoteForwardLiveProcess(argv: config.argv)
+                    // Written BEFORE the process is handed to the supervisor:
+                    // if this app dies without applicationWillTerminate, the
+                    // record is what lets the next launch find and kill the
+                    // orphan instead of reporting "Port held" at it.
+                    if let pidLedger,
+                       let record = ClaudeRemoteForwardProcessIdentity.snapshot(
+                           pid: process.processIdentifier
+                       )
+                    {
+                        pidLedger.remember(hostID: config.hostID, record: record)
+                    }
+                    return process
+                }
             )
         }
     }
@@ -94,6 +122,33 @@ public final class ClaudeRemoteForwardCoordinator {
                 stopAll()
             }
             return
+        }
+
+        // Orphans from a previous run die BEFORE the first forward dials. The
+        // gate sits after the listener check on purpose: only the instance
+        // holding the listener port gets here, so a second app instance can
+        // never reap the first one's healthy tunnels.
+        switch orphanReap {
+        case .running:
+            return
+        case .pending:
+            guard let reapOrphans else {
+                orphanReap = .done
+                break
+            }
+            orphanReap = .running
+            Log.claudeContext.info(
+                "Claude remote forwards waiting for the orphan reap before first start"
+            )
+            Task { @MainActor [weak self] in
+                await reapOrphans()
+                guard let self else { return }
+                self.orphanReap = .done
+                self.reconcile()
+            }
+            return
+        case .done:
+            break
         }
 
         let eligible = eligibleHosts()
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardLiveProcess.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardLiveProcess.swift
@@ -35,6 +35,10 @@ final class ClaudeRemoteForwardLiveProcess: ClaudeRemoteForwardProcess, @uncheck
     private let partialLine = Mutex<String>("")
     private let descriptor: Int32
 
+    /// The spawned child's pid, for the pid ledger that lets the NEXT launch
+    /// find this process should this one die without tearing it down.
+    var processIdentifier: pid_t { process.processIdentifier }
+
     /// - Parameter argv: complete, including `ssh` at index 0 (the shape
     ///   `Configuration.argv` produces and the enrollment service already uses).
     init(argv: [String], sshExecutableURL: URL = URL(fileURLWithPath: "/usr/bin/ssh")) throws {
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift
@@ -1,0 +1,131 @@
+import Foundation
+
+#if canImport(Darwin)
+import Darwin
+#endif
+
+/// Kills forward `ssh` children a previous app run left behind, before this
+/// run's forwards dial the same remote port.
+///
+/// The bug this closes (field report, 2026-08-05): quit-and-reopen sometimes
+/// landed the pane on "Port held — close ssh sessions to that host." with a
+/// Retry that could only fail. The holder was this Mac's own orphan — an
+/// `ssh -N -R` from a run that ended without `applicationWillTerminate`
+/// (crash, force-quit) or whose teardown outran the bounded quit drain. It
+/// reparents to launchd, keepalives keep it healthy forever, and nothing else
+/// ever kills it.
+///
+/// Safety rules, in order of importance:
+///
+/// * **Never kill by pid alone.** A record is actioned only when the pid's
+///   CURRENT kernel identity — start time and resolved executable path —
+///   equals what the ledger captured at spawn. A reused pid cannot match, so
+///   an innocent process cannot be signalled, and a mismatch just retires the
+///   record.
+/// * **Only run while this instance holds the listener.** The caller
+///   (`ClaudeRemoteForwardCoordinator`) gates the reap behind its listener
+///   bind, which is what makes a SECOND app instance harmless: it cannot bind
+///   8473 while the first instance lives, so it can never reap the first
+///   instance's healthy tunnels.
+/// * **Escalate like the supervisor does.** SIGTERM, a bounded wait, SIGKILL,
+///   a bounded wait — on the injected clock, since the supervisor's own suite
+///   set the no-wall-clock rule for this subsystem. A survivor of SIGKILL
+///   keeps its record, so the next launch tries again.
+public struct ClaudeRemoteForwardOrphanReaper: Sendable {
+    public typealias Inspect = @Sendable (pid_t) -> ClaudeRemoteForwardPidRecord?
+    public typealias SendSignal = @Sendable (pid_t, Int32) -> Void
+    public typealias SleepFor = @Sendable (Duration) async throws -> Void
+
+    private let ledger: ClaudeRemoteForwardPidLedger
+    private let inspect: Inspect
+    private let sendSignal: SendSignal
+    private let sleepFor: SleepFor
+    private let terminationGrace: Duration
+    private let killGrace: Duration
+    private let pollInterval: Duration
+
+    public init(
+        ledger: ClaudeRemoteForwardPidLedger,
+        inspect: @escaping Inspect = { ClaudeRemoteForwardProcessIdentity.snapshot(pid: $0) },
+        sendSignal: @escaping SendSignal = { pid, signalNumber in
+            #if canImport(Darwin)
+            _ = Darwin.kill(pid, signalNumber)
+            #endif
+        },
+        sleepFor: @escaping SleepFor = { try await Task.sleep(for: $0) },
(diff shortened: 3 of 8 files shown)


T0mSIlver:
> Independent review by opencode (GLM-5.2), commissioned on the initial commit 60cffe1.


# Code review: PR #226 — reap orphaned remote-forward ssh children at launch

## Verdict

**Approve with minor changes.** The core design is sound: pid-identity is checked against kernel start-time + resolved executable path (not pid alone), the reap is correctly gated behind the listener bind so a second instance cannot reap a healthy one's tunnels, the reap gate is a strict once-per-launch state machine (`pending → running → done`) that prevents forwards from starting before the reap finishes, and the tests are deterministic (no wall-clock, no real filesystem, no real signals). The findings below are about a real TOCTOU the doc claims is impossible, missing completion logs on a lifecycle path (repo rule), a misleading failure mode when `kill(2)` itself fails, and a small test-coverage gap on the executable-path field of the identity. None of them block the bug fix shipping.

## Findings

### 1. Important — doc overstates the kill-safety guarantee; the inspect→signal pair has a real (if tiny) TOCTOU

**`Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift:20-24, 77-87`**

The type doc says, as the load-bearing safety claim:

> A record is actioned only when the pid's CURRENT kernel identity — start time and resolved executable path — equals what the ledger captured at spawn. **A reused pid cannot match, so an innocent process cannot be signalled**

The implementation backs this with a single `inspect` *before* the signal:

```swift
guard let current = inspect(pid_t(record.pid)), current == record else { ... }
Log.claudeContext.notice(...)                      // <-- time passes
sendSignal(pid_t(record.pid), SIGTERM)             // <-- kill hits whatever is at pid NOW
```

There is no re-check between `inspect` and `sendSignal`, and a `Log.claudeContext.notice` call sits in the window. Concretely: inspect pid 4242 → matches → orphan dies → kernel reuses 4242 for an unrelated process → `sendSignal(4242, SIGTERM)` hits the unrelated process. The same race re-opens between every poll in `waitUntilGone` and the eventual SIGKILL at line 95. On macOS this is *practically* unreachable (incremental pid allocation, wrap-around at ~100k, deliberate skip of recently-used pids), so I do not think this is an exploitable blocker — but the doc's categorical "an innocent process cannot be signalled" overpromises what a single inspect-then-act pattern can deliver.

**Suggested fix:** either soften the doc to "the identity is verified immediately before each signal; a pid reused in the microsecond window between verify and signal is the residual, non-exploitable race", or — stronger — collapse the window by dropping the `Log.notice` between verify and signal (log *after* the SIGTERM is sent), and consider re-verifying identity right before the SIGKILL escalation.

---

### 2. Important — `kill(2)`'s return value is discarded, so an `EPERM` is misreported as "survived SIGKILL"

**`Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift:50-54, 92-105`**

```swift
sendSignal: @escaping SendSignal = { pid, signalNumber in
    _ = Darwin.kill(pid, signalNumber)   // return value discarded
}
```

The reaper's failure state — "survived SIGKILL; the remote port may stay bound" — is decided purely by polling identity afterwards. If the signal was never delivered (`EPERM` because the orphan was somehow re-parented to a uid we can't signal, or `ESRCH` because it died in the window), the poll still sees the process (or a different process at the same pid with same identity in the reuse edge case) and the user/log is told the orphan *survived* SIGKILL, when in fact it was never delivered. Same-uid reparenting to launchd makes this very rare in practice, but the existing repo rule for lifecycle paths ("loud … log requests, completions, and failures", AGENTS.md) is exactly the rule whose failure mode this is — the wrong failure is logged loudly.

**Suggested fix:** capture `kill`'s return and `errno`, and when it is non-zero log a distinct failure (`ESRCH` → "already gone", `EPERM` → "no permission, port may stay bound") and treat it as a non-escalatable terminal state rather than falling through to the "survived" branch.

---

### 3. Minor — completion of SIGTERM / SIGKILL is not logged on a lifecycle path

**`Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift:84-106`**

AGENTS.md: "Backend/lifecycle code paths log their requests, completions, and failures — a silent failure path is how the ensureReady single-flight bug cost an hour of remote probing. Keep new paths loud."

The reaper logs the request (notice at line 84) and the two failure escalations (errors at 92, 103), but the two completions — "orphan honoured SIGTERM, record retired" and "orphan killed by SIGKILL, record retired" — are silent. The only success signal is the absence of an error, which is the exact shape the rule warns against. Contrast the supervisor it mirrors (`ClaudeRemoteForwardSupervisor.tearDown`), which logs every state transition.

**Suggested fix:** one `Log.claudeContext.info(...)` in each of the two `if await waitUntilGone(...) { ledger.forget(...); return }` branches naming the host and the signal that worked.

---

### 4. Minor — the executable-path field of the identity has no test

**`Tests/localvoxtralTests/ClaudeRemoteForwardOrphanReaperTests.swift:62-71, 81-92`**

`ClaudeRemoteForwardPidRecord.executablePath` exists specifically as "belt-and-braces on top" of the start-time check (`ClaudeRemoteForwardPidLedger.swift:13-15`), and `snapshot` goes to the trouble of calling `proc_pidpath` rather than trusting `p_comm`. But every test fakes `inspect` with a `ProcessTableFake` whose `live` record is *constructed with whatever path the test chose*, and every test that asserts a non-match varies `startSeconds`, never `executablePath`. A future regression that dropped the path from the `Equatable` synthesis (or stopped populating it in `snapshot`) would pass this suite.

**Suggested fix:** one test where the ledger record has `/usr/bin/ssh` and the live record at the same pid+start time has `/bin/cat` (or where `proc_pidpath` returns an empty path), asserting the record is retired with no signal — pinning that the path field participates in identity, not just the start time.

---

### 5. Minor — reap processes records sequentially; worst case is `maxHosts × (terminationGrace + killGrace)`

**`Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift:69-75`**

`reap()` awaits each host's kill in a `for` loop. With `maxHosts = 32` and the default 2 s + 1 s graces, the worst case (every record a live orphan that survives SIGTERM) is ~96 s where the gate holds *every* forward start. In practice orphans are 1–2 and the common path returns at the first inspect, so this is fine — but it is a latency cliff the design allows and the doc does not name. Worth either noting the bound in the type comment or concurrency-bounding the loop (a `TaskGroup` would let records reap in parallel without changing the per-host contract).

---

### 6. Nit — `claudeRemoteForwardPort` is read once but the ledger file is keyed by hostID only, not by port

**`Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardPidLedger.swift:71-75`**

`defaultFileURL()` puts the ledger beside the host registry, which is correct, but the records themselves carry no `remoteForwardPort`. If a future per-install port change (issue #215's mechanism) lands between two launches, a stale record from the old port is still identity-matched and killed — which is actually the *right* behavior (the old orphan holds the old port regardless), so this is not a bug. Mentioning only because the ledger doc could say so explicitly, to head off a future "the port changed, so this record doesn't apply" reading.

---

## Verified correct

- **Gate prevents forwards starting before the reap finishes.** `ClaudeRemoteForwardCoordinator.swift:131-152`: `OrphanReapPhase.pending` transitions atomically (MainActor-serial) to `.running`, spawns exactly one `Task`, returns. `.running` short-circuits subsequent reconciles. The `Task` is the only thing that transitions to `.done`, and it re-enters `reconcile()` itself. `testForwardsWaitForTheOrphanReapBeforeStarting` + `testTheOrphanReapRunsExactlyOnce` pin both halves.
- **The reap cannot kill this instance's own live forwards.** Because of the gate, no `ClaudeRemoteForwardSupervisor` exists yet when the reap runs, so no record in the ledger can name a process this run spawned. `makeSupervisor`'s `pidLedger.remember` (line 91) is only reached on the first `start()` after `.done`.
- **Multi-instance argument is airtight for the documented cases.** `isListenerBound` is wired to `coordinator?.isListening ?? false` where `coordinator` is the *listener* coordinator in *this* process; `listener?.isRunning` only flips true after `bind(2)` succeeds, and `SO_REUSEADDR`-without-`SO_REUSEPORT` (line 178) means a second instance's bind fails. So a non-listener instance early-returns at the `guard isListenerBound()` (line 117) before the gate, and its `orphanReap` stays `.pending` harmlessly. `testNoReapWhileTheListenerIsUnbound` pins it.
- **Identity check is sound for the realistic threat model.** `proc_bsdinfo` start-time + `proc_pidpath` resolved path on top of pid (`ClaudeRemoteForwardPidLedger.swift:34-56`) covers reboot-pid-reuse (start time differs), rapid-restart-by-same-app (start time differs), and a same-pid impostor with a different binary (path differs). `testAReusedPidIsNeverSignalled` covers the first.
- **`forget` is correctly pid-scoped.** `ClaudeRemoteForwardPidLedger.swift:125-132` re-checks `records[hostID]?.pid == pid` under the lock, so a teardown racing a respawn for the same host cannot erase the new record. `testForgetIsPidScoped` pins it.
- **Corrupt / unreadable / missing ledger all read as empty.** `load()` (lines 134-157) catches throw → `[:]`, catches decode failure / version mismatch → `[:]`, treats nil data as `[:]`. Two of three are pinned by `testCorruptContentsReadAsEmpty` and `testMissingFileReadsAsEmpty`.
- **No wall-clock in tests.** `ClaudeRemoteForwardOrphanReaperTests` injects `sleepFor: { _ in }`; `ClaudeRemoteForwardCoordinatorTests` uses `TeardownGate` (continuation-based) + `Task.yield()` for ordering. No `Date()`, no real `Task.sleep` polling.
- **Tests don't touch the real filesystem or signal real processes.** `MemoryLedgerStore` is keyed by URL path and never hits disk; `ProcessTableFake` is the only thing on the receiving end of `sendSignal`.
- **`pollCount` arithmetic is correct** for the asserted cases (2 s / 50 ms → 40; 1 ms / 1 s → 1) and the `Int64` clamp protects against zero/negative durations. `testPollCountCoversTheGraceWindow` pins both.
- **Zombies are handled correctly.** `proc_pidpath` returns 0 for a zombie, `snapshot` returns nil, the reap retires the record — so the reap never blocks on an unreaped zombie.
- **The reap is bounded.** Per record: at most `terminationGrace + killGrace` (default 3 s). On app quit the reap `Task` is not awaited (it is fire-and-forget), so it cannot hang termination; process exit terminates it.
- **No new ledger write on clean process exit** — by design ("removed by the reaper, never on process exit", `ClaudeRemoteForwardPidLedger.swift:72-75`), so the file converges rather than racing the supervisor's own teardown.

---

## Response to the review — addressed in 4721f07

| # | Finding | Disposition |
|---|---|---|
| 1 | TOCTOU between inspect and signal; doc overclaims | **Fixed.** Doc now names the microsecond verify-to-signal residual instead of claiming impossibility; the notice log moved after the SIGTERM (it sat inside the window); and the SIGKILL escalation now re-verifies identity immediately before firing — `waitUntilGone`'s last poll was up to one interval stale, and SIGKILL is the one signal nothing can decline. |
| 2 | `kill(2)` return discarded; EPERM misreported as "survived SIGKILL" | **Fixed** (logging half). The default `sendSignal` now logs any non-ESRCH failure with its errno, so the log names the true cause. The decision logic stays poll-based on purpose: EPERM against a same-uid child we spawned is not a reachable state worth its own terminal branch, and the poll's answer ("still there / gone") is the one the port cares about. |
| 3 | Success completions not logged | **Fixed.** Both branches ("honoured SIGTERM", "killed") now log with the host id, plus a third for "exited before SIGKILL". |
| 4 | `executablePath` untested in identity | **Fixed.** `testADifferentExecutableAtTheSamePidIsNeverSignalled`: same pid + same start time, different binary → retired, no signal. Suite is now 61 tests. |
| 5 | Sequential reap worst-case latency | **Doc'd**, not parallelized: real orphan counts are one or two, the common path returns at the first inspect, and only the forwards wait (the listener is already up). A `TaskGroup` here buys nothing measurable and costs the simple sequential contract. |
| 6 | Ledger records carry no port | **Doc'd** as deliberate: an orphan holds whatever port it was spawned with, so a per-install port change between launches must not exempt the old orphan from reaping. |

Post-fix proof: `./scripts/remote-build.sh test --filter ClaudeRemoteForward`
```
Test Suite 'ClaudeRemoteForwardCoordinatorTests' passed at 2026-08-05 23:22:47.790.
Test Suite 'ClaudeRemoteForwardOrphanReaperTests' passed at 2026-08-05 23:22:47.792.
Test Suite 'ClaudeRemoteForwardPidLedgerTests' passed at 2026-08-05 23:22:47.793.
	 Executed 61 tests, with 0 failures (0 unexpected) in 0.045 (0.049) seconds
```

PR #228, opened 2026-08-06
SSH probe: ProxyJump machinery is not a second connection; abstentions name their cause

T0mSIlver:
## What

Fixes the field abstention that blocked every remote-herdr join for a `ProxyJump` ssh config (owner dogfooding, 2026-08-06): OpenSSH spawns the jump hop as `ssh -W <ip>:22 <jumphost>` — a direct child on the SAME tty, in the SAME foreground process group — so `SSHDestinationTTYProbe` saw two foreground ssh clients on the surface and returned `.undeterminable` on every dictation. The `-W` child also poisoned the machine-wide uniqueness scan (its argv parse is refused, which read as "an ssh we cannot account for").

The rule: an ssh that is a **direct child of another scanned ssh is that connection's machinery, not a second connection** — invisible to both the exactly-one surface count and the uniqueness claim. The partition rides kernel ppid, which no launcher gets to write, so it cannot hide a connection (the demoting parent is itself counted). Deliberately preserved paranoia:

- sibling ssh processes in one foreground group (review round 7) have no ssh parent and still abstain;
- a shell-mediated `ProxyCommand` grandchild stays a root and still abstains — an extra root can only cause an abstention, never a join;
- an unreadable root elsewhere still spoils uniqueness, exactly as before.

Second change, same story: `.undeterminable` now carries a content-free `SSHProbeIndeterminacy` category (device/table read, multiple foreground clients, untrusted executable, unreadable argv, refused argv, probe unavailable) into `Log.claudeContext` and the dogfood join-abstention string. Three field dictations in a row were diagnosed blind because every branch collapsed into one word; the actual cause had to be reconstructed from the remote host's process table.

`docs/agent/invariants.md` updated to state the machinery rule and its bounds.

[dogfood-package]

## Proof

- Unit suite green — full run after the fix:
  `Executed 2613 tests, with 2 tests skipped and 0 failures (0 unexpected) in 95.990 (96.131) seconds`
- Bug fix: regression tests added, shown failing before the fix (commit 1, `b284e32`) and passing after (commit 2):
  - Red (pre-fix), `remote-build.sh test --filter SSHDestinationTTYProbeTests`:
    ```
    testAProxyJumpChildIsMachineryNotASecondSurfaceClient : XCTUnwrap failed: expected non-nil value of type "SSHSurfaceConnection"
    testAMultiHopProxyChainIsEntirelyMachinery : XCTUnwrap failed: expected non-nil value of type "SSHSurfaceConnection"
    testAnotherTerminalsProxyChildDoesNotRemoveUniqueness : XCTAssertTrue failed
    Executed 64 tests, with 3 failures (0 unexpected)
    ```
  - Green (post-fix): `Executed 65 tests, with 0 failures (0 unexpected)` (probe suite) and `Executed 51 tests, with 0 failures (0 unexpected)` (RemoteHerdrJoinTests, covers the categorized abstention falling through to the title marker).
- `testAProxyJumpChildIsMachineryNotASecondSurfaceClient` is the field shape verbatim (`ssh -t sandbox-vpn herdr` + child `ssh -W [ip]:22 dell1`, same pgid/tty, from the owner's Mac `ps` output).
- Review-property preservation pinned by test: `testSiblingForegroundClientsAreNotMachineryAndStillAbstain`, `testAnotherTerminalsRootToTheSameHostStillRemovesUniquenessDespiteItsChild`, `testAnOrphanedUnreadableSSHElsewhereStillRemovesUniqueness`.
- LLM lanes: not required — no change to prompts, models, or anything that reaches the model; this alters only whether a join resolves, and the join/context paths' behavior is covered by the unit suites above. (Paths are under `Sources/localvoxtral/ClaudeContext/*`, so the lane filter will run the polishd lane anyway.)
- Field verification planned: owner installs via `./scripts/try-pr.sh <this PR> --dogfood` and dictates over the ProxyJump connection; the dogfood record should now either join via `.remoteHerdrPane` or name the exact refusal category.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/SSHDestinationTTYProbe.swift
+++ b/Sources/localvoxtral/ClaudeContext/SSHDestinationTTYProbe.swift
@@ -27,6 +27,30 @@ struct SSHSurfaceConnection: Sendable, Equatable {
     }
 }
 
+/// WHY the probe could not pin an ssh session down. Deliberately content-free
+/// — a category, never a path, host, or option letter — so it is safe to put
+/// in the log and the dogfood record. Three field dictations were diagnosed
+/// blind (2026-08-06) because every branch collapsed into one word; the cause
+/// had to be reconstructed from the REMOTE host's process table.
+enum SSHProbeIndeterminacy: String, Sendable, Equatable {
+    /// The tty device path did not resolve.
+    case deviceUnreadable = "device unreadable"
+    /// The machine-wide process scan failed. Not evidence of absence.
+    case tableUnreadable = "process table unreadable"
+    /// More than one foreground ssh CONNECTION on the surface.
+    case multipleForegroundClients = "multiple foreground ssh"
+    /// The surface process's executable is not one we trust to be OpenSSH.
+    case untrustedExecutable = "untrusted executable"
+    /// The surface process's argv could not be read.
+    case unreadableArguments = "unreadable argv"
+    /// The argv was read and refused: a destination-moving or non-interactive
+    /// option, an unknown option letter, or a destination shape outside the
+    /// grammar.
+    case refusedArguments = "refused argv"
+    /// The seam default: no live probe was injected, so the arm is disabled.
+    case probeUnavailable = "probe unavailable"
+}
+
 /// What the process table says about ssh clients on ONE terminal device.
 ///
 /// Three answers, and the difference between the last two is the whole point:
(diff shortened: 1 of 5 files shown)


T0mSIlver:
opencode (GLM-5.2) adversarial review: **approve-with-nits**. Verified: machinery rule cannot demote a genuine connection (kernel ppid, orphan reparenting targets launchd, conservative direction — a residual race abstains rather than mis-joins); round-7 sibling refusal, unreadable-root uniqueness, tty-less exclusion all preserved; cause categories content-free by type. Findings applied in `843f615`: orphan test now uses a real reparented ppid (1), sibling test pins `.multipleForegroundClients` exactly, multi-hop test data uses one foreground pgid per device, self-parent guard annotated as test-scaffolding defense. Probe suite 65/65 after fixes.

PR #229, opened 2026-08-06
Remote herdr join: classify the herdr invocation; block competing herdr views, not every second connection

T0mSIlver:
## What

Stacked on #228. Makes the remote-herdr join survive the normal multi-connection workflow (owner request, 2026-08-06): today, keeping *any* second ssh session to the enrolled host — a plain shell for builds — kills every join via the machine-wide uniqueness rule, and the abstention falls through to the title-marker arm where a stale marker can win.

Design follows an Opus research memo with herdr protocol facts verified in source (`herdrdev/herdr@master`, v0.8.0, protocol 19):

- **herdr focus is server-global and multi-client attach is a mirror** (`src/app/api/panes.rs::handle_pane_current` resolves the app's single active pane; `tests/multi_client.rs` proves frames broadcast to all clients, shared agent-panel scroll, min-size clamp). Two whole-view clients of one server display the SAME focused pane — abstaining between them protects nothing.
- herdr still exposes **no** read-only attachment signal at protocol 19 (only `client.window_title.set/clear`, both mutations; no client records in `session.snapshot`; no client lifecycle events) — so the invocation argv remains the binding, and it is trustworthy: the exec-time vector of a verified OpenSSH binary is the command ssh actually ran.

Two changes:

1. **`indicatesHerdr: Bool` → `HerdrInvocation` classification.** Only a bare whole-view client (`herdr`, optionally `--session <name>` / `--session=<name>`) proceeds. This also closes two mis-joins reachable **today with a single connection**, which the memo found: `ssh host herdr terminal attach <id>` renders ONE pane, and `herdr --session <x>` attaches a DIFFERENT server (named sessions have separate sockets) — both satisfied the boolean while the join reads the server-global focus of the candidates' socket. Unknown herdr verbs/flags land in `.otherHerdrSubcommand` and are refused, so future herdr verbs cannot silently become mis-joins.
2. **Blanket uniqueness → competing-herdr-view rule.** Another tty-holding ssh root blocks only when it *could be displaying a different herdr view* of this destination: a client with a non-identical `--session` selector, a herdr subcommand shape, a refused argv that mentions `herdr` anywhere (substring — one-sided, can only over-block), or anything unreadable/untrusted. Plain shells and same-selector whole-view clients stop blocking.

Deliberately preserved: the surface's own argv must still be a herdr client (round 5b — sole plain ssh never joins, detached-herdr case); suspended connections on the same device still count; unreadable roots still block; sibling foreground sshs still abstain; every fail-closed confirmation downstream (single socket, pane id, broker marker, agent_session, foreground) is untouched. `docs/agent/invariants.md` states the server-global focus model as an explicit **external assumption** with an upgrade-recheck obligation.

[dogfood-package]

## Proof

- Full unit suite green after the change: `Executed 2623 tests, with 2 tests skipped and 0 failures (0 unexpected) in 97.949 (98.090) seconds`
- Affected suites: `SSHDestinationTTYProbeTests` 75/75, `RemoteHerdrJoinTests` 51/51.
- Red/green: the API rename (`isOnlyConnectionToDestination`/`indicatesHerdr` → `hasCompetingHerdrClient`/`herdr`) makes a literal pre-fix compile impossible, so the red run reproduces the PRE-CHANGE semantics under the new API (boolean herdr signal + blanket same-destination refusal) as a two-line mutation:
  ```
  Executed 75 tests, with 15 failures (0 unexpected) in 0.247 (0.250) seconds
  failing: testAPlainShellToTheSameHostDoesNotCompete,
  testASecondWholeViewClientOfTheSameServerDoesNotCompete,
  testAHerdrSubcommandOnTheSurfaceIsClassifiedAsOther,
  testARefusedArgvElsewhereCompetesOnlyWhenItMentionsHerdr,
  testMatchingSessionSelectorsDoNotCompete, testHerdrCommandClassification,
  testASessionSelectorOnTheSurfaceIsCarried, … (15 total)
  ```
  Mutation reverted → 75/75 green (same command).
- Review-property preservation pinned by test: `testASuspendedHerdrVerbOnTHISDeviceStillCompetes` (round 5b transposed), `testAnUnreadableSSHElsewhereCompetes`, `testAnUntrustedExecutableElsewhereCompetes`, `testTwoForegroundSSHClientsOnOneSurfaceAbstainEvenToTheSameHost` (round 7), `RemoteHerdrJoinTests.testAPlainSSHSessionNeverReachesTheArmEvenAsTheSoleConnection`.
- LLM lanes: paths are under `Sources/localvoxtral/ClaudeContext/*` so the polishd lane runs by filter; no prompt/model/eval change in this PR.
- Field verification planned: owner installs the stack via `./scripts/try-pr.sh <this PR> --dogfood`, keeps a second plain ssh to `sandbox-vpn` open, dictates — expected `arm: remoteHerdrPane` with the extra connection present; then opens a `--session scratch` client and confirms the join abstains with the named cause.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/SSHDestinationTTYProbe.swift
+++ b/Sources/localvoxtral/ClaudeContext/SSHDestinationTTYProbe.swift
@@ -3,27 +3,63 @@ import Foundation
 #if canImport(Darwin)
 import Darwin
 
+/// What an ssh remote command says about herdr — a CLASSIFICATION, not a
+/// boolean, because "mentions herdr first" spans shapes that display entirely
+/// different things (verified against herdr v0.8.0 / protocol 19, 2026-08-06):
+///
+/// * a bare client attach displays the server's GLOBAL focused pane — herdr
+///   focus is server-global and multi-client attach is a mirror
+///   (`src/app/api/panes.rs::handle_pane_current`, `tests/multi_client.rs`),
+///   which is the semantic the whole remote arm reads through `pane.current`;
+/// * `herdr terminal attach <id>` renders exactly ONE pane, and
+///   `herdr --session <name>` attaches a DIFFERENT server (named sessions have
+///   separate sockets) — joining the global focus of the candidates' socket
+///   while the user looks at either of those is a mis-join reachable with a
+///   single connection, which is why the boolean had to go.
+enum HerdrInvocation: Sendable, Equatable {
+    /// The remote command is not herdr (or there is no remote command).
+    case notHerdr
+    /// A bare `herdr` whole-view client attach — the only shape whose display
+    /// is the server's global focus. `sessionSelector` is the normalized
+    /// `--session` value (nil = the default session); two connections display
+    /// the same server only when their selectors are byte-identical.
+    case plainClient(sessionSelector: String?)
+    /// herdr with any other subcommand or flag shape: single-pane attaches,
+    /// observers, named-session verbs, `server`, or tokens this classifier
+    /// does not know. Refused by the arm — never a join, never a guess.
+    case otherHerdrSubcommand
+}
+
 /// The ssh connection a terminal surface is displaying, as far as the local
 /// process table can prove it.
 struct SSHSurfaceConnection: Sendable, Equatable {
     /// Normalized destination host/alias.
     var destination: String
-    /// No OTHER terminal on this machine hosts an ssh to the same destination.
+    /// Another terminal on this machine hosts an ssh to the same destination
+    /// that COULD be displaying a different herdr view: a herdr client with a
+    /// different session selector, a single-pane herdr attach, or an ssh whose
+    /// argv cannot rule herdr out.
     ///
-    /// Without this the probe binds to a HOST rather than to a CONNECTION: with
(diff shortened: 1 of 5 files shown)


T0mSIlver:
opencode (GLM-5.2) adversarial review: **approve-with-nits**, one MAJOR finding — the selector comparison proves same-server only while the remote derives sockets from the selector alone; a shell with `HERDR_SOCKET_PATH`/`XDG_RUNTIME_DIR` exported can attach two bare `herdr` invocations to different servers, invisible from the Mac. Addressed in `19a8c0a` as an explicitly documented, deliberately accepted residual (invariants.md states it plus the downstream bounds: candidates spanning two sockets abstain at the single-socket rule; pane-id + broker-marker must still agree). Also applied: repeated `--session` now refused instead of last-wins, resolver reports the surface's own classification before neighbor competition (accurate abstention cause), suspended bare-herdr mirror test (the round-5b shape under the new rule), notHerdr-surface competition boundary pinned, `mentionsHerdr` hostname over-block documented. Reviewer verified: sole plain ssh never joins, suspended herdr-verb still blocks, round-7 abstains, classification conservative on every unknown shape. Full suite 2625/2625 after fixes.

PR #231, opened 2026-08-08
Overlay: name the Claude Code session this dictation is grounded in

T0mSIlver:
## What

Names the Claude Code session a dictation is grounded in, in the overlay buffer's header — or says that none attached.

A join that silently did not happen and one that did look identical today. The transcript commits either way; the only difference is whether the polish call was grounded, and you find out by reading text that mis-transcribed the filenames you just said. That is how the ProxyJump/herdr abstention #228 fixes was found in the first place.

**Why the overlay rather than a status line or the menu bar.** The join is resolved on this Mac at dictation start, and the overlay opens on this Mac at dictation start. So the badge is a read of the ONE resolved answer: no wire format, no stamp file, no staleness gate, no host-vs-session attribution, no parsing in the remote shim, and nothing new on the dictation-start latency path (the probe already ran). Coverage is exact rather than partial: `capturedClaudeJoin` is consumed only inside the `if let endpointURL = polishingConfig?.endpointURL` branch, and polish is Overlay Buffer only — the join never matters where there is no overlay.

**Three states, not a boolean.**

| Badge | Meaning |
|---|---|
| `🔗 <workspace>` | this dictation is grounded in that session |
| `🔗̸ No Claude session` | live sessions exist and none attached |
| *(nothing)* | neither context feature is on, or no session could have joined |

A checkmark cannot tell a correct join from a mis-join onto a stale session — the residual documented on the marker and remote-herdr arms, where a lingering title marker from an earlier session can win. Naming the workspace is what makes a wrong join recognizable at a glance. And a Mac that is not running Claude Code wears no permanent complaint: silence is the honest answer when nothing was attempted.

Same quiet weight as the existing "Polished" pill, deliberately — an unjoined dictation still commits its text correctly, so this annotates the panel. It is not an error, and the warning color under the transcript stays reserved for text that failed to insert. The distinction rides the icon (`link` / `link.slash`) and the label, not color, so it needs no second appearance-matched palette.

**Invariants kept.** The badge DESCRIBES the resolved join; it never resolves one. Re-asking at display time is exactly what the once-per-dictation rule exists to prevent — a second resolution could name a different session than the context actually attached. Its one extra read is `registry.hasLiveSessions()`, which touches no title, TTY, socket, or process table, and is consulted only when no join resolved (a resolved join already proves a session exists). `docs/agent/invariants.md` updated.

The label is a length-capped workspace `displayName` with control characters neutralized. A LOCAL workspace name is the last component of a real directory, and a macOS path component forbids only `/` and NUL — a newline is a legal directory name, and the panel is measured from the body text alone, so a label carrying a line break would draw outside the height the panel was sized for. Cc (newline, tab) becomes a space; deleting it would glue two runs into a directory the user does not have. Cf (bidi overrides, zero-width joiners) is dropped, being zero-width already.

Ordering trap, documented at both ends: the join resolves before the realtime socket connects, but `startSession` clears the state machine — so the badge is latched on the view model (exactly like `sessionSecureInputActive`) and pushed after `startSession`, not at resolution time.

[dogfood-package]

## Proof

- Unit suite green — full run after the change:
  `Executed 2638 tests, with 2 tests skipped and 0 failures (0 unexpected) in 105.216 (105.354) seconds`
- New suite alone, `./scripts/remote-build.sh test --filter OverlayClaudeJoinBadgeTests`:
  `Executed 14 tests, with 0 failures (0 unexpected) in 0.005 (0.005) seconds`
- Red/green, and it caught a real deficiency rather than confirming one. The first full run failed on the label test:
  ```
  OverlayClaudeJoinBadgeTests.swift:189: error: testControlCharactersAreStrippedFromTheLabel :
  XCTAssertEqual failed: ("Optional("reponamex")") is not equal to ("Optional("repo name x")")
  Executed 2638 tests, with 2 tests skipped and 1 failure (0 unexpected) in 105.953 (106.101) seconds
  ```
  `\n` and `\t` are Cc, so filtering `CharacterSet.controlCharacters` deleted them outright and glued `repo\nname` into `reponame` — a label naming a directory that does not exist. Fixed by splitting the two categories (Cc → space, Cf → dropped) rather than by relaxing the expectation; green above.
- Behavior pinned by name, not just by count:
  - `testAResolvedLocalJoinNamesItsWorkspace`, `testAResolvedRemoteJoinNamesItsOpaqueLabel` — joins built through the REAL registry + resolver, so the label under test is what the wire path actually produces per origin.
  - `testAJoinWithoutAWorkspaceStaysJoined` — a cwd-less join is still a real join; falling back to `.unjoined` would call a working setup broken.
  - `testNoJoinWithLiveSessionsIsUnjoined` / `testNoJoinAndNoSessionsIsSilent` / `testFeaturesOffHideTheBadgeEvenWithAJoin` — the three-way gate.
  - `testTheRegistryIsNotConsultedWhenAJoinResolved`, `testTheRegistryIsNotConsultedWhenFeaturesAreOff` — the extra read stays off the paths that do not need it.
  - `testBidiOverridesAreStrippedFromTheLabel`, `testControlCharactersAreStrippedFromTheLabel`, `testWhitespaceRunsCollapseSoALabelCannotPadThePill`, `testAnOverlongLabelIsTruncatedWithAnEllipsis`, `testALabelAtTheLimitIsLeftAlone`, `testALabelOfNothingButControlCharactersIsRejected`.
  - `OverlayBufferStateMachineTests.testClaudeJoinBadgeRidesTheSessionAndClearsOnTheNextOne` / `…IsIgnoredWhenIdle` — the badge annotates the whole session including the polished hold and a failed commit, and a new session starts clean. These two are also what pin the ordering rule the view model depends on: set before `startSession`, it is dropped.
  - `ClaudeSessionRegistryTests.testHasLiveSessionsTracksLiveSessionsExactly` — asserted equal to `!liveSessions().isEmpty` on both sides of the TTL, on the injected clock (no wall-clock).
- LLM lanes: nothing here alters what reaches the model — the badge is display-only and reads a join that was already resolved. `Sources/localvoxtral/ClaudeContext/*` is in `llm-lane-filter.sh`, so the polishd lane runs by filter anyway.
- **UI change — NOT yet verified by hand.** This was written from a Linux box; I cannot run the app. What is verified is the state machine's contract, the badge derivation, and the label sanitizer, all above. What is NOT: how the pill actually looks in the header next to a long phase title (the secure-input title is the worst case) and next to the "Polished" pill, and whether `.layoutPriority(-1)` truncates the workspace name the way it should rather than squeezing the title. `[dogfood-package]` is set — owner please install via `./scripts/try-pr.sh <this PR> --dogfood`, dictate once with a Claude session joined and once without, and confirm both pills coexist without pushing the title out.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeSessionRegistry.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeSessionRegistry.swift
@@ -624,6 +624,19 @@ public final class ClaudeSessionRegistry: Sendable {
         }
     }
 
+    /// Whether ANY session is live, without materializing them.
+    ///
+    /// The overlay's join badge asks this once per unjoined dictation purely to
+    /// decide between "nothing attached" and silence, and it needs no snapshot
+    /// — `liveSessions()` would copy every session's recent files and snippets
+    /// to answer a question about emptiness.
+    public func hasLiveSessions() -> Bool {
+        let timestamp = now()
+        return state.withLock { state in
+            state.sessions.values.contains { isFresh($0, now: timestamp) }
+        }
+    }
+
     public func evict(sessionID: String) {
         state.withLock { removeLocked(&$0, sessionID: sessionID) }
     }
--- a/Sources/localvoxtral/ClaudeContext/TerminalScreenClaudeJoin.swift
+++ b/Sources/localvoxtral/ClaudeContext/TerminalScreenClaudeJoin.swift
@@ -300,6 +300,18 @@ struct ClaudeSessionJoinResolver {
     /// can ONLY join via marker, because their TTY names another machine's
     /// device and `resolve(tty:)` refuses remote candidates outright.
     ///
+    /// Whether the registry currently holds ANY live session.
+    ///
+    /// Not a join and not a step toward one: the overlay's join badge asks it
+    /// to tell "nothing attached" apart from "there was nothing to attach", so
+    /// a Mac that simply is not running Claude Code shows no badge instead of a
+    /// permanent complaint. It reads no title, no TTY, no socket and no
+    /// process table — it cannot influence, or be influenced by, what `resolve`
+    /// decides.
+    func hasLiveSessions() -> Bool {
+        registry.hasLiveSessions()
+    }
+
     /// This remains the only place a join is resolved, once per dictation, at
     /// start — whichever mechanism answers.
     func resolve(target: TerminalScreenTarget) async -> ClaudeSessionJoin? {
--- a/Sources/localvoxtral/DictationOverlayController.swift
+++ b/Sources/localvoxtral/DictationOverlayController.swift
@@ -95,7 +95,8 @@ final class DictationOverlayController {
             errorMessage: nil,
             secureInputActive: false,
             metrics: OverlayLayoutMetrics(bodyFontSize: fontSizeProvider()),
-            polished: false
+            polished: false,
+            claudeJoin: .hidden
         )
         hostingView = TransparentHostingView(rootView: initialView)
         // Without this, NSHostingView probes the SwiftUI content at ∞×∞ and
@@ -143,7 +144,8 @@ final class DictationOverlayController {
             errorMessage: snapshot.errorMessage,
             secureInputActive: snapshot.secureInputActive,
             metrics: metrics,
-            polished: snapshot.polished
+            polished: snapshot.polished,
+            claudeJoin: snapshot.claudeJoin
         )
 
         let contentHeight = metrics.contentHeight(
--- a/Sources/localvoxtral/DictationOverlayView.swift
+++ b/Sources/localvoxtral/DictationOverlayView.swift
@@ -145,10 +145,10 @@ struct DictationOverlayView: View {
     private var polishedBadge: some View {
         Label("Polished", systemImage: "wand.and.stars")
             .labelStyle(.titleAndIcon)
-            .font(.system(size: 10, weight: .semibold))
+            .font(.system(size: metrics.badgeFontSize, weight: .semibold))
             .foregroundStyle(.secondary)
-            .padding(.horizontal, 6)
-            .padding(.vertical, 1)
+            .padding(.horizontal, metrics.badgeHorizontalPadding)
+            .padding(.vertical, metrics.badgeVerticalPadding)
             .background(
                 Capsule(style: .continuous).fill(Color.primary.opacity(0.08))
             )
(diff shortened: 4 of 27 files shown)


T0mSIlver:
Applied all three findings from the opencode review (e809de1).

**5a — badge folded into `startSession`.** It was a follow-up `markClaudeJoin` call that had to run *after* `startSession` (which resets the panel) to survive, with nothing in the type system saying so. It is now a parameter of `startSession(preResolvedAnchor:claudeJoin:)` and the setter is gone entirely. The review's reasoning holds: unlike `polished`, which genuinely arrives later at commit, the join is always already known at session start — it resolves before the realtime socket connects and the panel opens after it. This removes the one load-bearing ordering the PR could not unit-test, rather than testing it. 12 test doubles updated for the signature.

**6a — header pills now scale with the overlay font setting.** This turned out to be a real, measurable bug rather than a theoretical one, and it was pre-existing (it applied to the "Polished" badge equally). Red run with the old hardcoded 10pt, `--filter OverlayLayoutMetricsTests`:

```
OverlayLayoutMetricsTests.swift:136: error: testBadgePillFitsTheHeaderAtEverySize :
XCTAssertLessThanOrEqual failed: ("14.0") is greater than ("13.0") - badge pill overflows the header at body size 10.0
OverlayLayoutMetricsTests.swift:118: error: testBadgeFontScalesWithTheBodyFontSetting :
XCTAssertEqualWithAccuracy failed: ("1.1818181818181819") is not equal to ("0.4924242424242424") +/- ("0.0001") - the pill keeps its proportion to the title at every size
Executed 10 tests, with 4 failures (0 unexpected) in 0.211 (0.212) seconds
```

`headerHeight` is `ceil(16 * scale)`, so it shrinks with the body-size setting while an unscaled pill did not — 14pt of pill in 13pt of header at the minimum, and a speck at the maximum. `OverlayLayoutMetrics` gains `badgeFontSize` / `badgeHorizontalPadding` / `badgeVerticalPadding`, so both pills now scale from the single body-size setting like every other font in that file — which its own header comment already promised ("All other fonts and the panel width scale proportionally so the buffer keeps its shape at any size"). Green after the fix: `Executed 10 tests, with 0 failures`.

**3a — U+2028/U+2029 asserted.** The doc comment claimed the whitespace collapse covers the line/paragraph separators; now `testUnicodeLineAndParagraphSeparatorsCollapseToASpace` pins it, so an ASCII-only refactor cannot silently let a line separator into a single-line header.

**6b** — run-on line in `invariants.md` fixed, and the bullet now states the parameter-not-setter rule.

Full suite after all of it: `Executed 2641 tests, with 2 tests skipped and 0 failures (0 unexpected) in 107.293 (107.433) seconds`.

The hand-verification caveat in the PR description still stands, and 6a slightly raises its value: the pill sizing changed at both ends of the font-size range, so the dogfood pass is worth doing at a non-default overlay font size as well as the default.

PR #236, opened 2026-08-09
Remote herdr join: agents-panel nonce binding as primary surface authorization (live mic indicator, marker-arm suppression, enroll-time config offer)

T0mSIlver:
## What

Top of the stack. Implements `docs/agent/remote-herdr-panel-binding.md` (committed in this PR — the owner-approved design, 2026-08-09): the herdr **agents-panel nonce binding** becomes the PRIMARY surface authorization for `.remoteHerdrPane` joins, replacing ssh-argv inference for the positive path.

Per plausible enrolled host (readable ssh destination when available; when an ssh is present but unreadable, ≤3 speculative hosts with live herdr-bearing sessions; **a surface with no ssh at all never probes** — local shells stay zero-cost), the resolver keeps the single-socket rule, opens the forward, finds the unique live session claiming herdr's focused pane, stamps that pane via `pane.report_metadata` with a fresh `lv-mic-<nonce>` token (~51.7 random bits, fixed 17 columns, 8 s TTL), and requires that exact token in the focused terminal's existing grid route within two injected-clock reads. herdr renders the sidebar only for whole-view App clients (attach/observe render the raw pane alone), so a grid match positively proves the surface displays a whole-view client of the stamped server — the discriminator #229 approximated with argv, obtained directly. All downstream confirmations are unchanged: broker marker in the pane title, `agent_session` claim, foreground process.

This closes the two abstention classes argv could never close: the **Ghostty ssh-wrapper** surface (unreadable argv → speculative panel proof) and the **manual flow** (`ssh host`, then typing `herdr` — readable destination, panel-authorized despite `.notHerdr` argv).

Also in this PR, per the owner's decisions:

- **Live mic indicator**: a matched token stays visible in the panel for the dictation (refresh ~4 s / TTL 8 s, injected clocks), explicitly cleared before its forward closes; the view model owns the lease on every exit path.
- **Title-marker suppression**: the four "ssh present but unreadable" probe categories (`multipleForegroundClients`, `untrustedExecutable`, `unreadableArguments`, `refusedArguments`) now suppress the title-marker fallback — killing the documented stale-marker mis-join residual — while probe-wide categories (`deviceUnreadable`, `tableUnreadable`, `probeUnavailable`) leave local marker joins intact. Each side pinned by test.
- **Enroll-time config offer**: consent-gated, conservative remote patch of `[ui.sidebar.agents]` over the existing enrollment ssh channel — appends the three-row block only when no agents table/rows key exists (including trailing-comment header variants), otherwise surfaces the snippet in Settings; inferential "row likely not configured" diagnostics only from destination-known probes.
- Content-free abstention categories for every new failure path (log + dogfood), and the invariants doc restates the trust argument with its herdr-upgrade external-assumption obligations.

Review-hardened twice:

- Orchestrator review: speculative probing originally fired on `.noSSHClient` (the design doc's own error) — every local-shell dictation would have paid cold-forward latency and flashed nonces in remote panels, with false config diagnostics. Amended: no-ssh surfaces never probe; speculation requires an ssh present-but-unreadable; diagnostics gated to destination-known probes. The original `noSSHClient` test had passed only because panel metadata was unwired (passing-for-the-wrong-reason); it now pins the fix with the machinery fully wired.
- GLM adversarial review (SHIP-WITH-FIXES): its potential blocker — JSON `null` vs empty-string clear — was resolved against herdr source (`tokens: HashMap<String, Option<String>>`, `normalize_metadata_tokens` pins both `None` and `""` as clears) and documented; applied fixes: trailing-comment config detection, panel-authorized view-model lifecycle coverage (with concurrent `stopAndWait` callers awaiting one clear/close), and coalesced `rowNotRendered` dogfood notes. GLM independently confirmed: no path reaches a `.remoteHerdrPane` join without a grid-verified nonce for the exact stamped socket or the full #229 argv chain; per-host fresh nonces cannot cross-authorize; downstream confirmations not weakened; no nonce/host/path in any log.

[dogfood-package]

## Proof

- Full unit suite green (after both review rounds): `Executed 2653 tests, with 2 tests skipped and 0 failures (0 unexpected)`
- Red/green pairs:
  - Marker suppression: `testUnreadableSSHOnSurfaceSuppressesAStaleTitleMarkerJoin` — red `XCTAssertNil failed: received a .titleMarker join` → green.
  - Panel-primary: `testPanelBindingIsPrimaryWhenSSHArgvSaysPlainShell` — red `XCTUnwrap failed` → green.
  - Enrollment patch: `testHerdrPanelConfigurationUsesTheEnrollmentSSHChannelAndConservativePatch` — red (1 failure in 68) → green.
  - Orchestrator amendment: `testNoSSHClientOnTheSurfaceFallsThroughToTheTitleMarker` (strengthened, panel wired) — red `Executed 1 test, with 3 failures` → green; `testASpeculativeSettleTimeoutDoesNotClaimTheRowIsMissing` — red `XCTAssertTrue failed` → green.
  - GLM F2: trailing-comment header — red `71 tests, 3 failures` (exit 127, duplicate table appended) → green 71/71.
  - GLM F3: view-model lifecycle — red (3 assertions, `stopAndWait` returned early) → green 62/62.
- Targeted suites: `RemoteHerdrJoinTests` 62/62, `HerdrPanelBindingProbeTests` 6/6, `ClaudeRemoteEnrollmentServiceTests` 71/71, `ClaudeIntegrationSettingsModelTests` 55/55, `HerdrSocketClientTests` 18/18.
- LLM lanes: not required — nothing that reaches a model changes (ClaudeContext paths run the polishd lane by filter).
- UI: Settings shows the config offer/status in the existing Context group (content only — group structure unchanged per owner rule); popover strings are one short sentence. Not hand-verified on a Mac; covered by `ClaudeIntegrationSettingsModelTests`. Field verification planned: owner installs the stack via `./scripts/try-pr.sh <this PR> --dogfood`, adds the `$lvmark` row on the enrolled host (or accepts the enroll-time offer), and dictates over the ProxyJump connection with the Ghostty wrapper active — expected `arm: remoteHerdrPane` with the mic row visible, and named categories on any abstention.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeIntegrationSettingsModel.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeIntegrationSettingsModel.swift
@@ -259,6 +259,7 @@ public final class ClaudeIntegrationSettingsModel {
         /// Per-host, because the pane shows one row per host and the outcome
         /// has to render in the row whose button ran it.
         case updateRemotePlugin(hostID: String)
+        case configureHerdrPanel(hostID: String)
     }
 
     public struct EnrollmentConfirmation: Identifiable, Equatable, Sendable {
@@ -313,6 +314,10 @@ public final class ClaudeIntegrationSettingsModel {
     /// dictates and sees no cmux context can read WHY here instead of in the
     /// log. `.ok` renders nothing — a working feature says nothing.
     var cmuxStatus: CmuxSocketStatus = .ok
+
+    /// Set by the join probe after a successful stamp whose value never
+    /// appeared in the focused grid. This is inferential, not a config read.
+    var herdrPanelStatus: HerdrPanelConfigurationStatus = .ok
 
     /// The password field's live text. Never seeded from the Keychain: the
     /// stored secret is not shown back to anyone, and an empty field on a
@@ -933,6 +938,22 @@ public final class ClaudeIntegrationSettingsModel {
         Log.claudeContext.info("Claude remote setup confirmation requested")
     }
 
+    public func requestHerdrPanelConfiguration(hostID: String) {
+        guard !isPerformingEnrollmentAction,
+              let host = hosts.first(where: { $0.id == hostID }),
+              host.sshHostAlias != nil
+        else { return }
+        enrollmentStepStatuses = []
+        enrollmentResultsAction = nil
+        enrollmentConfirmation = EnrollmentConfirmation(
+            action: .configureHerdrPanel(hostID: hostID),
+            title: "Configure this exact herdr agents-panel row?",
+            preview: ClaudeRemoteEnrollmentService.herdrPanelConfigSnippet,
+            confirmButtonTitle: "Confirm Configure"
+        )
+        Log.claudeContext.info("Claude remote herdr panel configuration confirmation requested")
+    }
+
     public func cancelEnrollmentActionConfirmation() {
         enrollmentConfirmation = nil
     }
@@ -944,6 +965,8 @@ public final class ClaudeIntegrationSettingsModel {
             await performPlanAction(confirmation)
         case .updateRemotePlugin:
             await performPluginUpdate(confirmation)
+        case .configureHerdrPanel:
+            await performHerdrPanelConfiguration(confirmation)
         }
     }
 
@@ -969,6 +992,8 @@ public final class ClaudeIntegrationSettingsModel {
             // Routed to performPluginUpdate: that action belongs to a host row,
             // has no plan and no token, and must not run against one.
             return
+        case .configureHerdrPanel:
+            return
         }
         enrollmentConfirmation = nil
         isPerformingEnrollmentAction = true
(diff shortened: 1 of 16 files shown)


T0mSIlver:
Field abstention from the first dogfood dictation (`remoteHerdrPanel: row-not-rendered; settle-timeout` while the stamped row was visibly rendering): the settle loop was capped at 2 reads ~120 ms apart, tighter than a ProxyJump chain's stamp→paint→render round trip — the 0.5 s budget was unreachable. Fixed in 21dde25: the loop now polls every 120 ms for a full 1.0 s budget (hard cap 9 reads so a frozen test clock cannot loop). New tests: `testASlowRemoteRenderStillMatchesWithinTheSettleBudget` (red under the 2-read cap), `testAFrozenClockIsStillBoundedByTheHardReadCap`. Full suite green: `Executed 2655 tests, with 2 tests skipped and 0 failures`.

T0mSIlver:
Second field abstention (`panel candidate has multiple live herdr sockets`): the registry held sessions on two socket paths for the host — typically a session registered under a previous herdr boot still inside the 4 h TTL. Fixed in b94f40c by implementing the recorded follow-up: the panel probe now iterates the host's live sockets (sorted, ≤3), stamping each with its OWN fresh nonce — only the displayed server can render its token, so the nonce disambiguates what the blanket rule had to refuse. A stale socket's forward fails to open and is skipped. First rendered match ends the socket loop (one grid renders one server; the per-host forward service would otherwise tear down the matched forward). The argv fallback keeps its single-socket abstention — it has no way to tell servers apart. Red/green: `testTwoLiveSocketsResolveToTheOneWhoseNonceRenders` red under the committed guard (`Executed 1 test, with 1 failure`), green after; plus `testAStaleSocketWhoseForwardFailsIsSkipped` (the exact field shape). Full suite: `Executed 2724 tests, with 2 tests skipped and 0 failures`.

PR #240, opened 2026-08-29
MicrophoneCaptureService: support multi-channel input devices (pro audio interfaces)

Alexis1537:
## What

A 16-input Universal Audio Thunderbolt interface (default input on a Mac mini with no built-in mic) breaks capture twice over:

1. **The device cannot be opened.** Its native format — 16ch interleaved Float32 @ 48kHz, a channel count with no standard channel layout — makes `AVAudioFormat(streamDescription:)` return nil, so `startCapture` throws `invalidInputFormat` ("Microphone input format is invalid") for a device used successfully elsewhere.
2. **Once opened, dictation hears silence.** `AVAudioConverter` has no downmix matrix from a discrete multi-channel layout to mono, and silently converts every frame to zeros: sessions run, transcripts come back empty.

Fixes, both scoped to `MicrophoneCaptureService`:

- `makeDeviceFormat` falls back to a `DiscreteInOrder` channel layout when the layout-less initializer returns nil (both call sites: capture start and the format-change listener).
- Converter creation is extracted into `makeTranscriptionConverter`, which maps the interface's **first input channel** to the mono output (`channelMap = [0]`) for `channelCount > 2` — that's where a microphone lives on such interfaces. Mono/stereo devices are untouched (standard downmix keeps working, covered by a test).

Both helpers are internal rather than fileprivate so the regression tests can drive them; no behavior change for existing devices.

## Proof

- [x] Unit suite green — `swift test` on this branch:
  ```
  Executed 2749 tests, with 20 tests skipped and 0 failures (0 unexpected) in 100.366 seconds
  ```
- [x] Bug fix: regression test added (`MicrophoneMultiChannelFormatTests`, 4 tests, Metal-free, no audio hardware needed — a synthetic 16ch ASBD and a 440Hz tone on channel 0).

  Failing before the fix (helpers reverted to pre-fix behavior):
  ```
  error: testMakeDeviceFormatWrapsSixteenChannelDevice : XCTAssertNotNil failed - 16ch device format must wrap (discrete layout fallback); nil here is the "Microphone input format is invalid" field failure
  error: testSixteenChannelToMonoConversionPreservesFirstChannelSignal : XCTUnwrap failed: expected non-nil value of type "AVAudioFormat"
  Executed 4 tests, with 4 failures (0 unexpected) in 0.246 seconds
  ```
  Passing after:
  ```
  Executed 4 tests, with 0 failures (0 unexpected) in 0.004 seconds
  ```
- [x] Hand-verified on the affected hardware (UA Thunderbolt, 16 in, macOS 26): stock build → "Microphone input format is invalid"; format fix alone → mic opens, every session finalizes with an empty transcript (silence confirmed by feeding a synthetic single-channel tone through the exact converter: output RMS was 0.0000, and `channelMap = [0]` restores it to full level); both fixes → dictation works end to end, verified against the raw delta log.

Diagnosed and written with Claude Code on the affected machine; happy to test follow-up changes on this hardware.

### DIFF
--- a/Sources/localvoxtral/MicrophoneCaptureService.swift
+++ b/Sources/localvoxtral/MicrophoneCaptureService.swift
@@ -119,7 +119,9 @@ private func auhalInputCallback(
     let inputFormatDescriptor = MicrophoneCaptureService.inputFormatDescriptor(for: inputFormat)
 
     if converterState.converter == nil || converterState.inputFormatDescriptor != inputFormatDescriptor {
-        guard let converter = AVAudioConverter(from: inputFormat, to: context.outputFormat) else {
+        guard let converter = MicrophoneCaptureService.makeTranscriptionConverter(
+            from: inputFormat, to: context.outputFormat
+        ) else {
             if context.debugLoggingEnabled {
                 context.service?.debugLog(
                     "converter creation failed sampleRate=\(inputFormat.sampleRate) channels=\(inputFormat.channelCount)"
@@ -132,7 +134,6 @@ private func auhalInputCallback(
             }
             return noErr
         }
-        converter.sampleRateConverterQuality = AVAudioQuality.max.rawValue
         converterState.converter = converter
         converterState.inputFormatDescriptor = inputFormatDescriptor
         converterState.consecutiveFailureCount = 0
@@ -350,7 +351,7 @@ final class MicrophoneCaptureService: @unchecked Sendable {
         )
 
         if formatStatus == noErr, asbd.mSampleRate > 0,
-           let newFormat = AVAudioFormat(streamDescription: &asbd)
+           let newFormat = Self.makeDeviceFormat(&asbd)
         {
             let ctx = unmanagedCtx.takeUnretainedValue()
             let currentDescriptor = Self.inputFormatDescriptor(for: ctx.deviceAVFormat)
@@ -489,7 +490,7 @@ final class MicrophoneCaptureService: @unchecked Sendable {
                     "get device format", getFormatStatus)
             }
 
-            guard let deviceFormat = AVAudioFormat(streamDescription: &deviceASBD) else {
+            guard let deviceFormat = Self.makeDeviceFormat(&deviceASBD) else {
                 throw MicrophoneCaptureError.invalidInputFormat
             }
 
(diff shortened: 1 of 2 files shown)


T0mSIlver:
Verified this on my side before reviewing the code, and it holds up.

I ran a probe against your two helpers on my Mac, feeding a 440Hz tone on channel 0 of a synthetic device format across a range of channel counts:

```
ch             | plainInit | rms(no map) | rms(map=[0])
1              | ok        | 0.3532      | 0.3532
2              | ok        | 0.3532      | 0.3532
3              | NIL       | 0.0000      | 0.3532
4,5,6,8,10,16  | NIL       | 0.0000      | 0.3532
```

Both failures are real. It also answers the one thing I was worried about, which was whether `channelCount > 2` might change behaviour for 3-8ch devices whose standard downmix works today. It can't: a device stream-format ASBD carries no channel layout, so everything above stereo already lands on the nil path and the silent downmix. Your threshold matches exactly where the discrete fallback engages.

Full unit suite on your branch here: 2728 tests, 0 failures. Fewer than your 2749 because my runner skips the integration suites.

One small thing I'll clean up myself in a follow-up rather than send you round again: the new test file adds 7 strict-concurrency warnings, and the repo rule is that new code stays warning-free under Swift 6.2. The fix is `@preconcurrency import AVFoundation` plus a `Mutex(false)` in place of the captured `var fedInput`, mirroring what `convertToPCM16Mono` already does at `Sources/localvoxtral/MicrophoneCaptureService.swift:885`. Worth knowing for next time, but nothing for you to do here.

Two follow-ups I'd like to pick up separately:

- Channel 0 is a guess. It's right for your interface and strictly better than silence, but on a 16-in unit the mic can sit on any preamp, so an input-channel picker is the real fix.
- `kAudioOutputUnitProperty_ChannelMap` on the AUHAL would avoid rendering 16 channels per callback just to throw 15 away.

I've approved CI. Fork PRs run on GitHub-hosted macOS by design, so the speechd realtime lane self-skips there and green means tier 0 only. Fine for this change, and I ran the rest locally.

Thanks for the hardware report and the offer to test follow-ups. Nobody else here has a multi-channel interface.

T0mSIlver:
That red check is ours, not yours. CI crashed the Swift compiler rather than finding anything wrong with your code:

```
error: emit-module command failed due to signal 6
Assertion failed: (Bits.ApplyExpr.ThrowsIsSet), function throws, file Expr.h, line 5002.
Apple Swift version 6.2.1 (swift-6.2.1-RELEASE)
While silgen emitIsolatingDestructor SIL function "@$s12localvoxtral18DictationViewModelCfD"
 for 'deinit' (at Sources/localvoxtral/DictationViewModel.swift:775:5)
```

Fork PRs run on GitHub-hosted runners, which install a standalone Swift 6.2.1 toolchain. That build has assertions enabled and trips over an isolated `deinit` in a file your PR never touches. Same-repo branches build with Xcode's toolchain, where the assertion is compiled out, so this lane has been quietly broken since 6.2.1 started resolving and you're simply the first fork PR to hit it. A fix is in flight separately.

Merging on the strength of the full unit suite run against your branch here (2728 tests, 0 failures) plus the channel-count probe in my earlier comment.

Alexis1537:
Very happy to have been able to contribute to your excellent app!

PR #224, opened 2026-08-04
SSH enrollment: in-app verification, comment-free copyables, legible sheet

T0mSIlver:
Stacks on #223 (`settings-context-pane`) — merge that first.

## What

Owner direction, 2026-08-04: *"Commands you have to copy-paste have comments, that's just dumb — display it in the app or not at all. Limit the amount of small text in the whole ssh flow, it's tiring to read."*

**1. Copyable text carries zero comments.** `sshConfigSnippet` keeps only its `# BEGIN`/`# END` delimiters — those are functional, the idempotent replace keys on them — and drops the six-line essay about `ExitOnForwardFailure`. `verifyCommands` and `uninstallCommands` leave `SetupPlan` entirely.

```
SetupPlan { sshConfigSnippet, remoteCommands }          // was: + verifyCommands, uninstallCommands, notes
+ VerificationCheck { kind, passed, summary, hint, detail }
+ executeVerification(sshHostAlias:port:timeout:) throws -> [VerificationCheck]
```

**2. Verification happens in the app**, through the same injected `Runner` seam, producing interpreted verdicts instead of nine `#` lines telling the user how to read raw output:

| probe | result | verdict |
|---|---|---|
| unauthenticated `POST /v1/hook/SessionStart` over the forward | `401` | ✓ Tunnel is up and localvoxtral answered |
| | `000` / curl connect failure | ✗ No tunnel is live right now (+ hint: the forward exists only while a session is open) |
| | any other code | ✗ Something else answered on port 8473 |
| | ssh exit ≠ 0 | ✗ Could not reach `<alias>` over SSH |
| `claude plugin list` (existing PATH resolver) | lists `localvoxtral-remote` | ✓ The plugin is installed |
| | does not | ✗ Not installed (+ hint: run step 2) |
| | exit 127 / resolver message | ✗ Claude Code was not found on `<alias>` |

The tunnel probe **deliberately omits `ClearAllForwardings`** (unlike `executeRemoteSetup`): clearing the forward would test a tunnel the probe just disabled. The plugin probe keeps it, for the same reason setup does. The probe script always exits 0 and prints the status code, so a non-zero exit can only mean ssh itself failed — without that, curl's exit 7 and ssh's 255 are the same observation and "no tunnel" gets reported as "cannot reach the host".

[...]

PR #247, opened 2026-08-30
Remote forward: prove who holds a refused port before calling it contention

T0mSIlver:
## What

Settings showed **"Port held — close ssh sessions to that host."** for a host whose
tunnel was working. The log's diagnosis was correct and its conclusion was not:
the port on the remote host was bound by an interactive `sshd-session` of the
owner's, carrying `RemoteForward 29891 127.0.0.1:8473` — the block the app's own
enrollment plan put in `~/.ssh/config`. So the holder was **the forward the
supervisor was trying to create, already established and delivering hooks**, and
the app told the user to close it.

`remote port forwarding failed` says a port is held, never by whom, and the two
holders want opposite things:

* a stranger — an unrelated service, a leftover tunnel, **another Mac's forward** —
  is a real failure the user has to clear;
* our own `RemoteForward`, put there by an ordinary ssh session, is the desired end
  state reached by another route. That is the normal case for anyone who ssh's to
  the host they enrolled, i.e. everybody.

This splits them. Proof gives a new non-failure state `externallyForwarded`
("Tunnel up — an ssh session already holds it.", no Retry button, nothing asked of
the user); everything else stays `portUnavailable`.

### The discriminator, and why it is not the status code

The brief proposed reusing the 401 that an unauthenticated POST gets. Two problems,
and both are load-bearing:

1. **It is not evidence.** The listener's source, response shape and paths are in
   this repository. Any process that binds that port can reproduce our 401 byte for
   byte — including the `WWW-Authenticate: Bearer` and the empty JSON body. Adopting
   it would let a squatter be blessed as a context channel while it collects the
   remote plugin's bearer token from every hook.
2. **It is not even a correct diagnosis.** A SECOND Mac enrolled against the same
   host answers a genuine, honest 401 of its own — which is precisely the contention
   `ClaudeRemoteForwardPort` exists to surface (issue #215). A status-code probe
   adopts that too.

So the probe never reads the answer. It posts a fresh 128-bit nonce **into** the
disputed port from the remote host and asks whether **this process's own listener
saw it arrive**:

* our forward holds the port → the request rides the tunnel and lands here — match;
* a stranger holds the port → it receives the nonce and can do nothing with it. The
  listener binds `127.0.0.1` on the Mac, and the only route to it from that host is
  the forward the stranger is the reason we do not have;
* another Mac holds the port → the nonce lands on *that* Mac's listener, which has
  never heard of it. No match here, which is the right answer.

### Fail-closed

`.unproved` is the default, not a branch: no probe wired in (the seam is nil — which
is what the herdr `-L` supervisor gets), no `curl` on the host, ssh refused, a
timeout, an unparseable anything — all keep the pre-existing `portUnavailable`. The
probe's **exit status is deliberately never consulted** in either direction; only
arrival counts.

The one pre-auth look in the listener is not an oracle. It compares only a value this
process minted and is still waiting for (constant-time, one-shot, bounded set), it
returns the same 401 with the same headers and the same empty body either way, and a
match short-circuits — so a matching nonce cannot carry a payload past authentication
even with a valid token riding along. It runs before the token solely so a self-probe
does not land in the rejection tally the user reads for "which of three fixes do I
need".

The probe connection carries no credential (`ssh -o BatchMode=yes -o
ClearAllForwardings=yes -o ConnectTimeout=5 -- <alias> /bin/sh -s`, 10 s cap), and the
nonce goes down **stdin**, never argv — an argv is world-readable on the remote host.
`ClearAllForwardings=yes` matters here specifically: without it this connection
inherits the alias's own `RemoteForward` and contends for the very port it is asking
about. (That is the opposite of the herdr `-L`, where the same option deletes the
`-L` — `docs/agent/invariants.md` covers both.)

### Lifecycle

`externallyForwarded` is a claim with an expiry. The holder is a session the app does
not own and cannot be notified about, so the supervisor re-attempts its own `-R` on a
long injected-clock park (5 min): the session ending means the bind now succeeds and
the app **takes the tunnel over**; a holder that stops proving ownership drops back to
`portUnavailable`. Both directions are pinned by tests. This is the one relaxation of
the type's "a refused bind is terminal" rule and it is bounded by that interval — two
connections per 5 minutes against the user's own host, not a reconnect storm. The
re-check deliberately does not blink the row through "Connecting…".

### Copy

| state | before | after |
| --- | --- | --- |
| our own session holds it | `Port held — close ssh sessions to that host.` (+ a Retry that could only fail) | `Tunnel up — an ssh session already holds it.` (not a failure; no Retry) |
| anything else holds it | same string | `Port held by another program on that host.` |

The old advice is now wrong twice over: in the common case it names the working
tunnel, and in what is left the holder was measurably *not* an ssh session of ours.
Both are one short sentence (owner popover rule, asserted).

### Not fixed here, and not a bug

"Last context: 20 days ago" beside a live tunnel is honest — `lastSeenAt` is the last
hook this Mac *ingested*, and the tunnel merely answers on demand.

### One correction to the report

`ClaudeRemoteEnrollmentService.executeVerification` does not exist — #224 shipped
`verifyCommands`, the copy-pasteable text a person runs, which already documents the
three-outcome reading. There was no in-app probe to reuse; the runner seam
(`ClaudeRemoteEnrollmentService.Runner` / `processRunner()`) is reused instead.

## Proof

- [x] **Unit suite green** — `./scripts/remote-build.sh test`

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-08-30 14:49:54.305.
	 Executed 2765 tests, with 2 tests skipped and 0 failures (0 unexpected) in 104.097 (104.235) seconds
```

- [x] **Bug fix: regression tests added — failing before, passing after.**

"Before" is this tree with the two behavioural decisions reverted and the new types
kept, so the tests compile and measure exactly the fix: `resolveOwnership()` returns
`.unproved` unconditionally (the old "a refusal is always contention"), and the
listener's pre-auth nonce branch is deleted.

**BEFORE — `./scripts/remote-build.sh test --filter ClaudeRemoteForwardSupervisorTests`**

```
Test Case '-[...testARefusedBindThatIsOurOwnLiveForwardIsNotContention]' failed (20.047 seconds).
Test Case '-[...testARefusedBindWithUnprovedOwnershipStaysContention]' failed (0.003 seconds).
Test Case '-[...testAnExternallyHeldForwardIsTakenOverOnceTheSessionEnds]' failed (20.247 seconds).
Test Case '-[...testAnExternallyHeldForwardStopsClaimingTheChannelWhenTheProofFails]' failed (20.008 seconds).
	 Executed 26 tests, with 7 failures (3 unexpected) in 60.339 (60.341) seconds

Tests/localvoxtralTests/ClaudeRemoteForwardSupervisorTests.swift:494: error: -[...testARefusedBindThatIsOurOwnLiveForwardIsNotContention] : failed - timed out waiting for a state; saw [....State.connecting, ....State.portUnavailable]
Tests/localvoxtralTests/ClaudeRemoteForwardSupervisorTests.swift:540: error: -[...testARefusedBindWithUnprovedOwnershipStaysContention] : XCTAssertEqual failed: ("0") is not equal to ("1")
Tests/localvoxtralTests/ClaudeRemoteForwardSupervisorTests.swift:561: error: -[...testAnExternallyHeldForwardIsTakenOverOnceTheSessionEnds] : failed - timed out waiting for a state; saw [....State.connecting, ....State.portUnavailable]
Tests/localvoxtralTests/ClaudeRemoteForwardSupervisorTests.swift:591: error: -[...testAnExternallyHeldForwardStopsClaimingTheChannelWhenTheProofFails] : failed - timed out waiting for a state; saw [....State.connecting, ....State.portUnavailable]
```

`saw [.connecting, .portUnavailable]` **is the field bug**, reproduced in a test.

**BEFORE — `./scripts/remote-build.sh test --filter ClaudeRemoteContextListenerTests`**

```
Test Case '-[...testAMatchingNonceCannotCarryAPayloadPastAuthentication]' failed (0.013 seconds).
Test Case '-[...testAnOwnershipProbeArrivesWithoutBeingCountedAsARejection]' failed (0.001 seconds).
	 Executed 44 tests, with 4 failures (0 unexpected) in 0.043 (0.045) seconds

Tests/.../ClaudeRemoteContextListenerTests.swift:193: error: -[...testAnOwnershipProbeArrivesWithoutBeingCountedAsARejection] : XCTAssertTrue failed - arrival IS the proof; nothing else is read
Tests/.../ClaudeRemoteContextListenerTests.swift:194: error: -[...testAnOwnershipProbeArrivesWithoutBeingCountedAsARejection] : XCTAssertTrue failed - our own probe must not look like a host that needs re-enrolling
Tests/.../ClaudeRemoteContextListenerTests.swift:239: error: -[...testAMatchingNonceCannotCarryAPayloadPastAuthentication] : XCTAssertEqual failed: ("200") is not equal to ("401")
Tests/.../ClaudeRemoteContextListenerTests.swift:240: error: -[...testAMatchingNonceCannotCarryAPayloadPastAuthentication] : XCTAssertTrue failed - no record, no marker, no session
```

**AFTER — same filters, this tree**

```
Test Suite 'ClaudeRemoteForwardSupervisorTests'
	 Executed 26 tests, with 0 failures (0 unexpected) in 0.031 (0.032) seconds

Test Suite 'ClaudeRemoteContextListenerTests'
	 Executed 44 tests, with 0 failures (0 unexpected) in 0.033 (0.035) seconds

Test Suite 'ClaudeRemoteForwardOwnershipTests' passed at 2026-08-30 14:47:26.230.
	 Executed 12 tests, with 0 failures (0 unexpected) in 0.004 (0.005) seconds
```

The fail-closed cases are their own tests, not comments:
`testAHostThatAnswersButCannotDeliverTheNonceIsNotOurs` (a squatter answering
perfectly is still `.unproved`), `testAProbeHeaderWeNeverMintedIsJustAnUnauthenticatedRequest`,
`testAMatchingNonceCannotCarryAPayloadPastAuthentication`,
`testTheNonceTravelsOnStdinAndNeverInArgv`, `testAnInvalidAliasNeverReachesAProcess`.

- [x] **`integration-herdr` green** — `./scripts/remote-build.sh integration-herdr`

```
Test Suite 'HerdrIntegrationTests' passed at 2026-08-30 14:53:56.204.
	 Executed 11 tests, with 0 failures (0 unexpected) in 60.725 (60.726) seconds
```

Run rather than justified away. Strictly, nothing about how the herdr `-L` is opened
changed — `ClaudeRemoteHerdrForwardService.argv` is untouched, the ownership seam is
nil on that path by construction, and `externallyForwarded` is unreachable for a `-L`
— but the supervisor's lifecycle is shared code, so the lane is the cheap answer.

- [x] **No wall clock added.** The re-check park is the existing injected `sleepFor`
  seam; the tests assert `.seconds(300)` was requested rather than waiting for it.
  Nothing here reaches `beginDictationSession`.

- [ ] UI change verified by hand — **not done.** The only UI surface is the
  per-host status string and the presence of the Retry button, both driven by
  `State.text` / `State.isFailure` and both asserted (supervisor suite +
  `ClaudeIntegrationSettingsModelTests`). The field state that motivates this needs a
  real contended port on a real host to reproduce.

## Files

* `ClaudeRemoteForwardOwnership.swift` — new: ownership verdict, nonce witness, probe.
* `ClaudeRemoteForwardSupervisor.swift` — `externallyForwarded`, the probe seam, the
  bounded re-check, new copy.
* `ClaudeRemoteContextListener.swift` — the pre-auth nonce observation.
* `ClaudeRemoteForwardCoordinator.swift` / `ClaudeRemoteListenerCoordinator.swift` /
  `localvoxtralApp.swift` — one shared witness between listener and supervisors.
* `docs/agent/invariants.md`, `docs/architecture.md` — the trust argument and the
  lifecycle bound, where the next person will look for them.

## Review (adversarial, GLM 5.2 via opencode — release train)

One read-only adversarial pass over the branch with `docs/agent/invariants.md`
and the full diff. Verdict: **three real findings, all fixed in `3db0940`;
three rejected with the argument below.** `origin/main` (through #246) was
merged into the branch first — clean, no conflicts.

### Fixed

**1. The invariants claim was absolute and is not true.** The doc said the only
route from that host to our listener is "the forward the stranger is the reason
we do not have". Every supervised `-R` ends at the SAME local listener
(`-R <derived>:127.0.0.1:<listenerPort>`, `ClaudeRemoteForwardSupervisor.swift:329`),
so an arrival identifies the LISTENER, not the remote port that carried the
nonce. Where a host has a second live forward to this Mac — a legacy
`RemoteForward 8473` in the user's own config block, carried by an interactive
session; the cohort `staleConfiguredForward` exists for — a hostile holder of
the disputed port can read our request off its own socket and replay it down
that other forward, and the witness matches. Both the doc and the
`ClaudeRemoteForwardProbeWitness` comment now state the residual, and say that
`.ourListener` is a diagnosis and never an authorization.

The reviewer graded this a merge blocker; I did not, and shipped it as a stated
residual instead. The replay needs code execution on the enrolled host, and
anything with code execution there **already holds the plugin's bearer token**
(it is in that host's plugin config) and already receives every hook. It gains
a suppressed status row, not a credential — so this is a diagnostic that can be
fooled by an adversary who has already won, not a control that can be bypassed.
What was genuinely wrong was writing the absolute form into the one file agents
trust; that is what is fixed.

**2. "nonce on stdin so it is never in a remote argv" is false.** stdin keeps
the nonce out of the **ssh** argv on this Mac — load-bearing, since this
machine is also the self-hosted CI runner — but the remote `sh` then runs
`curl` with `-H '<header>: <nonce>'` as a literal argument, readable in that
host's process list for the `--max-time` window. `--header @file` is the fix
(the plugin shim uses exactly that) and is deliberately **not** adopted: it
needs curl >= 7.55 on an arbitrary enrolled host, and below that the probe
would stop producing arrivals with no signal at all — a silent degrade to "not
ours" is worse than the exposure, which on its own buys a reader nothing
(spending the nonce still needs a live forward to this listener). Claim
corrected; `testTheNonceTravelsOnStdinAndNeverInArgv` renamed to
`...NeverInTheSSHArgv` and given an assertion that pins the remote-argv
exposure, so changing it is a decision and not an unnoticed diff.

**3. Two claimed bounds had no assertion.** `ConnectTimeout=5` and the 10 s
outer cap appeared in no test — dropping either shipped green while a
black-holed route parked the main-actor-awaited supervise loop for the full
budget. Same for "the re-check deliberately does not blink the row through
Connecting…". Both are now asserted, and both assertions were shown to bite
(red/green below). `arm()` also logs an eviction at the `maximumArmed` cap:
it silently turns another host's probe into `.unproved`, after which the
supervisor logs "bound by something that is not this Mac's listener" — a
negative that probe never established.

### Rejected

- **"A late-arriving probe is counted as a real `missingToken` rejection"**
  (millisecond window where `consume` disarms before the request is parsed).
  Real, and its stated harm does not survive the next commit on this train:
  the probe sends no `Authorization` header, so under #248 it lands in
  `.absentAuthorization`, which `isFault` is false for, which logs at `.notice`
  and which `Snapshot.isEmpty` excludes — it raises no hint and appears in no
  "which of three fixes" surface. Not worth a grace list.
- **"No test runs the composed path (script -> real ssh -> curl -> tunnel ->
  listener -> witness -> supervisor)."** Accurate, and already stated in the
  body's unchecked box: reproducing it needs a genuinely contended port on a
  real enrolled host. `integration-herdr` cannot cover it (the ownership seam
  is nil on the `-L` path by construction). Named, not papered over.
- **The `~100-col` comment reflow nit** in `ClaudeRemoteForwardOrphanReaper` —
  cosmetic, and the file has no enforced column rule.

### Proof for the review fixes

**RED** — `ConnectTimeout=5` deleted from `argv` and `transition(to: .connecting)`
re-added before the re-check park, i.e. exactly the two properties removed:

```
$ ./scripts/remote-build.sh test --filter ClaudeRemoteForwardOwnershipTests
ClaudeRemoteForwardOwnershipTests.swift:282: error: -[...testTheProbeIsBoundedAtBothTheConnectAndTheWholeRun] : XCTAssertTrue failed - a host that black-holes the port must not eat the whole budget in connect(): ["ssh", "-o", "BatchMode=yes", "-o", "ClearAllForwardings=yes", "--", "builder", "/bin/sh", "-s"]
	 Executed 13 tests, with 1 failure (0 unexpected) in 0.191 (0.191) seconds

$ ./scripts/remote-build.sh test --filter ClaudeRemoteForwardSupervisorTests
ClaudeRemoteForwardSupervisorTests.swift:580: error: -[...testAnExternallyHeldForwardIsTakenOverOnceTheSessionEnds] : XCTAssertFalse failed - the re-check re-dials silently; states after the proof were [....State.connecting, ....State.forwarding]
ClaudeRemoteForwardSupervisorTests.swift:498: error: -[...testARefusedBindThatIsOurOwnLiveForwardIsNotContention] : XCTAssertEqual failed: ("connecting") is not equal to ("externallyForwarded")
	 Executed 26 tests, with 2 failures (0 unexpected) in 0.073 (0.074) seconds
```

**GREEN** — both restored, same filters:

```
$ ./scripts/remote-build.sh test --filter ClaudeRemoteForwardOwnershipTests
	 Executed 13 tests, with 0 failures (0 unexpected) in 0.004 (0.005) seconds

$ ./scripts/remote-build.sh test --filter ClaudeRemoteForwardSupervisorTests
	 Executed 26 tests, with 0 failures (0 unexpected) in 0.026 (0.027) seconds
```

Findings 1 and 2 are documentation-accuracy fixes and change no behaviour, so
they carry no red/green — the code they describe is unchanged.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteContextListener.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteContextListener.swift
@@ -434,6 +434,7 @@ public final class ClaudeRemoteContextListener: Sendable {
         // re-running its enrollment (field report, 2026-07-26). The category is
         // derived from the header's shape and the authentication result only —
         // no token material reaches the log, the tally, or the UI.
+
         // The ONE thing looked at before the token, and the exception proves
         // the rule above rather than weakening it: the value compared is a
         // random nonce THIS process minted moments ago and is still waiting
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteContextListener.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteContextListener.swift
@@ -103,6 +103,11 @@ public final class ClaudeRemoteContextListener: Sendable {
     /// listener on every rebind, and evidence the user has not read yet must not
     /// be erased by enrolling a second host.
     private let rejections: ClaudeRemoteRejectionTally
+    /// Nonces the forward supervisor minted for an in-flight ownership probe.
+    /// Injected for the same reason the tally is: the coordinator rebuilds the
+    /// listener on every rebind, and a probe in flight across one must not be
+    /// lost. See `ClaudeRemoteForwardProbeWitness` for why this exists at all.
+    private let forwardProbes: ClaudeRemoteForwardProbeWitness
     /// Authenticated host activity can pre-start a slow herdr `-L` away from
     /// dictation latency. The path remains an opaque remote label.
     private let onRemoteHerdrActivity: @Sendable (String, String) -> Void
@@ -143,6 +148,7 @@ public final class ClaudeRemoteContextListener: Sendable {
         hosts: ClaudeRemoteHostRegistry,
         limits: ClaudeRemoteListenerLimits = .default,
         rejections: ClaudeRemoteRejectionTally = ClaudeRemoteRejectionTally(),
+        forwardProbes: ClaudeRemoteForwardProbeWitness = ClaudeRemoteForwardProbeWitness(),
         now: @escaping @Sendable () -> Date = { Date() },
         uptimeNanos: @escaping @Sendable () -> UInt64 = { DispatchTime.now().uptimeNanoseconds },
         onRemoteHerdrActivity: @escaping @Sendable (String, String) -> Void = { _, _ in }
@@ -151,6 +157,7 @@ public final class ClaudeRemoteContextListener: Sendable {
         self.hosts = hosts
         self.limits = limits
         self.rejections = rejections
+        self.forwardProbes = forwardProbes
         self.now = now
         self.uptimeNanos = uptimeNanos
         self.onRemoteHerdrActivity = onRemoteHerdrActivity
@@ -427,6 +434,25 @@ public final class ClaudeRemoteContextListener: Sendable {
         // re-running its enrollment (field report, 2026-07-26). The category is
         // derived from the header's shape and the authentication result only —
         // no token material reaches the log, the tally, or the UI.
+        // The ONE thing looked at before the token, and the exception proves
+        // the rule above rather than weakening it: the value compared is a
+        // random nonce THIS process minted moments ago and is still waiting
+        // for, so a peer that did not receive it cannot produce a match, and a
+        // peer that did learns nothing from having produced one. The response
+        // is byte-for-byte the same 401 either way — this branch exists to keep
+        // our own probe out of the rejection tally, whose whole job is telling
+        // the user which of three real fixes to apply. Nothing is ingested,
+        // nothing is authorized: the arrival itself is the entire signal.
+        if forwardProbes.note(
+            headerValue: request.headers[ClaudeRemoteForwardProbeWitness.headerName]
+        ) {
+            Log.claudeContext.info(
+                "Claude remote forward ownership probe arrived on this listener"
+            )
+            respond(fd: fd, status: 401)
+            return
+        }
+
         let shape = ClaudeRemoteHTTPCodec.authorizationShape(
             in: request.headers["authorization"], limits: limits.http
         )
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardCoordinator.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardCoordinator.swift
@@ -64,6 +64,9 @@ public final class ClaudeRemoteForwardCoordinator {
         listenerPort: UInt16 = ClaudeRemoteListenerLimits.default.port,
         isListenerBound: @escaping @MainActor () -> Bool,
         makeSupervisor: MakeSupervisor? = nil,
+        // Handed to every supervisor this coordinator builds itself. Nil keeps
+        // the fail-closed behaviour: a refused bind stays `portUnavailable`.
+        ownershipProbe: ClaudeRemoteForwardOwnershipProbe? = nil,
         pidLedger: ClaudeRemoteForwardPidLedger? = nil,
         reapOrphans: (@Sendable () async -> Void)? = nil,
         reconcileHerdrEnrollment: @escaping @MainActor (Set<String>) -> Void = { _ in }
@@ -94,7 +97,8 @@ public final class ClaudeRemoteForwardCoordinator {
                         pidLedger.remember(hostID: config.hostID, record: record)
                     }
                     return process
-                }
+                },
+                ownershipProbe: ownershipProbe
             )
         }
     }
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOrphanReaper.swift
@@ -8,8 +8,7 @@ import Darwin
 /// run starts hook `-R` or herdr `-L` forwards.
 ///
 /// The bug this closes (field report, 2026-08-05): quit-and-reopen sometimes
-/// landed the pane on "Port held — close ssh sessions to that host." with a
-/// Retry that could only fail. The holder was this Mac's own orphan — an
+/// landed the pane on the port-held status with a Retry that could only fail. The holder was this Mac's own orphan — an
 /// `ssh -N -R` from a run that ended without `applicationWillTerminate`
 /// (crash, force-quit) or whose teardown outran the bounded quit drain. It
 /// reparents to launchd, keepalives keep it healthy forever, and nothing else
--- a/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOwnership.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeRemoteForwardOwnership.swift
@@ -1,0 +1,235 @@
+import Foundation
+import Synchronization
+
+/// Who holds the remote listen port when OpenSSH refuses this Mac's bind.
+///
+/// The refusal itself cannot tell the two apart, and they want opposite things
+/// from the user (field report, 2026-08-29):
+///
+/// * a stranger — another Mac's forward, a leftover tunnel, an unrelated
+///   service — is a real failure the user has to go and clear;
+/// * **our own `RemoteForward`, already established by an ordinary ssh session
+///   that read the enrollment block out of `~/.ssh/config`**, is the desired
+///   end state reached by another route. It is what happens to everybody who
+///   actually ssh's to the host they enrolled, and the app was telling them to
+///   close the session providing the working tunnel.
+public enum ClaudeRemoteForwardOwnership: Sendable, Equatable {
+    /// A nonce this process minted seconds ago came back in on this Mac's own
+    /// listener, through the port in question. Nothing else can produce that.
+    case ourListener
+    /// Everything else, including every way the check could not be run.
+    /// The only safe default: see `ClaudeRemoteForwardProbeWitness`.
+    case unproved
+}
+
+/// Nonces minted for an in-flight ownership probe, and which of them the
+/// listener has SEEN arrive.
+///
+/// **This is the whole trust argument, so it is worth being explicit about what
+/// is and is not evidence.**
+///
+/// The obvious probe — curl the port from the remote host and read the status
+/// code — is not evidence of anything. Our 401 (and the 411 an empty POST gets)
+/// is public: the listener's source, its response shape and its paths are all in
+/// this repository, so any process on the remote host can bind the port and
+/// reproduce those bytes exactly. Worse, it is not even a correct *diagnosis*:
+/// a SECOND Mac enrolled against the same host answers a genuine, honest 401 of
+/// its own, and that is precisely the contention `ClaudeRemoteForwardPort`
+/// exists to make visible.
+///
+/// So the probe does not read the answer at all. It sends a fresh random nonce
+/// TO the port and then asks THIS process whether its own loopback listener saw
+/// that nonce. That inverts the question into one only the real topology can
+/// answer:
+///
+/// * if the port is held by our own `RemoteForward`, the request travels the
+///   tunnel and lands here — match;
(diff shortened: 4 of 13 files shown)


T0mSIlver:
### CI: green on run 3; the two earlier reds were the herdr lane's host state, not this change

`33313803329` — **all checks passed** (8m20s), herdr lane included:

```
Test (unit suite)                                        -> success
Integration tests (live STT service)                     -> success
Live herdr integration (real server + real ssh -L forward) -> success
```

Recording why the first two runs were red, because a reader landing on this PR
later should not have to re-derive it:

| run | herdr lane |
|---|---|
| `33312804025` (first) | 1 failure: `testAttachClientRendersNoSidebarSoItCannotEchoThePanelToken` — `timed out waiting for the surface to paint lv-mic-pijqw2g4rp` |
| `33312804025` (rerun) | 10 of 11, cascading: `stampRefused`, then `pane_not_found: pane w1:p1 not found` |
| `33313803329` | 11/11 green |

Three things say this is the lane's own account state and not the diff:

1. **Not deterministic.** A code fault fails the same way twice; this went from
   one paint timeout to a whole-fixture collapse, then to green with no change to
   any file the lane touches.
2. **`pane_not_found` is the fixture's own documented signature for a herdr
   server that never got its config.** From `scripts/herdr-integration-fixture.sh`:
   *"`onboarding = false` matters: herdr's first-run setup screen has no workspace
   and therefore no pane, so `pane.current` answers pane_not_found forever."* The
   lane "takes over this account's herdr config … for the duration of the run"
   (ci.yml), so a run killed mid-flight is exactly how the next one sees that.
3. **The one herdr test that exercises the code this PR touches passed in every
   red run**: `testForwardLeaseIsReusedAcrossDictationsAndTornDownWhenIdle` — the
   supervised `ssh -L` lease. Everything that failed was pty painting and
   `pane.report_metadata` stamping, which this diff does not reach.

Independently, the lane passed **twice** on the build host against this exact
tree, before and after the CI reds:

```
Executed 11 tests, with 0 failures (0 unexpected) in 60.685 (60.686) seconds
```

Nothing here asks for the lane to be relaxed — it caught real host state and said
so loudly, which is what it is for.

Issue #287, opened 2026-09-08
integration-herdr: cover herdr 0.9 federation before any join code trusts a machine list


T0mSIlver:
Planned soon, and it comes before the join work in the federated arm issue.

## Why

Everything localvoxtral is about to rely on in herdr 0.9.0 was read from source at tag `v0.9.0`, and nothing has been measured against a running 0.9 server. The `integration-herdr` lane exists to hold exactly this class of assumption to a live server, so it has to cover federation before any join code trusts a machine list.

Two changes depend on it: the fail-closed guard in #286, which reads herdr's machine state, and the federated join arm, which would name its target machine from that same state.

The fixture pins no version (`scripts/herdr-integration-fixture.sh:95` uses whichever binary is installed), so the lane may already be running against 0.9 while asserting nothing about machines.

## What the lane should assert, each as its own named test

- `herdr machine list --json` reports `selected` for the machine the client is currently viewing, and no selection when it is showing Local. Source: `src/cli/machine.rs:262`, over `<state_dir>/client/endpoints.json` and `client/endpoint-selection.json`, written on every switch at `src/client/mod.rs:1192`.
- While a remote machine is selected, the local server still answers `pane.current` with its own focused pane. This is the shape behind #286, and the lane should fail loudly if herdr ever stops behaving this way.
- The client's agents panel renders rows for panes on every federated machine at once, so an `lvmark` stamped on machine B renders while machine A is displayed (`src/client/shell/endpoint_agents.rs`). This is what retires the whole-view App client discriminator for 0.9 clients.
- The panel row comes from the local client's config: with `[ui.sidebar.agents]` configured locally and absent on the remote, the token still renders (`ClientShellConfig::from_config`).
- An observer of a stamped pane renders no token. `herdr terminal session observe <target>` exists and has since v0.8.2 (`src/cli.rs:43`), so this sits next to the `terminal attach` case the lane already asserts, and the panel-binding doc's "documented hope" about `terminal_observe` goes away.

## Fixture shape

`scripts/herdr-integration-fixture.sh` already starts a live server, a whole-view client on a real pty, and an `ssh -L`. The federated arm adds a local 0.9 client, one `herdr machine add <fixture ssh target> --label ... --remote-session ...`, and switching the selection between Local and that machine. `machine add` needs herdr present on the remote and non-interactive ssh, both of which the fixture already arranges.

[...]

Issue #288, opened 2026-09-08
Dictate into any machine herdr federates: import herdr's machine list and add a federated join arm


T0mSIlver:
Planned soon, after the guard in #286 and the lane in #287.

## What herdr 0.9.0 makes possible

A 0.9 client attaches several SSH machines at once and shows their panes in one local UI (`herdr machine add`, `src/cli/machine.rs`). The user no longer opens an ssh session to reach a remote agent, so the remote-herdr arm's whole basis, an ssh on the surface TTY, disappears from the setup it was built for. What replaces it is better evidence than anything we infer today: herdr itself records which machine the user is viewing.

## Part A: machines as the enrollment source

`herdr machine list --json` returns each machine's label, ssh target and explicit session name, needing no running server (`src/cli/machine.rs:262`). Every one of those targets has already been proven to work non-interactively, because `machine add` runs remote setup before saving.

Settings can offer to import them instead of asking the user to type ssh destinations, one row per machine with the readiness dots the sidebar already uses (#284). The rest of enrollment does not change: each host still needs the plugin installed and the app-managed reverse tunnel to the context listener. Only identification gets easier.

## Part B: a `.federatedHerdrPane` arm

1. Bind the surface to a local herdr client with the existing `HerdrClientTTYProbe`.
2. Name the machine from herdr's own selection state, rather than inferring a host from the process table.
3. Ask that machine's JSON API for the focused pane, over the `ssh -L` forward we already supervise.
4. Keep the panel nonce, demoted from primary authorization to a freshness confirmation and the visible mic indicator.

For a federated user this removes the ssh argv classification (`HerdrInvocation`), `SSHDestinationTTYProbe`, the competing-view scan, the bounded speculative host loop, and the cold-forward latency and stray nonces that loop pays for. It also removes the case the panel-binding doc opens with, where Ghostty's shell-integration ssh wrapper makes the argv unreadable, because there is no ssh on the surface to read.

## Constraints the design has to respect

- The `ssh -L` forward stays. `herdr remote-client-bridge` carries the client protocol to the remote's `herdr-client.sock` and serves one connection at a time (`src/remote/saved.rs:15`, `src/remote/host_unix.rs:8`), so it is not a route to the JSON API socket.
- Abstain when more than one local herdr client is running. The selection file is global and the last client to switch wins (`src/client/endpoint/catalog.rs:105`).
[...]

PR #290, opened 2026-09-08
Local herdr join: abstain while a 0.9 client is showing a federated machine

### LINKED ISSUE #286: Local herdr join can ground a dictation in an unviewed pane once herdr 0.9 federates SSH machines
## What breaks

herdr 0.9.0 lets one local client show panes from several machines at once (`herdr machine add`). When the user selects a remote machine, the local herdr server keeps its own focused pane and only stops presenting it.

`resolveViaHerdr` (`Sources/localvoxtral/ClaudeContext/TerminalScreenClaudeJoin.swift:558`) still asks the local socket for `pane.current` and joins whatever live session sits in that pane. Every cross-check passes, because that session really is in that pane: herdr's `agent_session` claim agrees and the registered pid is in the pane's foreground list. The dictation then carries the wrong session's context, while the keystrokes land in the remote pane the user is looking at.

The failure direction is a wrong join rather than an abstention, which is what `docs/agent/invariants.md` forbids for this arm.

The correct join is unreachable in the same setup. `.remoteHerdrPane` requires an ssh process on the surface TTY, and a federated client has none, so that arm abstains by construction.

## Evidence

Read from herdr source at tag `v0.9.0` (checkout `../herdr`). Not yet measured against a running 0.9 server.

- `src/cli/machine.rs` adds `herdr machine add|list|rename|enable|disable|remove`. Profiles (opaque id, label, ssh target, explicit remote session, enabled) live in `state_dir()/client/endpoints.json`; the machine the user is currently viewing lives in `state_dir()/client/endpoint-selection.json` and is rewritten on every switch (`src/client/mod.rs:1192`).
- Federation is client side. Each saved machine gets its own connection bridged over `ssh -T <target> herdr remote-client-bridge` (`src/remote/saved.rs:15`). The local server never learns about remote panes, and each remote server answers only for its own.
- `herdr machine list --json` prints id, label, target, session, enabled and selected. It reads those two files and needs no running server (`src/cli/machine.rs:262`).

`scripts/herdr-integration-fixture.sh:95` uses whichever herdr binary is installed, so the `integration-herdr` lane may already run against 0.9 while asserting nothing about machines.

## Proposed guard, fail closed

Before `resolveViaHerdr` queries the socket, read herdr's saved machine state and abstain unless Local is selected. A missing `endpoints.json` means Local, so setups without saved machines behave exactly as they do today.

Abstain as well when more than one local herdr client is running. The selection file is global and the last client to switch wins, and `src/client/endpoint/catalog.rs:105` says each live client keeps its own selection independent of the others.

Both abstentions name their cause in `Log.claudeContext`, like the ones already in that function.

## Regression test, tier 0

An injected machine-state reader that reports a selected remote machine must make the arm abstain, in a case where the pane query and the registry would otherwise resolve a join. A second case covers two live local herdr clients. Show both failing before the guard and passing after.

T0mSIlver:
Closes #286.

## What

herdr 0.9.0 attaches several SSH machines to one client. While a remote machine is selected, the local herdr server keeps a focused pane it has merely stopped presenting, so `pane.current` names a pane nobody is looking at. The local herdr arm joined it anyway, with every cross-check passing on it, which made the failure a wrong join rather than an abstention.

`resolveViaHerdr` now reads herdr's saved machine state before it asks the socket anything:

- Showing a federated machine, or state that exists and cannot be read, abstains.
- With machines saved it also requires a lone herdr client surface. herdr keeps one selection per user rather than one per client (`src/client/endpoint/catalog.rs:105`), so a second client on screen makes that selection unable to say which machine the focused surface shows.
- With no machines saved nothing changes, second herdr window included: the surface count is only consulted once a catalog exists.

`HerdrMachineFederationReader` reads the same two files `herdr machine list` reads, `<state dir>/client/endpoints.json` and `client/endpoint-selection.json`, and resolves the selection the way herdr does: the selection file wins when it names an enabled profile, the catalog's own copy answers when it does not, anything else is Local. Reading files rather than running `herdr machine list --json` is deliberate: federation lives in herdr's client, no server knows about it, and a GUI app cannot reliably locate the herdr binary.

The residual is in the doc and fails open: herdr's state directory moves with `XDG_STATE_HOME`, which a GUI app does not see in the user's shell, so such a user reads back "no machines saved" and keeps the unguarded behavior.

## Proof

```
Test Suite 'localvoxtralPackageTests.xctest' passed at 2026-09-08 23:57:11.950.
	 Executed 3147 tests, with 2 tests skipped and 0 failures (0 unexpected) in 52.396 (52.528) seconds
```

- [x] Bug fix: regression test added — failing before the fix, passing after.

Before, with the guard block removed from `resolveViaHerdr` and everything else identical (`./scripts/remote-build.sh test --filter TerminalScreenClaudeJoinTests`):

```
Test Case '-[localvoxtralTests.TerminalScreenClaudeJoinTests testHerdrJoinAbstainsWhenTheMachineStateIsUnreadable]' failed (0.205 seconds).
Test Case '-[localvoxtralTests.TerminalScreenClaudeJoinTests testHerdrJoinAbstainsWhenTheSurfaceCountIsUnknown]' failed (0.001 seconds).
Test Case '-[localvoxtralTests.TerminalScreenClaudeJoinTests testHerdrJoinAbstainsWhileTheClientShowsAFederatedMachine]' failed (0.001 seconds).
Test Case '-[localvoxtralTests.TerminalScreenClaudeJoinTests testHerdrJoinAbstainsWithMachinesSavedAndASecondHerdrSurface]' failed (0.000 seconds).
	 Executed 53 tests, with 4 failures (0 unexpected) in 0.223 (0.224) seconds
```

After, same filter:

```
	 Executed 53 tests, with 0 failures (0 unexpected) in 0.016 (0.017) seconds
```

The two cases that must still JOIN (`testHerdrJoinsWhenMachinesAreSavedAndTheClientShowsLocal`, `testASecondHerdrSurfaceIsIrrelevantWithoutSavedMachines`) pass in both runs, so the four failures above are the guard and not a broken arm.

No UI change. The `integration-herdr` lane is untouched by this diff; #287 covers holding these herdr 0.9 facts to a live server.

### DIFF
--- a/Sources/localvoxtral/ClaudeContext/ClaudeSurfaceProbeCommand.swift
+++ b/Sources/localvoxtral/ClaudeContext/ClaudeSurfaceProbeCommand.swift
@@ -107,6 +107,8 @@ enum ClaudeSurfaceProbeCommand {
             registry: registry,
             focusedTerminalTTY: { await ttyReader.focusedTerminalTTY(bundleID: $0) },
             herdrClientProbe: { HerdrClientTTYProbe.isHerdrClient(onTTYDevicePath: $0) },
+            herdrFederation: { HerdrMachineFederationReader.live().federation() },
+            herdrClientSurfaceCount: { HerdrClientTTYProbe.clientSurfaceCount() },
             herdrPanes: HerdrSocketClient(),
             sshDestinationProbe: { SSHDestinationTTYProbe.connection(onTTYDevicePath: $0) },
             enrolledHosts: { hosts?.hosts(matchingSSHDestination: $0) ?? [] },
--- a/Sources/localvoxtral/ClaudeContext/HerdrClientTTYProbe.swift
+++ b/Sources/localvoxtral/ClaudeContext/HerdrClientTTYProbe.swift
@@ -42,5 +42,29 @@ enum HerdrClientTTYProbe {
     private static let liveProcessNames: @Sendable (dev_t) -> [String]? = { device in
         TTYProcessTable.entries(onDevice: device)?.map(\.name)
     }
+
+    /// How many terminal surfaces this user has a herdr client on, or nil when
+    /// the process table cannot be walked.
+    ///
+    /// A herdr server is a detached daemon with no controlling terminal, so
+    /// counting DEVICES that host a `herdr` process counts clients, and counts
+    /// two clients in one window as the two surfaces they are.
+    ///
+    /// One caller, the local herdr arm: herdr keeps ONE machine selection per
+    /// user rather than one per client, so with a second client on screen that
+    /// selection cannot say which machine the focused surface is showing
+    /// (issue #286). Only the count matters, never which device.
+    static func clientSurfaceCount() -> Int? {
+        clientSurfaceCount(processes: TTYProcessTable.allProcesses())
+    }
+
+    static func clientSurfaceCount(processes: [TTYProcessTable.Entry]?) -> Int? {
+        guard let processes else { return nil }
+        let user = geteuid()
+        let devices = processes.lazy
+            .filter { $0.name == "herdr" && $0.effectiveUserID == user }
+            .compactMap(\.ttyDevice)
+        return Set(devices).count
+    }
 }
 #endif
--- a/Sources/localvoxtral/ClaudeContext/HerdrClientTTYProbe.swift
+++ b/Sources/localvoxtral/ClaudeContext/HerdrClientTTYProbe.swift
@@ -46,14 +46,16 @@ enum HerdrClientTTYProbe {
     /// How many terminal surfaces this user has a herdr client on, or nil when
     /// the process table cannot be walked.
     ///
-    /// A herdr server is a detached daemon with no controlling terminal, so
-    /// counting DEVICES that host a `herdr` process counts clients, and counts
-    /// two clients in one window as the two surfaces they are.
+    /// A herdr server is a detached daemon (`setsid`, no controlling terminal),
+    /// so what this counts is clients. The unit is a JOB on a terminal device,
+    /// not a process and not a device: a client's own child processes share its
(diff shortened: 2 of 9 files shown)


T0mSIlver:
Reviewed by GLM 5.3 flash (opencode, read-only, with the herdr v0.9.0 source available to it). No blocker; four findings acted on in fcb8b89.

- Version pin. herdr refuses a catalog or selection version it does not know and runs the client Local-only. The reader decoded any version, so a future schema renaming these fields would have read as "no machines saved" and retired the guard on a herdr upgrade alone. Anything but version 1 now abstains.
- Explicit `null` selection. herdr treats it as Local and lets it override the copy the catalog carries; the reader fell back to the catalog, abstaining where herdr shows Local. Now follows herdr.
- Surface counting. Counting terminal devices missed two clients sharing one terminal (suspend the first, start the second). It counts jobs now, which is one per client and its children.
- Tests: the null override, the startup-restore shape, both version refusals, a symlinked state file, two jobs on one device, and an assertion that the guard runs before the socket is asked for a pane at all.

One finding I checked and did not take at face value. The review argued that `herdr machine remove`/`disable` in another terminal leaves a running client displaying a machine the files no longer list, so the guard reads Local and the wrong join returns. Its premise was that live clients never re-read the files. They do: `src/client/catalog_reload.rs` polls the catalog once a second and calls `select_local()` when the machine being displayed is removed or disabled, which is also what the 0.9 CLI docs promise. So it is a bounded race, not a standing divergence, and it is recorded as the second residual in `invariants.md` rather than fixed here. The real closure is herdr reporting its active endpoint on the socket, which is the upstream ask in #288.

Suite after the changes:

```
	 Executed 3152 tests, with 2 tests skipped and 0 failures (0 unexpected) in 52.375 (52.509) seconds
```

PR #296, opened 2026-09-15
UI gate: warn once per burst, compiled helper, batch and ax find, parallel state probes

T0mSIlver:
## What

The SSH UI gate is too slow for click-by-click iteration. Measured today from the dev box (fresh ssh, LAN, and identical over a multiplexed connection, so ssh is not the cost): `state` 2.3 s, `ax dump` 1.3 s, `ax click` 13.5 s. The time goes to the gate's own ceremony: every GUI verb speaks the takeover warning, waits 3 s and speaks "done"; every verb re-interprets the 850-line Swift helper with `swift`; `state` runs its probes serially.

This PR keeps every security invariant (no shell verb, token charset, pid scoping, locked-screen refusal, everything logged) and the owner rule that the FIRST GUI action of a burst announces audibly and waits, and changes:

- **Takeover lease.** After the first GUI verb warns, a 0600 lease file (default 120 s, `LV_UI_TAKEOVER_LEASE_SECONDS`, `0` = old behaviour) lets later GUI verbs skip the warning and the wait; "done" is spoken once when the burst ends; a failed verb still says "failed" immediately.
- **Compiled helper.** `swiftc -O` once per helper-source digest into the state dir; stale binaries deleted; interpreted fallback when swiftc is missing or fails, said on stderr and in `state`.
- **`batch` verb.** Newline-separated verbs on stdin, bounded in lines and bytes, every line validated through the same dispatch before anything runs, executed in order, stop at first failure, framed output, one gate-log entry per line; `batch` inside a batch is refused.
- **`ax find <selector>`.** The dump narrowed to matches of the existing selector grammar.
- **`state`.** The five live probes run in parallel; nothing is cached, the lock state least of all.
- **`scripts/mac-ui.sh`.** Dev-box client over one multiplexed ssh connection; `batch` pipes stdin.
- Also found: `validate_selector` accepted a trailing comma. Fixed.

The gate is installed by hand (owner runbook); this PR changes no installed file. Report from the agent that built it follows as the Proof section; the "after" timings are expectations until the owner reinstalls.

## Proof

# UI gate speed-up — report

Branch `ui-gate-fast` off `origin/main` (cf4ec41). Commits (nothing pushed, no PR opened):

```
ea6280d docs: UI gate lease, compiled helper, batch, ax find, mac-ui.sh, hand reinstall
5a59245 scripts/mac-ui.sh: dev-box client for the UI gate over one multiplexed ssh
41724f7 UI gate: takeover lease, compiled helper, batch, ax find, parallel state probes
 docs/agent/field-debugging.md       |  22 ++
 scripts/ci/test-ui-gate.sh          | 616 ++++++++++++++++++++++++++++++++-
 scripts/mac-ui.sh                   |  93 +++++
 scripts/mac/README.md               | 109 +++++-
 scripts/mac/localvoxtral-ui-gate.sh | 666 +++++++++++++++++++++++++++++++-----
```

Files touched: the gate script, the test suite, the new wrapper, the two docs, this report. Nothing else.

### 1. Takeover lease (`announce_takeover` / `arm_done_announcer`)

- First GUI verb of a burst: `say` warning + `sleep $LV_UI_WARN_SLEEP_SECONDS` (3) exactly as before, then writes `$STATE_DIR/takeover.lease` (0600, `umask 077` + rename): `since=<epoch>`, `nonce=<token>`.
- A GUI verb that finds a lease younger than `LV_UI_TAKEOVER_LEASE_SECONDS` (default 120; env/conf-overridable; a non-integer conf value falls back to 120; `0` disables the lease = v1 warn-on-every-verb) skips `say`+`sleep` and rewrites the lease (fresh `since`, fresh nonce). A lease dated in the future is not honoured.
- **"done" once per burst — the choice.** Every GUI verb, at exit, detaches one announcer: a subshell (`trap '' HUP`, all three fds to `/dev/null`, `disown`) that `sleep`s one lease window, re-reads the lease, and speaks "done" only if its own nonce is still the one on file. A later verb rewrites the nonce, so every earlier announcer wakes to a foreign nonce and exits silently; the last verb's announcer is the one that speaks. Picked over "kill the previous sleeper by pid" (a recycled pid makes `kill` the hazard) and over "the next verb that finds an expired lease says done" (no burst-end announcement until the owner's next session). No loop, no pid to track, lifetime bounded by `sleep` itself — that is what "cannot leave a stuck background process" reduces to. `announcer=<pid>` is appended to the lease for the test harness to join on; the gate never reads it.
- A failed verb says "failed" synchronously at exit, lease or not (and still arms the burst's "done").
- `state` reports `takeover: {leased, remaining_seconds, lease_seconds}`, read from the file and the clock, never remembered. The lock state is never cached anywhere.
- Clock seam `LV_UI_NOW_EPOCH` (`now_epoch()`), same trust class as `LV_UI_WARN_SLEEP_SECONDS=0` in the conf; unset/malformed = wall clock. It is what lets the suite move through a lease window without sleeping.

### 2. Compiled helper (`prepare_helper` / `helper`)

- The heredoc moved into `helper_source()`. `write_helper` stages it to `ui-gate-helper.swift.new.$$` and `cmp`s; the file is replaced only when it differs (inode pinned by a test).
- `prepare_helper` (once per invocation): SHA-256 of the source → `$STATE_DIR/ui-gate-helper-<first 16 hex>`. Binary present → compiled. Else `$LV_UI_SWIFTC -O -o <bin>.tmp.$$ <source>` (stderr to `ui-gate-helper-compile.log`), `chmod 0700`, rename, delete every other `ui-gate-helper-*` binary/marker (keeps `*.log`, `*.tmp.*`), log `HELPER compiled <name>`. swiftc absent → interpreted (`no-swiftc`). swiftc failed → `<bin>.failed` marker, log line, interpreted (`swiftc-failed`), not retried until the source changes or the marker is removed.
- Which path ran: a compile or a failed compile writes a `HELPER` line to the gate log when it happens; every interpreted invocation prints `helper running INTERPRETED (<reason>) — slow` on stderr (a per-invocation log line would land after the verb's ALLOW line and break the "last line is ALLOW" contract the suite and `gate-log` rely on); `state` reports `setup.helper` (`{"mode":"compiled","binary":…}` / `{"mode":"interpreted","reason":…}`).
- `LV_UI_SWIFTC` is a seam (default `swiftc`): the suite points it at a stand-in, or at nothing, so a Mac running the suite never compiles the real helper under the stubs.

### 3. `batch` (`run_batch`, `dispatch_command`)

- The former top-level dispatch is now `dispatch_command <line>`: the same whole-command rejection, split, charset loop and `case`, for a lone command and for every batch line.
- stdin bounded by bytes at the read (`head -c MAX+1`, `LV_UI_BATCH_MAX_BYTES` 16384) then by lines (`LV_UI_BATCH_MAX_LINES` 64); blank lines are spacers; each line still bounded by `LV_UI_MAX_COMMAND_BYTES` inside dispatch.
- Pass 1 runs every line through `dispatch_command` in a subshell with `DISPATCH_CHECK_ONLY=1`; every `run_*` calls `check_only_stop` right after its ARGUMENT validation and before its first lock probe / state read / side effect (`term open` skips only its lock-first probe in this mode; the lock is read when the line runs). A `deny` inside logs `DENY <line> (batch line k/N check: <reason>)`; the parent exits 126 with "nothing was run".
- Pass 2: `log_command ALLOW` for `batch (lines=N)`, then each line in its own subshell with the gate's EXIT trap re-installed (a failing line says "failed", a GUI line takes/refreshes the lease, a `shot` cleans its file), logged as its own entry `(batch line k/N: …)`, stop at the first non-zero status and say where.
- Framing: `==lvui-batch-<tag>== lines=N`, then per line `… line k/N begin <verb, ax type text redacted>` and `… line k/N end status=<n>`; `<tag>` is per-batch random, so no verb output can forge a frame.
- `batch` is denied as a line in both passes (`BATCH_DEPTH`), and takes no arguments.

### 4. `ax find <selector>` (`run_ax_find`, helper `axfind`)

Same `validate_selector`, same `ElementSelector`/`collectMatches` walker as `axclick`; `index=` picks one match, `window=` scopes roots. Renders each match with a shared top-level `renderNode(…, maxDepth:, budget:)` (refactored out of `axdump`): dump keeps 40 levels / 20k nodes, find gets 8 levels / 4k nodes. Read-only, no warning, needs unlocked screen + app under test.

### 5. `state`

Five live probes run concurrently into a private `mktemp -d` under the state dir and are joined with `wait`: lock, idle (ioreg), power (pmset), helper `preflight`, control-socket connect. The helper mode is settled in the parent before the fork (a first compile happens there, once). Nothing cached. The idle/pmset pipelines keep the exact text section 14 of the suite pins.

### 6. `scripts/mac-ui.sh`

Host from `LV_UI_HOST` / `git config localvoxtral.uihost` / default `tom@192.168.1.167`; key `LV_UI_KEY` / `~/.ssh/localvoxtral-ui-gate`; `ControlMaster=auto`, `ControlPath=~/.ssh/lv-ui-%C`, `ControlPersist=600`; `batch` gets stdin, every other verb runs with `-n`; `--host`, `--disconnect`. No verb of its own; never touches the installed gate.

### Also found and fixed

`validate_selector` accepted a trailing (or leading, or doubled) comma: `local IFS=,; for pair in $selector` drops the empty field before `[[ -n "$pair" ]]` can see it, so `role=AXButton,` passed the shell and failed only in the helper. Found by the new `ax find` section (which runs with an app under test); section 2's identical `ax click role=AXButton,` case had been passing only because no app was under test there. Now refused on the string, before the loop.

## Timings

Measured today against the INSTALLED (pre-change) gate from this dev box, fresh ssh, LAN:

```
== state: rc=0 2.31s
== gate-log 3: rc=0 0.05s
```

So ssh + gate start is 0.05 s; everything else is probes. The new gate cannot be measured on the Mac until the owner reinstalls it, so per-verb figures are expectations, with the reasoning:

| verb | before (measured) | after (expected) | what is left |
| --- | --- | --- | --- |
| `state` | 2.3 s | ~0.6–1.0 s | the slowest of five parallel probes: the shared lock probe's `swift -` heredoc arm (an interpreter start-up that from sshd always answers `no-session` before the ioreg arm answers for real). Under 1 s is expected, not guaranteed |
| `ax dump` / `ax find` | 1.3 s | ~0.6–0.8 s | lock probe (~0.5 s) + compiled helper start (~0.1–0.2 s) |
| `ax click`, first of a burst | 13.5 s | ~7 s | `say` of the warning (synchronous, ~3–4 s of speech) + the 3 s wait, unchanged by rule; "done" is off the critical path |
| `ax click`, inside the lease | 13.5 s | ~0.6–0.8 s | lock probe + compiled helper |
| `batch` of N lines | N × (above) + N ssh | 0.05 s + Σ per-line | each line re-probes the lock on purpose (a `dictate hold` line can outlast a screen lock) |
| first GUI verb after a reinstall | — | + one compile | `swift build -c release` of the helper took 7.26 s on the build host; `swiftc -O` should be of that order, said on stderr |

The probe that must stay slow: `scripts/ci/screen-lock-state.sh` arm 1 is `swift -` on a heredoc and is a shared file with its own two-caller policy, outside the four files this task allows. From sshd it can never answer (`no-session`), so its ~0.5 s is pure interpreter start-up; the follow-up is to skip arm 1 when `SSH_CONNECTION` is set, in that file.

Wrapper smoke against the installed gate:

```
== --host
host=tom@192.168.1.167
key=/home/dev/.ssh/localvoxtral-ui-gate
== state (cold, opens the ControlMaster): rc=0 2.45s stdout=765B stderr=
== state (multiplexed): rc=0 2.26s stdout=765B stderr=
== gate-log 3 (multiplexed): rc=0 0.05s stdout=164B stderr=
== batch on the OLD gate (expected: denied): rc=126 0.05s stdout=0B stderr=localvoxtral ui gate: denied command
== --disconnect done
```

### Red: the new suite against the pre-change gate

`git show origin/main:scripts/mac/localvoxtral-ui-gate.sh` saved as `old-gate.sh`; run with `LV_UI_GATE_TEST_GATE=old-gate.sh LV_UI_GATE_TEST_KEEP_GOING=1 ./scripts/ci/test-ui-gate.sh` (both are new harness knobs: the first points the suite at another copy of the gate, the second makes `fail` record and continue so every new assertion is shown in one run; CI sets neither). Result: `89 assertion(s) FAILED`, exit 1. The 89th is "the doctor did not compare the installed gate with this checkout", an artefact of the override (section 27 compares the doctor against `$GATE`); the other 88, abridged:

```
FAIL: lease: no lease file was written after the first GUI verb
FAIL: lease: the lease file is mode '', not 0600
FAIL: lease: the lease did not record the clock:
FAIL: lease: a GUI verb 50 s into a 120 s lease spoke the takeover warning again
FAIL: lease: a GUI verb inside the lease still waited the warning pause
FAIL: lease: the lease was not refreshed by the leased verb:
FAIL: lease: a failed verb inside the lease spoke the takeover warning
FAIL: lease: a read-only verb touched the lease:
FAIL: state reported a live lease 1100 s after it was written: {...}
FAIL: state did not report the live lease: {...}
FAIL: state did not report the lease's remaining seconds: {...}
FAIL: lease: expected two distinct announcers, got '' and ''
FAIL: lease: the first announcer exited before the window — the sleep hold did not take
FAIL: lease: done was spoken while both announcers were still inside their window
FAIL: lease: expected exactly one done for a two-verb burst, say.log: localvoxtral u i gate taking control in 3: clicking in localvoxtral
FAIL: compiled: expected exactly one swiftc invocation, got:
FAIL: compiled: swiftc was not run with -O:
FAIL: compiled: swiftc was not handed the staged helper source:
FAIL: compiled: no digest-named helper binary in .../.localvoxtral-ui-gate: app.state
FAIL: compiled: the helper binary is mode '', not 0700
FAIL: compiled: ax dump did not run through the compiled binary: .../ui-gate-helper.swift axdump 4242 all
FAIL: compiled: the gate log does not record the compile: ... ALLOW ax dump all (pane=all)
FAIL: compiled: a second invocation recompiled:
FAIL: compiled: the cached call did not run the binary: .../ui-gate-helper.swift axdump 4242 all
FAIL: compiled: a changed helper source did not recompile:
FAIL: compiled: no binary after the recompile: app.state
FAIL: compiled: the changed source compiled to the SAME digest name:
FAIL: fallback: the gate log does not record the failed compile: ... ALLOW ax dump all (pane=all)
FAIL: fallback: stderr did not say the interpreted path ran and why:
FAIL: fallback: no failure marker was left for this digest: app.state
FAIL: fallback: a failed compile was retried on the next verb:
FAIL: fallback: the second interpreted run was silent about it:
FAIL: state did not report the interpreted helper: {...}
FAIL: state did not report why the helper is interpreted: {...}
FAIL: state did not report the compiled helper: {...}
FAIL: state did not name the helper binary: {...}
FAIL: no swiftc: stderr did not say why the interpreted path ran:
FAIL: batch: a valid three-line batch failed: localvoxtral ui gate: denied command
FAIL: batch: no tagged header line in the output:
FAIL: batch: no begin frame for line 1:            (… and lines 2, 3, and their end frames)
FAIL: batch: the begin frame does not name the verb:
FAIL: batch: ax dump's output is missing:
FAIL: batch: line 1 did not run                   (… and lines 2, 3)
FAIL: batch: no batch-level ALLOW entry: ... DENY batch (unallowlisted)
FAIL: batch: line 1 has no gate-log entry of its own   (… and lines 2, 3)
FAIL: batch: the refusal did not say nothing was run: localvoxtral ui gate: denied command   (×8, one per invalid batch)
FAIL: batch: the refusing line was not logged with its position and reason: ... DENY batch (unallowlisted)
FAIL: batch: nesting was not logged as the reason
FAIL: batch: expected the failing line's status 1, got 126 (localvoxtral ui gate: denied command)
FAIL: batch: the failing line's end frame is missing:
FAIL: batch: line 1 did not run before the failure
FAIL: batch: the failing line was never attempted
FAIL: batch: stderr did not say where it stopped: localvoxtral ui gate: denied command
FAIL: batch: the failing GUI line did not say failed
FAIL: batch: gate-log did not run while locked:
FAIL: batch: the locked-screen denial is not framed:
FAIL: batch: ax type inside a batch failed: localvoxtral ui gate: denied command
FAIL: batch: ax type's text did not reach the helper as argv:
FAIL: batch: the frame did not redact ax type's text:
FAIL: batch: ax type's log entry is missing or unredacted: ... DENY batch (unallowlisted)
FAIL: batch: blank spacer lines were not tolerated: localvoxtral ui gate: denied command
FAIL: batch: blank lines were counted as verbs:
FAIL: batch: a leased GUI batch failed: localvoxtral ui gate: denied command
FAIL: batch: three GUI lines under a lease warned 0 times
FAIL: ax find by role and title: expected exit 0, got 126 (stderr: localvoxtral ui gate: denied command)
FAIL: ax find by role and title: no ALLOW line was logged (last line: ... DENY ax find role=AXButton,title~Text+Processing (unknown ax subverb))
FAIL: ax find did not print the helper's JSON:
FAIL: ax find did not hand the selector to the helper scoped to the app pid:
FAIL: the helper has no axfind subcommand
FAIL: axfind does not resolve its matches through collectMatches
FAIL: axfind does not parse its selector with ElementSelector
FAIL: axfind does not render matches with the shared renderNode
```

(The `ax find … bogus=1` / trailing-comma / nested-`batch` DENIALS pass on the old gate too — it denies every unknown subverb and `batch` wholesale — so the denial half of those cases is by construction not red; the "nothing executed" and "logged with its reason" halves are, above.)

The first red attempt died at "no lease file was written": two assignments from `stat`/`sed` on a missing file ended the run under `set -e` before keep-going could record them. Fixed in the harness (`file_mode`, `|| true`); no assertion was weakened.

### Green: the new suite against the new gate

```
== 29. the takeover lease: one warning per burst, one done per burst ==
PASS: the first GUI verb warns, waits, writes a 0600 lease, and done follows the window
PASS: a GUI verb inside the lease window skips the warning and the wait, and refreshes the lease
PASS: a GUI verb after the lease window warns and waits again
PASS: a failed verb inside the lease says failed immediately, without re-warning
PASS: a lease from the future is not a lease
PASS: read-only verbs leave the lease alone, and state reports whether the next click will warn
PASS: done is spoken exactly once per burst: a superseded announcer stays silent
PASS: LV_UI_TAKEOVER_LEASE_SECONDS=0 restores warn-and-done on every verb
PASS: a junk lease value in the conf falls back to the default
== 30. the helper is compiled once per revision, and interpreted only as a fallback ==
PASS: the first helper call compiles the source with swiftc -O and runs the binary
PASS: an unchanged source reuses the binary and is not rewritten
PASS: a changed helper source compiles to a new digest and deletes the stale binary
PASS: a failing swiftc falls back to the interpreter, says so, and is not retried per verb
PASS: state reports setup.helper: the mode, and the binary or the reason
PASS: without swiftc the helper runs interpreted and says so
== 31. batch — several verbs, one connection, the same gate ==
PASS: allowed: a valid batch runs every line in order, framed, each logged on its own
PASS: denied: batch refused whole for: ax dump all|ax click bogus=1|key escape|
PASS: denied: batch refused whole for: ax dump all|echo pwned|
PASS: denied: batch refused whole for: key escape|state; id|
PASS: denied: batch refused whole for: ax dump all|ax type role=AXTextField hello|
PASS: denied: batch refused whole for: ax dump all|key cmd+q|
PASS: denied: batch refused whole for: ax dump all|launch|
PASS: denied: batch refused whole for: ax dump all|term open ghostty bash -c id|
PASS: denied: batch refused whole for: ax dump all|dictate hold soon|
PASS: denied: a batch with any invalid line is refused before anything runs
PASS: denied: batch containing batch: batch|
PASS: denied: batch containing batch: ax dump all|batch|
PASS: denied: batch containing batch: batch|ax dump all|
PASS: denied: no batch line may be batch
PASS: a batch stops at the first failing verb
PASS: denied: a batch stops at a locked-screen refusal, after the read-only lines before it
PASS: ax type inside a batch keeps its -- rule, types the text, and redacts it everywhere
PASS: denied: batch over the line cap
PASS: denied: batch over the byte cap
PASS: denied: batch with nothing on stdin
PASS: denied: batch with only blank lines
PASS: denied: batch takes no arguments
PASS: batch is bounded in lines and bytes, and blank lines are spacers, not verbs
PASS: a batch of GUI verbs under the lease speaks the warning once
== 32. ax find — the dump narrowed to a selector ==
PASS: denied: ax find without a selector
PASS: denied: ax find with an unknown selector key
PASS: denied: ax find naming no matchable attribute
PASS: denied: ax find with an empty value
PASS: denied: ax find with two arguments
PASS: denied: ax find with a metacharacter
PASS: denied: ax find with a trailing comma
PASS: allowed: ax find by role and title
PASS: ax find is read-only, pid-scoped, and silent
PASS: denied: ax find on a locked screen
PASS: denied: ax find with no app under test
PASS: the helper's axfind shares the selector, walker and renderer with click and dump
```

"ax find with a trailing comma" went red on the FIRST green run (exit 0 instead of 126) — the `validate_selector` finding above; green after the fix.

### Full suite

```
$ ./scripts/ci/test-ui-gate.sh
…
== 28. nothing here needs a bash newer than the Mac's /bin/bash 3.2 ==
PASS: the gate, the wrapper, the installer, the doctor and this suite are bash 3.2 clean
PASS: every array slice in the gate sits behind a length guard

ui gate tests passed
./scripts/ci/test-ui-gate.sh  8.81s user 7.55s system 105% cpu 15.491 total
```

415 PASS (364 before this branch, 10.7 s), 0 FAIL, 2 SKIP as before on Linux (`swiftc` type-check, live ioreg/pmset). None of the 364 pre-existing assertions was edited except the two heredoc-extractor patterns (`cat <<'SWIFT'` now lives in `helper_source()`). Harness additions: lease OFF by default (`LEASE_SECONDS`, so every earlier section reads as written), `NOW_EPOCH`, `SWIFTC_CMD` → `LV_UI_SWIFTC`, `wait_for_done_announcer` joins the detached announcer after every gate run (no announcer from one case writes into the next case's `say.log`), a `swiftc` stand-in that installs the swift stub as the binary, the `sleep` stub records durations and can hold an announcer's window open, `file_mode`, `LV_UI_GATE_TEST_KEEP_GOING`, `LV_UI_GATE_TEST_GATE`.

### `bash -n` and shellcheck

`bash -n` clean on the gate, the wrapper and the suite (run before every suite run above). shellcheck 0.10.0 (static release, downloaded into the scratchpad; not installed on the box), `-s bash`:

- gate: 5 findings, the IDENTICAL five as `origin/main`'s copy (SC2016 ×4 on single-quoted backticks in messages, SC2048 in `list_contains`) — no new finding.
- `scripts/mac-ui.sh`: clean, rc=0.
- test suite: 23 findings before, 23 after (the same pre-existing set; the two `ls | grep` and one `for … in $(sed)` I first wrote were replaced with a glob loop and a `while read`).

### Real compile of the helper on the Mac

The build gate allows `swift build`; `swiftc` and the suite itself are not payloads it accepts. So the embedded helper (847 lines, the section-13 extractor) was staged as a throwaway, untracked SwiftPM executable package inside the worktree and built in the release configuration (`-O`), then the package was deleted (never committed):

```
$ ./scripts/mac-health.sh
== Mac build host: builder@192.168.1.167 ==
OK: SSH gate reachable (gate v2)
$ ./scripts/remote-build.sh build --package-path .lv-helper-probe -c release
==> Running on builder@192.168.1.167: cd work/localvoxtral-agent-a1d5fc2f64adc1452-debcb725 && swift build --package-path .lv-helper-probe -c release
Building for production...
[3/5] Compiling lvuihelper main.swift
[4/5] Linking lvuihelper
Build complete! (7.26s)
```

So the new Swift (`axfind`, the shared `renderNode`) compiles and links with the real toolchain, optimised, with no diagnostics. The gate's own `swiftc -O` path (naming, caching, invalidation, fallback) was exercised only with the stand-in compiler in the suite.

## Not verified

- **The new gate on the Mac.** Never installed (by rule). All per-verb "after" timings are expectations; the `state` < 1 s target hinges on the shared lock probe's interpreter arm, which is outside this task's files.
- **TCC attribution of the compiled binary under sshd.** Expected to match the interpreter's (both are children of sshd; the grants are on `sshd-keygen-wrapper`), confirmed only by the first `ax dump`/`state` after the owner reinstalls (`tcc.accessibility` in `state`).
- **`swiftc` on the sshd PATH.** `swift` is found there today (the interpreted path works) and `/usr/bin/swiftc` is the CLT shim beside it; if it were not found, the gate says `INTERPRETED (no-swiftc)` on stderr and `state` reports it — nothing breaks, it stays slow.
- **`say` timing.** Not measured; the ~3–4 s speech estimate is from the sentence length.
- **Detached announcer under the real sshd.** Its fds are closed and HUP is ignored, so the session should end when the verb does; verified only under bash on Linux (the harness joins it). The first burst on the Mac will show whether an ssh command lingers for the lease window (it should not).

## Owner: exact hand-reinstall

From an owner GUI session on the Mac, after this lands on `main`:

```bash
cd ~/work/localvoxtral && git pull
install -m 0755 scripts/mac/localvoxtral-ui-gate.sh "$HOME/bin/localvoxtral-ui-gate.sh"
```

`authorized_keys`, the lock probe and the TCC grants are untouched. Check with `./scripts/mac-ui.sh state` from the dev box: `setup.gate.revision` must equal the first 12 hex of `shasum -a 256 scripts/mac/localvoxtral-ui-gate.sh`; `setup.helper.mode` reads `compiled` after the first helper-using verb (that one call compiles, tens of seconds, said on stderr). Optionally `git config localvoxtral.uihost tom@192.168.1.167` on the dev box (the default already matches).


T0mSIlver:
Measured on the owner's Mac after the merge, running the merged script from the checkout (`SSH_ORIGINAL_COMMAND='<verb>' scripts/mac/localvoxtral-ui-gate.sh`) against a dogfood build with the Settings window open. The PR body's "after" column was an expectation; these are the real numbers.

| verb | new gate (3 runs) | old installed gate | speed-up |
| --- | --- | --- | --- |
| `state` | 0.39 / 0.38 / 0.38 s | 2.27–2.29 s | ~6x |
| `ax dump` | 0.37 / 0.33 / 0.32 s | 1.26–1.27 s | ~4x |
| `ax find role=AXButton,desc=herdr` | 0.35 / 0.35 / 0.34 s | verb is new | — |
| `ax click` #1 of a burst (warns + waits) | 10.51 s | 13.5 s | ~1.3x |
| `ax click` #2, under the lease | 0.39 s | 13.5 s | ~35x |
| `ax click` #3, under the lease | 0.54 s | 13.5 s | ~25x |
| `menu click Settings` (first GUI verb) | 11.66 s | — | warns and waits, as the owner rule requires |
| `batch` (state + ax find + ax dump) | 1.04 s, 9832 B | verb refused by the old gate | one invocation replaces three round trips |

Behaviour confirmed on real hardware, not stubs:

- **One warning per burst.** The gate log shows `menu click Settings` taking the lease at 10:57:41, then three clicks at 11:01:03–11:01:14 allowed under it with no second warning. One takeover warning was spoken for the whole burst.
- **Compiled helper reused.** `setup.helper` stayed `{"mode":"compiled","binary":"ui-gate-helper-5dfd77c3401a1bbf"}` across every call. The one-time compile cost 8.67 s on its first run the night before, announced on stderr.
- **Announcers do not leak.** Three gate processes right after the burst, one per allowed GUI verb, and none 150 s later, which is the lease window doing what the design says.
- **Fail-closed still holds.** The screen locked for about ten seconds mid-session and the three clicks sent in that window were refused with `DENY … (screen is locked)` in 0.26 s each. The numbers above are from the clean re-run.

`ax find` earns its place: the same query that needs a 4.9 KB dump returns `[{"role":"AXButton","desc":"herdr",…}]`.

The batch saving is the ssh round trips rather than the work itself: 1.04 s for three verbs, against about the same total when run singly from the Mac itself, where there is no ssh handshake to pay. From the dev box each separate verb also pays a connection.

Still pending: the owner installs the script by hand, which is deliberate, with `install -m 0755 scripts/mac/localvoxtral-ui-gate.sh "$HOME/bin/localvoxtral-ui-gate.sh"`. Until then `state.setup.gate.revision` reports the old script's hash.

PR #327, opened 2026-09-17
Keychain prompts: CI stops asking the owner's Mac for API keys

T0mSIlver:
## What

Every keychain prompt on the owner's Mac in the last three days was CI. `securityd` on that Mac names the client:

```
2026-09-17 23:01:29.893 securityd [integrity] ACL partition mismatch: client cdhash:ffd209009473b5a1f203b4d4a749c895646e5b0e ACL (
2026-09-17 23:01:29.894 securityd [kcacl] displaying keychain prompt for /private/var/folders/.../T/tmp.YdkNrFXqrQ/localvoxtral.app(86782)
2026-09-17 23:01:48.108 securityd [kcacl] user approved 'always allow' for .../tmp.YdkNrFXqrQ/localvoxtral.app
2026-09-17 23:22:26.682 securityd [kcacl] user did not approve 'allow' for .../tmp.bOebQTsa43/localvoxtral.app
```

`/private/var/.../T/tmp.*/localvoxtral.app` is the launch smoke's copy of the bundle CI had just built. A 3-day tally of `displaying keychain prompt` on that machine returns those tmp paths and nothing else, one per CI run (21:01Z, 21:22Z, 21:27Z against runs started 20:58Z, 21:18Z, 21:24Z).

Root cause: the app has no Team ID (self-signed `localvoxtral-dev`, and ad-hoc for releases), so macOS partitions each keychain item by the writing build's code-signing hash — `cdhash:…`, not `teamid:…`. Every fresh build mismatches, and `SettingsStore.init` read all three API keys at startup, so each CI launch raised a dialog on the owner's desktop. "Always Allow" only whitelists that one build.

Two changes:

1. **CI keeps out of the login keychain.** `LOCALVOXTRAL_DISABLE_LOGIN_KEYCHAIN=1` selects a process-local secret store; the launch smoke's existing `LOCALVOXTRAL_SUPPRESS_STARTUP_PERMISSION_PROMPTS=1` implies it (so `ci.yml` ×2 and `release.yml` need no edit), and the UI-smoke, record-demo and capture-assets steps set it explicitly — those drive the real UI and must keep the production TCC path, so the two flags stay separate. The same flag stops the cmux socket-password reads. `open` does not pass a step's environment to the app it launches, so the three lanes that open the bundle go through `scripts/lib/launch-app.sh`'s `lv_open` (`open --env …`).
2. **The app reads a key only when something can use it** (owner's call): the engines selected at launch, an engine switched on later, the Settings window (it shows the fields), and the diagnostics export (it reports whether each key is set). Managed local mode authenticates with nothing, so the default setup never opens the keychain and a newly installed build never prompts. A refused read is not retried — one dialog is the budget.

Out of scope, stated so it is not lost: a user who *has* selected a hosted engine still gets one prompt per newly installed build. Killing that needs either an all-applications ACL on the items (any process running as the user could then read the keys silently) or a real Developer ID. Owner picked lazy reads over the ACL change.

## Proof

```
Executed 3437 tests, with 7 tests skipped and 0 failures (0 unexpected) in 60.309 (60.476) seconds
```

- [x] Regression test added — failing before the fix, passing after.

The check: a managed-local launch must read no secret. Run against `main`'s `Sources/` (same test body, `BeforeProbeTests`):

```
Test Case '-[localvoxtralTests.BeforeProbeTests testAManagedLocalLaunchReadsNoSecret]' started.
BeforeProbeTests.swift:32: error: XCTAssertEqual failed:
  ("[localvoxtral.SecretKey.realtimeAPIKey, localvoxtral.SecretKey.llmPolishingAPIKey, localvoxtral.SecretKey.mistralAPIKey]")
  is not equal to ("[]") - a managed-local launch must not read the keychain: each read can pop a prompt
Test Case '... testAManagedLocalLaunchReadsNoSecret]' failed (0.251 seconds).
Executed 1 test, with 1 failure (0 unexpected)
```

Same test on this branch, and the suite it ships in (`LazySecretLoadingTests`, 11 cases: per-engine loading, one read per key per process, no write-back of a loaded value, no retry after a refused read, both CI flags):

```
Test Suite 'BeforeProbeTests' passed — Executed 1 test, with 0 failures
Executed 11 tests, with 0 failures (0 unexpected) in 0.022 (0.023) seconds
  testAManagedLocalLaunchReadsNoSecret] passed
  testALaunchWithAHostedEngineLoadsExactlyThatEnginesKey] passed
  testTheDisabledPolishingEngineKeyIsNotRead] passed
  testSwitchingToAHostedEngineLoadsItsKeyThen] passed
  testOpeningSettingsLoadsEveryKeyOnce] passed
  testALoadedSecretIsNeverWrittenBack] passed
  testAKeyTypedByTheUserIsNotReadBackAfterwards] passed
  testARefusedReadIsNotRetried] passed
  testSecretsInUsePerEngineCombination] passed
  testTheCILaunchFlagsKeepTheAppOutOfTheLoginKeychain] passed
  testTheKeychainFlagAloneDoesNotSuppressTheStartupPermissionPass] passed
```

- [x] Field proof, this PR's own CI run: `securityd` on the owner's Mac must log **no** `displaying keychain prompt` while mac-lanes packages and launch-smokes this branch — where every run before it logged one. Posted as a comment once the run finishes.
- [ ] UI change: none. `SettingsView.onAppear` now loads the keys the panes display, which is what the fields already showed.

Tests changed, not weakened: four existing cases asserted the eager read (`testKeysComeBackFromTheSecretStoreOnTheNextLaunch`, `testAReadFailureReportsItselfInsteadOfClaimingTheKeyIsMissing`, `testAStoredKeyBeatsAnEnvironmentOverride`, `testMistralSettingsRoundTripUnderDocumentedKeys`). Each now calls `ensureAllSecretsLoaded()` — the Settings-open path — and keeps its original assertions; the stored-key-beats-env case additionally pins that the env value stands in until the key is fetched.


T0mSIlver:
## Field proof: this PR's own CI run put no dialog on the owner's Mac

`mac-lanes` packaged and launch-smoked this branch (run 35279751001):

```
Package app bundle                                    success  22:01:41Z → 22:02:39Z
Smoke test packaged app launch (end-user simulation)  success  22:03:07Z → 22:03:09Z
```

`securityd` on that Mac, from the moment the run started (00:00 local = 22:00Z):

```
$ log show --start '2026-09-18 00:00:00' --predicate 'subsystem == "com.apple.securityd"' \
    | grep -c 'displaying keychain prompt'
0
```

Zero — where every CI run before this one logged exactly one, naming the smoke's
`/private/var/.../T/tmp.*/localvoxtral.app` as the client.

And the positive half, from the smoke launch itself:

```
2026-09-18 00:03:08.890 localvoxtral[54988] [com.localvoxtral:Secrets] login keychain not used: this run sets LOCALVOXTRAL_SUPPRESS_STARTUP_PERMISSION_PROMPTS or LOCALVOXTRAL_DISABLE_LOGIN_KEYCHAIN
2026-09-18 00:03:09.003 localvoxtral[54988] [com.localvoxtral:Dictation] startup permission prompts suppressed (LOCALVOXTRAL_SUPPRESS_STARTUP_PERMISSION_PROMPTS=1)
```

Both CI jobs green: `build-test` 6m50s, `mac-lanes` 4m19s.

T0mSIlver:
## GLM 5.3 review (opencode, headless) — verdict SHIP with two fixes first

Both fixes are in `d9a8b89`, each with a regression test that fails on the commit before it.

**1. A CI-flagged launch on a not-yet-migrated install would have deleted the user's key.** The legacy plist sweep removes the plist copy once the write succeeds — and a write into the process-local store the CI flag selects always succeeds. The key would land in a dict that dies with the smoke run, the plist copy would be gone, and the migrated flag set. Such a run now skips the sweep entirely.

**2. A refused write let a later fetch clobber a key the user just typed.** `persistSecret` only marked a key loaded on success, so after a failed write (locked keychain) the next `ensureAllSecretsLoaded` — Settings reopened, an engine switched — read the stale stored value and assigned it over the typed one. The key is marked loaded either way now; that matches the contract the existing write-failure test documents ("works this session but is not saved").

Before (`239162a`, both new tests):

```
LazySecretLoadingTests.swift:280: XCTAssertEqual failed: ("nil") is not equal to ("Optional("mk-only-copy")") - the only copy of the key must survive a run that cannot store it
LazySecretLoadingTests.swift:283: XCTAssertFalse failed - and the real next launch must still run the sweep
LazySecretLoadingTests.swift:286: XCTAssertEqual failed: ("mk-only-copy") is not equal to ("") - the CI run gets no key of the user's
LazySecretLoadingTests.swift:259: XCTAssertEqual failed: ("mk-stale") is not equal to ("mk-typed") - the key still works this session
LazySecretLoadingTests.swift:260: XCTAssertFalse failed - a key this process is already holding is never fetched over
Executed 13 tests, with 5 failures (0 unexpected)
```

After (`d9a8b89`):

```
Executed 13 tests, with 0 failures (0 unexpected) in 0.026 (0.028) seconds
Executed 3440 tests, with 7 tests skipped and 0 failures (0 unexpected) in 61.217 (61.379) seconds
```

Also from the review: the diagnostics export's fetch was unpinned (deleting the call kept the suite green) — now covered by `testTheExportFetchesStoredKeysBeforeReportingThem`; `lv_open` was executed by nothing in tier 0 — now `scripts/ci/test-launch-app-shim.sh` asserts its argv against a stubbed `open` (including that only the literal `1` opts in, and that no lane `open`s the bundle directly); and the comment pointer left by the `loadSecrets` → `migrateLegacySecrets` rename.

The review's own checks that came back clean, worth recording: the consumer graph (no path reads a stored key as empty without a prior load — session start, overlay commit, the Settings-only Mistral status surfaces, onboarding's write-then-switch), the `isApplyingStoredSecret` / `loadedSecretKeys` bookkeeping, workflow coverage (the two launch smokes exec the binary directly so step env inherits; dmg-test and eval-e2e never launch the bundle; `try-pr.sh` and `install.sh` deliberately keep the real keychain), and that the four edited pre-existing tests are contract updates rather than weakenings.

PR #336, opened 2026-09-18
Polish: sound-alike matches are offered to the model, never applied; add About you

T0mSIlver:
## What

Polish was writing code terms over ordinary speech. The owner's dictation history (74 dictations, 47 changed by polish) has "on peut" → `toolInput`, "Sans objectif" → `--sans objectif`, "after starting a session" → "a `SessionStart`", "local Voxtral" → `localvoxtral.js`, "OpenAI et Codex" → "OpenShift Codex". Replaying the same raw dictations against GLM 5.3 and Mistral Medium 3.5 with no matcher in the way produced none of these, and fixed most of the names that had been missed (Qwen, Claude Code, Willison, "injection de prompt").

1. **Sound-alike matches no longer rewrite the transcript.** Only a span that normalizes to the term itself is pre-applied ("use auth dot ts" → `useAuth.ts`), and a lone word may only change letter case (French "Sans" equals the flag `--sans` once dashes are ignored). Edit-distance-one, phonetic and aligned-fallback hits are offered to the model instead.
2. **Offered as a plain term list, not `"heard" -> "term"` pairs.** On replay, pairs produced 5 wrong insertions on Mistral Medium where the list produced 3 and no section 0 (GLM: 2 / 2 / 0). The list holds four terms for a dictation of up to 60 words and one more per 15 words after that, never above twelve (a fixed four lost a name when a long dictation damaged five). Every survivor was a file extension nobody said, so such a term is withheld unless the extension was spoken ("dot t s") or the speaker was naming a file ("open…", "regarde…").
3. **Settings → Text Processing → About you.** A few lines the user writes about their work and the names they say. Appended to the system prompt (and to the polishd warmup prefix, so the cached prefix still matches). With it, "Coin 3.6 / Kuen / QN / qu'en 3.6" all came back as Qwen and `pantomvocours.com` as `tomvaucourt.com`.

Tried and dropped: the same "sounds like it AND fits better" rule in the bundled prompts. With no profile present it made both models guess ("Coin 3.6" → Claude / → Code, "H200" → H100), so the bundled prompts are unchanged and the rule lives only in the About-you block and the term-list header. `docs/agent/invariants.md` records both decisions.

Reviewed headless by GLM 5.3 (opencode): no blockers; 2 MAJOR + 9 minor, all fixed in the second commit except two left as they are — the order of an aligned hit relative to demoted phonetic hits inside the four-term cap, and routing the E2E eval's warmup through `PolishPromptWarmup.plan` (byte-identical today because the eval's profile is empty).

[run-llm-eval]

## Proof

- [x] Unit suite green — `./scripts/remote-build.sh` on the pushed head: `Executed 3470 tests, with 7 tests skipped and 0 failures (0 unexpected) in 65.073 (65.242) seconds`
- [x] Regression tests — behaviour half of `SoundAlikeNominationTests`, run against `origin/main` sources:
  ```
  Executed 5 tests, with 5 failures (0 unexpected)
  testEditDistanceOneHitIsNotPreApplied : ("look at the SettingsStore file") is not equal to ("look at the settings stor file")
  testExactPhoneticKeyHitIsNotPreApplied : ("click the terminal pane") is not equal to ("click the terminal pain")
  testLoneWordOnlyChangesLetterCase : ("--sans objectif défini clair") is not equal to ("Sans objectif défini clair")
  testFileNameWithUnspokenExtensionIsNeverOffered : XCTAssertTrue failed
  testNominationsRenderWithoutTheHeardSpan : XCTAssertFalse failed
  ```
  On this branch they pass (suite totals above). `testFieldProseIsNotRewrittenByScreenTerms` pins the five field sentences.
- [x] eval-e2e — `./scripts/remote-build.sh eval-e2e --provider mistral` on the first commit (mistral-medium-3-5; not re-run after the cap change, which only adds terms for dictations over 60 words — the corpus phrases are shorter):
  `== required: 14/14 metric checks passed, known-hard cases fully passing: 53/156, skipped 0, errors 0 ==` — the same 53/156 as the 2026-09-16 run on main. Repo-file cases still recover their names through the term list (`DictationViewModel.swift`, `AuthService.ts`, `useAuth.test.ts`).
- [ ] **Not measured: the bundled 4B.** The LAN llama.cpp server was down and heavy inference stays off the owner's Mac; this PR's LLM lane and the nightly E2E are the first 4B numbers. A weaker model may have leaned on pre-application more than the hosted ones did.
- Older tests whose expectations encoded pre-application were moved to the channel the term now travels in; the recall cases still require the same term from the same heard span, and the abstention cases now also assert nothing is offered.
- [ ] UI change: the About-you row compiles and is covered by no automated UI tier; not yet looked at by hand (`./scripts/try-pr.sh <this PR>` → Settings → Text Processing → Polishing).


T0mSIlver:
Landed on main already: #342 was branched from this branch, and its squash-merge (56037d3) carried every file of this PR byte-for-byte (verified with `git diff origin/main origin/polish-matcher-nominates-only` — only #342's own Settings row differs). Nothing left to merge here. #340 and #344 were restacked directly onto main (stack #350).

PR #365, opened 2026-09-20
Vibe: publish Mistral Vibe sessions from its hooks

T0mSIlver:
## What

First of a stack that makes Mistral Vibe CLI a supported agent, next to Claude Code and opencode. This PR publishes Vibe sessions to the app. #366 installs the hooks from Settings.

Vibe 2.25 has three hook types (`pre_tool`, `post_tool`, `post_agent`) and none of them carries the prompt, a session start or a session end. So the integration works from what exists:

- `localvoxtral-claude-hook --agent vibe` maps `post_tool` on `read_file`, `write_file` and `edit` to `PostToolUse`, and `post_agent` to `Stop`. A subagent payload (non-null `parent_session_id`) is dropped.
- The prior prompt comes from the session's `messages.jsonl`, because nothing else has it. `VibeTranscriptPrompt` reads the last 512 KiB of a regular file of that name owned by the user, opened `O_NOFOLLOW`, and keeps only the newest `role: user` message Vibe did not mark `injected`. The path never crosses the socket. This is a scoped exception to "transcripts are never scraped", approved on 2026-09-20 and written into `docs/agent/invariants.md`.
- Vibe starts each hook in a new session with no controlling terminal, behind an `sh -c` that may or may not have exec'd. `vibeAncestorPID` therefore climbs from the shim's `$PPID` while the process is still in the hook's own session and has no terminal, so the published pid and tty are Vibe's.
- Vibe never sends a session end, so pid liveness alone would let a reused pid keep a dead session joinable on its old tty. Vibe records carry the process start time (`agent_start_us`) and the registry requires the pid's current start time to equal it.
- Vibe records carry `agent: "vibe"` and the registry scopes their ids under `vibe:`. The two Claude-allocated session handles are withheld from Vibe records, because a Vibe started inside a Claude Code session inherits them and would join that Claude view.

Costs of Vibe's hook set, stated in the README and the invariants doc: a session is known only from its first file-tool call or the end of its first turn, and it ends by pid liveness and TTL.

Payload shapes, the detached hook session and the relative `file_path` were measured on Vibe 2.25.4 with a probe hook, and the test fixtures use those shapes.

## Proof

```
Executed 3700 tests, with 8 tests skipped and 0 failures (0 unexpected) in 82.274 (82.464) seconds
```

New behavior is pinned by `VibeHookInputParserTests`, `VibeTranscriptPromptTests`, `VibeAncestorPIDTests`, `VibeSessionScopeTests`, `VibeIntegrationFilesTests`, and by `VibeHookPublisherRunTests`, which runs the publisher against a live broker socket and asserts the registry snapshot (`vibe:abc`, prompt, Vibe's pid and tty, edited file, activity).

- LLM lanes: the lane filter matches this diff (`*ClaudeHookPublisher*`, `integrations/vibe/*`), so they run in CI.
- Not verified by hand: a real Vibe session on the Mac. The build host has no Vibe install; the owner's Mac does the first live run after #366 installs the hooks.
- Regression test for the pid-reuse finding, `testAReusedPidDoesNotKeepADeadVibeSessionJoinable`, with the registry check disabled and then restored:

```
VibeHookPublisherTests.swift:455: error: ... testAReusedPidDoesNotKeepADeadVibeSessionJoinable : XCTAssertNil failed: "ClaudeSessionSnapshot(sessionID: "vibe:abc", ...
Executed 5 tests, with 2 failures (0 unexpected) in 0.295 (0.296) seconds
```

```
Test Suite 'VibeSessionScopeTests' passed
```

- Review: Codex (gpt-5.6-sol, high). Six findings, all fixed in the second commit. See the comment below. GLM 5.3 reviewed it as well; its findings are in a second comment.


T0mSIlver:
Codex review (gpt-5.6-sol, high effort, read-only on 5555b15). Six findings, all fixed in 7e5adca.

|---|---|---|---|
| 1 | blocker | Vibe has no session end, so a reused pid keeps a dead Vibe session joinable on its old tty for the whole TTL. | Records carry the process start time (`agent_start_us`). `ClaudeSessionRegistry.isFresh` requires the pid's current start time to equal it, and an unreadable one is a mismatch. Red and green runs are in the PR body. |
| 2 | major | The session-log read had no deadline. `O_NONBLOCK` does nothing for a regular file, so a stalled volume could run into Vibe's hook timeout. | The read runs on its own thread and is abandoned after 250 ms. The hook then publishes without a prompt. |
| 3 | major | `sh "$HOME/…/publish.sh"` exits non-zero when the script is gone, which Vibe shows on every turn. | Both commands end in `2>/dev/null \|\| :`. `testTheHookCommandSucceedsSilentlyWhenTheShimIsGone` runs the shipped commands against an empty home. |
| 4 | minor | A payload without `parent_session_id` was accepted as top-level. | The field must be present and `null`. |
| 5 | minor | A user line without `injected` was accepted. | `injected` must be present and `false`. The fixtures now use the measured shape. |
| 6 | minor | The docs said non-user lines are never decoded past `role`, but `JSONSerialization` materializes the whole line. | Lines without Vibe's user-role marker are skipped unparsed, and the docs now say what is kept and sent. |

Codex checked and found correct: id namespacing across `claude`, `opencode`, `vibe` and `remote:` including persisted restore, the transcript path never reaching the wire, `O_NOFOLLOW` plus `fstat` on the open descriptor, `$HOME` quoting and the tool matcher, the shim's silence, and unchanged Claude Code and opencode records.

T0mSIlver:
GLM 5.3 review (opencode, read-only, on 7e5adca). No blockers. It ran the shim in a throwaway home and confirmed all six Codex fixes. Three new findings:

|---|---|---|---|
| 1 | major | A `post_tool` payload over the 8 MiB cap was left in the pipe, so Vibe's write broke mid-payload, and the 0.25 s stdin deadline was sized for Claude Code's few kilobytes. | Fixed in fc2dc0d. The Vibe mode reads under a 2 s deadline and drains past the cap. `VibeStdinTests` shows the writer finishing with draining and being cut off without it. Claude Code's path is unchanged. |
| 2 | minor | Rolling back to an older app after a Vibe session was persisted makes the old build fail to decode the whole session store, so every session is dropped until hooks re-register them. | Not changed here. The same happened when `opencode` was added, the store refills from live hooks within seconds, and decoding the agent leniently changes a store every agent shares. It belongs in its own PR. |
| 3 | minor | The README said a `vibe -p` run cannot be joined. One typed into a pane keeps that pane's terminal and joins while it runs. | README corrected. |

PR #366, opened 2026-09-20
Vibe: install the hooks from Settings

T0mSIlver:
## What

Second of the Mistral Vibe stack, on top of #365. Settings gets a **Mistral Vibe** pane whose one row installs the hooks #365 publishes from.

- `VibeHooksInstallService` copies the bundled shim to `~/.vibe/localvoxtral/publish.sh` and keeps a marked block of two `[[hooks]]` tables in `~/.vibe/hooks.toml`, Vibe's documented hook file. A `[[hooks]]` header opens a new table wherever it appears, so the block is valid at the end of any existing file.
- The rules are the opencode row's: a consent sheet before the first write, refusal on a symlink at any path component, refusal on an unreadable or non-UTF-8 file, the existing file's mode kept, atomic writes, a byte-identical re-install, and Remove restoring the user's file byte for byte. It also refuses when one of its two hook names exists outside the block (Vibe deduplicates by exact name), when a marker sits inside a TOML multi-line string, when a key follows the block before the next table header, and when either file changed between its read and its write. A conflict shows in the row as "hooks.toml needs a manual fix."
- The marked-block text rules already existed for the shell rc block. They move from `ClaudeShellRCSetup` into `MarkedTextBlock`, and both writers use them. `ClaudeShellRCSetup` keeps its API and its tests.
- An install that is complete and older than this build is rewritten at launch, like the Claude Code plugin since #349. A partial or unreadable install waits for the row.
- Sidebar row, status dot and the Mistral AI mark (Simple Icons, CC0), packaging into `Contents/Resources/vibe-hooks/`, the UI smoke and README-capture tab lists, and the user docs.

`VIBE_HOME` moves Vibe's directory and a GUI app cannot see it, so the installer always uses `~/.vibe`. The README covers the manual install.

## Proof

```
Executed 3725 tests, with 8 tests skipped and 0 failures (0 unexpected) in 68.125 (68.285) seconds
```

`VibeHooksInstallServiceTests` pins every status, install into an empty home, append after the user's hooks, in-place update, each refusal writing nothing, and remove. `LiveVibeHooksFileSystemTests` runs install and remove against a real temporary home and refuses a symlinked `.vibe`. `testTheShippedBlockInstallsAndReadsBackAsCurrent` installs the shipped `hooks.toml` and `publish.sh` and checks the block's command names the path the live file system writes.

- UI change, not verified by hand: the pane, the row's buttons and the consent sheet. `scripts/ui-smoke.sh` now asserts the tab and its Hooks row, and that lane is scheduled, not per-PR. Hand check for the owner: Settings → Mistral Vibe → Set up…, start `vibe`, run one turn, dictate into the pane and look for the join badge.
- Review: Codex (gpt-5.6-sol, medium). Seven findings: five fixed in the second commit, one answered with a wording change, one declined. See the comment below. GLM 5.3 reviewed it as well; its findings are in a second comment.


T0mSIlver:
Codex review (gpt-5.6-sol, medium effort, read-only on the first commit). Seven findings. Fixes are in 4c2ad3b.

|---|---|---|---|
| 1 | major | Marker lines inside a TOML multi-line string would be treated as our block, and the launch refresh would rewrite the user's string. | Fixed. A marker inside `"""` or `'''` is refused, and status reads unknown, so the launch refresh never runs on it. |
| 2 | major | A key written right after the end marker belongs to our last `[[hooks]]` table. Removing the block would hand it to the table before. | Fixed. Install, update and remove refuse, and the row reads "hooks.toml needs a manual fix." |
| 3 | major | Read-modify-write could overwrite an edit saved in between, and the launch refresh does this with no window open. | Narrowed. Both files are re-read just before the first write and the edit is refused on any change (`changedOnDisk`). The window is now those two calls. No lock exists that an editor would honor, and the code says so. |
| 4 | major | The symlink checks on path components have a TOCTOU gap, since later operations resolve the path again. | Declined here. The opencode writer has the same gap, exploiting it takes a process already running as the user, and `docs/agent/invariants.md` puts same-uid processes outside the threat model. Moving all the writers to `openat`/`renameat` is its own change. |
| 5 | major | The name-collision check missed a quoted `"name"` key, flagged names that only share our prefix, and was not part of status. | Fixed. Exact match on the two shipped names, bare or quoted key, either string style, and status reports `conflictingHooks`. A test pins the names to the shipped `hooks.toml`. |
| 6 | minor | Mixed LF/CRLF files are normalized, and Remove deletes a blank line that was there before install. | Wording. This is `MarkedTextBlock`'s existing behavior from the shell rc writer, moved here verbatim. The README no longer claims every other line is untouched. |
| 7 | minor | The shipped-files test could pass through the checkout fallback without proving the packaged lookup. | Fixed. `VibePackagedFilesTests` resolves both packaged locations from fixture directories and checks `package_app.sh` copies both files there. |

Codex checked and found correct: the `MarkedTextBlock` extraction is behavior-preserving for `ClaudeShellRCSetup`, app actions are serialized by the busy flag, the launch refresh is limited to a complete out-of-date install, appending is valid TOML for files without a final newline and for CRLF files, and the shim needs no executable bit because the command runs it through `sh`.

T0mSIlver:
GLM 5.3 review (opencode, read-only, on 4c2ad3b). No blockers. It confirmed the five Codex fixes and that the `MarkedTextBlock` extraction is byte-identical for the shell rc writer. Four new findings, all fixed in 5909db7:

|---|---|---|---|
| 1 | major | A `hooks.toml` that defines `hooks` as `hooks = [...]` or `[hooks]` cannot take `[[hooks]]` tables. Appending the block made the whole file unparseable (checked with `tomllib`), which stops every hook the user has, while the row said "Installed." | That shape is refused before any write, and the row reads "hooks.toml needs a manual fix." The README no longer claims the block is valid after any file. |
| 2 | minor | A multi-line string left open at the end of the file was not refused, so the block landed inside it. | Refused. |
| 3 | minor | The alert for a refused install showed the bare word "refused" for five different causes. | `refused` carries a `Refusal`, and each has its own sentence naming the fix. A test checks that they differ and name the file. |
| 4 | minor | A `name` written as a multi-line string could spell one of our hook names past the line-based check. | A multi-line `name` value counts as a conflict. |

PR #369, opened 2026-09-20
Vibe: accept Vibe sessions from enrolled ssh hosts

T0mSIlver:
## What

Third of the Mistral Vibe stack, on top of #366. The Mac accepts Vibe sessions from an enrolled ssh host. #373 adds the Settings row that installs this on a host.

A remote host runs no publisher of ours, so the agent cannot ride inside the record as it does locally. The Vibe shim sends `X-Lvx-Agent: vibe` beside a body in Claude Code's hook shape, which the listener's existing parser already reads.

Mac side:

- `ClaudeRemoteAgentCodec` reads an absent header as Claude Code and refuses a value it does not know, so a newer shim's agent is never filed under Claude Code's join rules. The origin stays `.remote` and the id stays scoped under the host whose token authenticated the request. The agent prefix goes in front (`vibe:remote:<host>:<id>`), so two agents on one host cannot share a key.
- For any agent but Claude Code the listener drops the Remote Control and Claude Desktop session handles, for the reason the local publisher withholds them, and records no Claude plugin version.
- Vibe has no session-end hook and a remote pid cannot be probed. A finished session would therefore stay joinable for the four-hour TTL on the terminal the next one starts in. The host's exit watcher (below) ends it. There is deliberately no Mac-side eviction of an older session by a newer one on the same surface: with Vibe A suspended and B started in the same pane, a dictation into a resumed A would join B. Two candidates make the join abstain.

Host side (`integrations/vibe/remote/`):

- `compact.py` runs on the interpreter that is running Vibe (owner decision). A `post_tool` payload embeds whole files, so it reduces the payload to the fields the Mac keeps plus the short excerpts a remote Claude Code session sends, and reads the prior prompt under the rules of `VibeTranscriptPrompt`. In the test a 300 KB payload leaves the host as under 4 KB.
- `post.sh` keeps the token in a 0600 file (owner decision), reads it into a variable it first unsets, and hands it to curl through a header file, so it reaches no argv and no child's environment. It backs off a dead tunnel like the Claude Code shim and prints nothing, because Vibe shows hook output as a failure.
- The first delivered hook of a session leaves one background watcher, which posts `SessionEnd` when the Vibe process exits. `LOCALVOXTRAL_VIBE_WATCHER=off` disables it.

The residual is written into `docs/agent/invariants.md`: if the watcher's `SessionEnd` never arrives, the dead session stays a candidate for that surface until its TTL.

## Proof

```
Executed 3753 tests, with 8 tests skipped and 0 failures (0 unexpected) in 87.840 (88.042) seconds
```

`VibeRemoteShimTests` runs the real `post.sh` and `compact.py` with a stub curl and reads what they sent back through the listener's own parsers: order of requests, the 300 KB payload, nothing from the session log but the last user message, header injection, the token in no argv and no child environment (with `TOKEN` exported by the parent), backoff, and the watcher posting `SessionEnd` after a fake Vibe (a Python parent that starts the hook in a new session) exits. The listener tests cover the agent header, the dropped handles, `SessionEnd` and the supersede rule.

- Also run by hand on Linux (dash, Python 3.14) against a capture server: same requests, and `SessionEnd` arrived after the fake Vibe exited.
- Not verified: a real Vibe session on a real enrolled host. That needs #373's setup row.
- Review: Codex (gpt-5.6-sol, high). Six findings, see the comment below. GLM 5.3 reviewed it as well; its findings are in a second comment.


T0mSIlver:
Codex review (gpt-5.6-sol, high effort, read-only on fb94c5b). Six findings, all fixed in 19a0c2b.

|---|---|---|---|
| 1 | blocker | The shim never sends `SessionEnd` and a remote pid cannot be probed, so a finished Vibe session stays joinable for four hours on the terminal the next one starts in. That is the normal exit path for Vibe. | The first delivered hook of a session leaves one background watcher on the host (atomic `mkdir` lock, none of the hook's descriptors, pid and start time compared, token re-read when it fires, five tries). It posts `SessionEnd` when Vibe exits. On the Mac, a new remote Vibe session evicts an older one from the same host on the same surface. The window neither covers is stated in the invariants doc. |
| 2 | major | `TOKEN=""` does not clear the export attribute of a variable inherited from the environment, so an exported `TOKEN` in Vibe's environment would hand the bearer token to curl and `compact.py`. The source-text test could not see that. | `set +a` and `unset TOKEN PORT` before the first assignment. The new test exports `TOKEN` in the parent and reads the stub curl's real environment. |
| 3 | major | One `os.pread` can return short, and a short read of a tail window is its older end, so an older prompt could leave the host. | The read loops until the window is full, and gives the prompt up on EOF. |
| 4 | major | Four `ps` calls at one second each, plus two requests, could pass Vibe's five-second hook timeout. | One `ps -ax` under a 0.5 s timeout. The budget is written next to the call. |
| 5 | minor | A Vibe request recorded a missing Claude plugin header as a pre-1.10.0 plugin. | A Vibe request records no plugin version. #373 records the Vibe hooks version under its own name. |
| 6 | minor | `command -v vibe` may name another install's interpreter, `*/python*` was loose, and shebang flags were dropped. | The interpreter is the executable of the hook's parent or grandparent (`/proc/<pid>/exe`, `ps -o comm=`), accepted only as an absolute path named exactly `python`, `python3` or `python3.N`, and run with `-I`. The shebang and `python3` are fallbacks under the same check. |

Codex checked and found correct: a token file without a trailing newline, the token absent from argv, the private header directory, injection through token, port, plan, event, pid and labels, the compactor's allowlist, subagent filtering matching the Swift parser, `X-Lvx-Agent` unable to cross host or agent namespaces, the dropped Claude handles, persisted `vibe:remote:` ids, Python 3.8 grammar and `dash -n`.

T0mSIlver:
GLM 5.3 review (opencode, read-only, on 19a0c2b). No blockers. It ran the shim and the watcher under dash and bash and confirmed the six Codex fixes. Seven new findings, six fixed in 9547473:

|---|---|---|---|
| 1 | major | A Vibe already gone when the hook reads its start time (Ctrl-C at the end of a turn, a closed pane) got no watcher and no `SessionEnd`, although the Mac had just learned of the session. | The watcher skips its wait and posts `SessionEnd` at once. Tested with a fake Vibe that does not wait for its hook. |
| 2 | minor | The lock is keyed by session id only, so a reused id (a resume) kept the old process's watcher, whose `SessionEnd` would evict the live session. | The lock stores the watched pid. A watcher of another pid is stopped and replaced. Tested. |
| 3 | minor | When `ps` fails or times out, `compact.py` published the `sh -c` wrapper as the Vibe pid, and the watcher then ended a live session two seconds after every hook. | With no process table, no pid is published and no watcher starts. Tested with a failing `ps`. |
| 4 | minor | A comment says the Mac reads `X-Lvx-Vibe-Hooks-Version`. In this PR nothing does. | Comment reworded. #373 adds the reader. |
| 5 | minor | The watcher re-read the token but not the port. | It re-reads both. |
| 6 | minor | Locks of killed watchers were never swept. | Locks older than two days are removed. |
| 7 | minor | The Mac-side eviction of an older remote Vibe session by a newer one on the same surface is wrong under job control: suspend Vibe A, start B in that pane, bring A back, and a dictation into A joins B. | The eviction rule is removed. Two candidates on one surface make the join abstain, which is the failure this codebase prefers. The invariants entry says why, so it is not rebuilt. |

PR #373, opened 2026-09-20
Vibe: set up the hooks on an enrolled host from Settings

T0mSIlver:
## What

Fourth and last of the Mistral Vibe stack, on top of #369. Each host row in **Settings → Remote hosts** gets a **Vibe hooks** line with Set up…, Update… and Remove, which installs what #369 accepts.

- The app keeps only a hash of a host's token, and the plaintext went into the Claude Code plugin's config at enrollment. The Vibe hooks therefore get a second credential for the same host (owner decision, chosen over rotating the one token, which would cut off the host's running Claude Code sessions). It authenticates as the same host id, every stored hash is still compared on every request, and Rotate token and Revoke clear it. It is minted per run, bound to the host's token as it was then (a rotation during the run makes the commit refuse), and the token file is the last thing setup writes. The registry trusts the new token next to the previous one until that write has succeeded, so an update that loses its connection cannot lock the host out. Remove withdraws the credential before it contacts the host.
- `setUpRemoteVibeHooks` follows the enrollment rules: `ssh -o BatchMode=yes -o ClearAllForwardings=yes -- <alias> /bin/sh -s`, the script on stdin, the token in no argv on this Mac, fixed error strings. It writes `~/.vibe/localvoxtral/remote/` (0700, files 0600) and edits the host's `~/.vibe/hooks.toml`.
- That edit uses the local install's rules instead of a second implementation in awk. The Mac reads the file back (base64, 256 KiB cap) with its `cksum`, `VibeHooksBlockEditor` computes the new text, and the host writes it only if the `cksum` is unchanged. The hooks.toml rules move out of `VibeHooksInstallService` into that editor, with one instance per block. The remote block has its own markers and hook names, so one machine can hold both.
- The listener records `X-Lvx-Vibe-Hooks-Version` under its own name, which drives the row's Update… state. A Vibe request records nothing about the Claude Code plugin.
- User docs: `docs/remote-claude-context.md`, the Vibe README's "On an ssh host" section, and the integration matrix.

## Proof

```
Executed 3792 tests, with 8 tests skipped and 0 failures (0 unexpected) in 100.430 (100.643) seconds
```

`VibeRemoteHooksSetupTests` does not check script text. Its fake host goes through the production ssh runner with a fake `ssh` that runs `/bin/sh -s` against a temporary `$HOME`, so the runner's real stdin and output limits apply. These runs move files, so they ask the runner for a larger `Invocation.Budget` by name; every other enrollment script keeps the 8 KiB preload. It covers the written files and modes, the token in one stdin and no argv, append after the user's hooks, a byte-identical second run, a `hooks.toml` edited between probe and write (exit 45, nothing written), a symlinked `~/.vibe`, a conflicting `hooks.toml`, no Vibe on the host, host output never reaching an error, remove restoring the user's file byte for byte, both blocks in one file, and the installed `post.sh` dialing with the written token and port. `ClaudeRemoteExtraCredentialTests` covers the credential: same host id, prepare without commit, rotate and revoke, relaunch, an older file.

- UI change, not verified by hand: the Vibe hooks line and its consent sheet. Hand check for the owner: enroll a host that has Vibe, press Set up…, run one Vibe turn over ssh, dictate into that window and look for the join badge, then quit Vibe and confirm the next dictation there has no join.
- `VibeHostFlowTests` drives the model: one credential the host holds after Set up, an update whose last script never reaches the host (the old token still authenticates), a rotation during setup (nothing activated), and Remove with the host unreachable (credential withdrawn, alert says the files remain).
- Review: Codex (gpt-5.6-sol, high). Seven findings, all fixed, see the comment below. GLM 5.3 reviewed it as well; its findings are in a second comment.


T0mSIlver:
Codex review (gpt-5.6-sol, high effort, read-only on the first commit). Seven findings, all fixed in the second commit.

|---|---|---|---|
| 1 | blocker | The production ssh runner refuses stdin over 8 KiB, and the staging script carries `post.sh` and `compact.py` (about 24 KiB). Every real Set up would have failed after the probe. The tests passed because the fake host ran the scripts directly. | `Invocation` carries a `Budget`, standard by default. The Vibe runs ask for 512 KiB. Past the pipe buffer the runner feeds stdin from a writer thread after launch (raw `write(2)`, `F_SETNOSIGPIPE`, our read end closed). The fake host now goes through the production runner with a fake `ssh`, and one test shows the standard budget refusing the staging script. |
| 2 | major | The runner returns 2,000 characters, so the probe's base64 `hooks.toml` lost its closing frame at about 1.4 KiB. | Same budget covers output and message length. A test round-trips a `hooks.toml` over 100 KiB. The 256 KiB cap is now enforced after decoding too, and an oversized file is refused by the host. |
| 3 | major | A credential prepared before a Rotate token could be committed after it. | A pending credential is bound to the host's token generation and the commit refuses when it changed. Rotate, Revoke and Remove are disabled during a Vibe run as well. |
| 4 | major | Remove kept the credential valid when the host was unreachable. | The credential is withdrawn first, the host cleaned up second. If the cleanup fails the alert says the files are still on the host. |
| 5 | major | The token was written before the port and `hooks.toml`, so an interrupted update left the host sending a token the registry did not know. | Setup stages everything else, reads it back, then the registry trusts the new token next to the previous one, then the token file is written last. The old one is retired only after that write succeeded. |
| 6 | minor | The read-back error quoted the version the host reported. | Fixed sentence naming only the version this build expected. |
| 7 | minor | The row's result line outlived the state it described, for example "Set up." after a rotation. | Cleared on rotate, revoke, remove and at the start of each Vibe action. |

Codex checked and found correct: alias, token, port, checksum and mode handling against shell injection, here-document delimiter selection including CRLF and a missing final newline, GNU and BSD `stat`, `base64` and `cksum` portability, loading files without the new key, authentication comparing every stored hash, the text-rule move leaving `VibeHooksInstallService` unchanged, and version reports routed by agent.

T0mSIlver:
GLM 5.3 review (opencode, read-only, on the Codex-fix commit). No blockers. It confirmed the seven Codex fixes, the writer thread in `processRunner`, and that the larger budget changes nothing for existing enrollment calls. Four new findings, all fixed in 1a5490d:

|---|---|---|---|
| 1 | major | On a host without `base64` (it is not POSIX) the probe prints a valid checksum and an empty `hooks=`. The Mac read that as an empty file, replaced the user's `hooks.toml` with our block alone, passed its own read-back, and reported success. | `cksum` prints the byte count. The decoded size must equal it, or the probe is unreadable and nothing is written. Tested end to end with a failing `base64` on the fake host: exit 42, the file unchanged. |
| 2 | major | A link planted at `<file>.lvx-tmp` was written through, for example into `~/.bashrc`. | Temporary names are refused as links, cleared, and created under `set -C`. Tested with a link to `~/.profile`. |
| 3 | minor | `issueCredential` retires the old credential before the host holds the new token, which is the ordering the setup flow must not use. Only tests call it. | It is no longer public and its comment says it is for tests. |
| 4 | minor | A comment claimed the row is hidden without an ssh runner. The code only checked for the bundled files. | The row now needs both. |

T0mSIlver:
## Hand checks left for this stack

Status on 2026-09-21, 00:30. Everything below runs against one build of the top of the stack, which contains all four PRs. CI: `build-test` and `mac-lanes` pass at the head commit of #365, #366, #369 and #373. Two hosted `build-test` jobs failed once in `LearnedTermWiringTests` (merged to main in #370 and #371, not touched by this stack, a different test each time) and passed on rerun.

### What an agent already ran

- **A real Vibe 2.25.4 session against the shipped remote hooks**, on the dev box, with a capture server in place of the Mac. `vibe -p` read a file and answered. The server received `PostToolUse` (absolute path, 60-character excerpt, not the file), `UserPromptSubmit` with the real prompt, `Stop`, and about two seconds after Vibe exited the watcher's `SessionEnd`. Every request carried `X-Lvx-Agent: vibe`, the hooks version and Vibe's pid, which the shim found through `/proc/<pid>/exe`. The watcher's lock was gone afterwards.
- One fact from that run: Vibe writes the user's message to `messages.jsonl` at the end of the turn. A file hook in the middle of a turn therefore carries no prompt on the first turn, and the previous turn's prompt later. The turn-end hook delivers the current one.
- **The build is installed for the UI gate.** A `workflow_dispatch` of CI with `dogfood=true` on `vibe/4-remote-setup` (run 35538995101, green) put it in the gate's `localvoxtral-dogfood.app` slot.

### What could not be run

- **Nothing in the GUI.** The Mac's screen locked while I was switching builds, and the gate refuses `launch` and every GUI verb on a locked screen. I had already quit the gate's previous instance (pid 73838), so **no localvoxtral app is running on the Mac until someone launches one.**
- The gate could not have run the join checks anyway: `term open` allows only `lv-attach`, so it cannot start `vibe` in a terminal.

### To start

With the screen unlocked, from the dev box:

```bash
./scripts/mac-ui.sh launch --dogfood localvoxtral-dogfood.app
```

or on the Mac, `./scripts/try-pr.sh 373`.

### Local (#365, #366)

1. Settings → Mistral Vibe. The row reads "Not installed." with **Set up…**. Press it, confirm the sheet. The row reads "Installed." and offers only **Remove**. An agent can do this step through the gate (`ax click` on `integrations.vibe.install`, then `integrations.vibeSheet.apply`, then `ax find` on `integrations.vibe.status`).
2. `~/.vibe/hooks.toml` holds a block between `# >>> localvoxtral >>>` and `# <<< localvoxtral <<<`, and any hooks that were there before are unchanged. `~/.vibe/localvoxtral/publish.sh` exists.
3. Start `vibe` in Ghostty, iTerm2 or Terminal.app inside a repo. Run one turn that reads or edits a file. Vibe shows no hook warning.
4. Dictate into that pane with Overlay Buffer. The join badge names the repo. A dictation before the first turn has ended has no badge; that is Vibe's missing session-start hook.
5. Quit Vibe, stay in the same shell, dictate again. No badge: the session ended with its process.
6. **Remove**. The block and the script are gone, and the rest of `hooks.toml` is intact.
7. Optional, the refusal path: put `hooks = []` at the top of `~/.vibe/hooks.toml`. The row reads "hooks.toml needs a manual fix." and offers no button.

### On an ssh host (#369, #373)

The dev box (`sandbox-vpn`) is enrolled, has Vibe 2.25.4 and curl, and its tunnel is on port 29891.

1. Settings → Remote hosts. The host's row has a **Vibe hooks** line reading "Not set up.". Press **Set up…**, confirm. It reads "Set up." An agent can do this through the gate too (`claude.remote.host.<id>.vibe.install`, then `claude.remote.vibeSheet.apply`).
2. On the host, `~/.vibe/localvoxtral/remote/` holds `post.sh`, `compact.py`, `token` and `port`, the directory is 0700 and `token` is 0600. `~/.vibe/hooks.toml` has the `localvoxtral remote` block. Claude Code sessions on that host keep working, because they use the host's first token.
3. `ssh` to the host from Ghostty, run one Vibe turn, dictate into that window. The badge names the remote workspace.
4. Quit Vibe on the host, wait five seconds, dictate again. No badge: the watcher's `SessionEnd` reached the Mac. On the host, `ls ${XDG_RUNTIME_DIR:-~/.cache}/localvoxtral/vibe-watch` is empty.
5. **Rotate token** on the host row. The Vibe line returns to "Not set up.". **Set up…** again restores it.
6. **Remove** on the Vibe line. The host's files and block are gone. With the host unreachable, Remove still withdraws the credential and the alert says the files remain.

### If a step fails

Dispatch `mac-crashlog.yml` first (`docs/agent/field-debugging.md`), then report the step number.

PR #393, opened 2026-09-21
Shortcuts: say which mode already holds the key

### LINKED ISSUE #391: Binding one shortcut to both modes fails with a message that hides the reason
Binding the same shortcut to both Overlay Buffer and Live Auto-Paste is
refused by Carbon, not by us: the second `RegisterEventHotKey` on the same
target with the same key and modifiers returns `eventHotKeyExistsErr`,
`registerDual` reports `.livePasteShortcutUnavailable`, and the view model
rolls the slot back and shows

> The selected Live Auto-Paste shortcut is unavailable.

which never says the conflict is with the app's own other shortcut. The user
sees a shortcut vanish for no stated reason.

This has always been reachable for identical modifier combinations. #389 makes
it much easier to reach: bare function keys are now bindable, and one spare F13
is exactly the key someone would try in both slots.

## Scope

Cross-check the two slots in `ShortcutRecorderField.canRecord`, before
registration, and reject with a message that names the cause — "Already used
for Overlay Buffer." / "Already used for Live Auto-Paste." The field needs the
other slot's shortcut passed in; today it knows only its own binding.

Keep the Carbon failure path as the backstop. It is what catches a shortcut
another app took, which we cannot see in advance.

## Proof

- Validation unit tests: same shortcut in both slots rejected with the naming
  message; the same shortcut in one slot with the other empty still accepted;
  clearing one slot frees it for the other.
- By hand: record F13 for Overlay Buffer, try F13 for Live Auto-Paste, confirm
  the message names the conflict and the Overlay Buffer binding survives.

Found by a GLM review of #389.

T0mSIlver:
## What

Binding the same shortcut to both dictation slots was refused by Carbon, not by
us: the second `RegisterEventHotKey` returns `eventHotKeyExistsErr`, the view
model rolled the slot back, and the user read "The selected Live Auto-Paste
shortcut is unavailable." — which never says the conflict is with the app's own
other shortcut. #389 made it easy to reach: a spare F13 is exactly the key
someone tries in both slots.

The recorder now cross-checks the other slot before registration and names it —
"Already used for Overlay Buffer." / "Already used for Live Auto-Paste." The
field is given the other slot's shortcut, since it knew only its own binding.
Reset runs the same check, because it writes the default shortcut without going
through the recorder. Clearing or rebinding one slot clears a message that named
it. The Carbon failure stays the backstop for keys another app holds, which we
cannot see in advance.

Closes #391.

## Proof

  ```
  Executed 3735 tests, with 8 tests skipped and 0 failures (0 unexpected) in 66.502 (66.667) seconds
  ```

- [x] Regression test, failing before and passing after. Before = the field
  ignoring the other slot, which is what the code did:

  ```
  ShortcutValidationTests.swift:236: error: -[localvoxtralTests.ShortcutValidationTests testRecorderField_namesTheOtherSlotHoldingTheShortcut] : XCTAssertEqual failed: ("nil") is not equal to ("Optional("Already used for Overlay Buffer.")")
   Executed 16 tests, with 1 failure (0 unexpected) in 0.245 (0.246) seconds
  ```

  After:

  ```
  Test Suite 'ShortcutValidationTests' passed at 2026-09-21 17:04:35.586.
   Executed 16 tests, with 0 failures (0 unexpected) in 0.002 (0.002) seconds
  ```

  Seven cases: both slots holding the same key (each direction, each message),
  empty other slot, different key, different modifiers, the recorder's
  function-key bit still matching a stored bare F13, and the bare-key rule
  taking precedence over a conflict.

- [ ] UI change verified by hand — **not done, and I cannot do it from here.**
  The flow needs an F13 keypress into the recorder control, and the SSH UI
  gate's `key` verb only sends escape/tab/return. What a hand-check would cover
  beyond the tests: record F13 for Overlay Buffer, try F13 for Live Auto-Paste,
  confirm the message names the conflict and the Overlay Buffer binding
  survives. Pane structure is unchanged — same groups, same rows, one string in
  the existing footer.


T0mSIlver:
CI's first run went red on `LearnedTermWiringTests.testGroundingFromMemoryRefreshesWithoutRewritingProvenance` — unrelated to this diff, and green on rerun with no change. That suite waits for the session on a 1 s wall-clock deadline and asserts regardless; filed as #395. Both required checks pass on the rerun of run 35616746990.

T0mSIlver:
Superseded by #394.

PR #396, opened 2026-09-21
Learned terms: wait for the commit, not for a second to pass

### LINKED ISSUE #392: LearnedTermWiringTests flakes in two different tests, on both hosts
`LearnedTermWiringTests` has failed three times in one afternoon, in two
different tests, and passes on rerun every time. Seen while landing #389, whose
diff touches none of this code.

## Observed

Mac build host, full unit suite:

```
LearnedTermWiringTests.swift:160: error: testConfirmedTermGroundsALaterDictationWithNoLiveSource :
XCTAssertEqual failed: ("Optional("open useauth.ts please")") is not equal to ("Optional("open useAuth.ts please")")
```

GitHub-hosted `build-test`, run 35612171343:

```
LearnedTermWiringTests.swift:200: error: testGroundingFromMemoryRefreshesWithoutRewritingProvenance :
XCTAssertEqual failed: ("Optional(3)") is not equal to ("Optional(4)")
```

Both pass when the suite is run alone (`--filter LearnedTermWiringTests`, 7
tests, 0 failures) and both passed on the very next full run.

## What the two failures have in common

Each test seeds the store with three `.repository` sightings under
`LearnedTermProjectResolver.shared`, then commits one dictation.

- The first asserts the remembered spelling reaches the polish request. It got
  the unremembered text, so grounding found nothing.
- The second asserts `dictations == 4` (three seeds plus the dictation) under
  `projectKey: LearnedTermProjectResolver.shared.key`. It got 3 — the
  dictation's own sighting is missing from that key.

Both are "the commit path and the seeds disagreed about the project", and the
second says specifically that the dictation landed somewhere the read did not
look.

## Two candidates, neither confirmed

1. **The resolved project key is not always `shared`.** The commit path calls
   `LearnedTermProjectResolver.resolve(repositoryRoot:workspace:)` with
   `debugRepoVocabularyRootOverride`, while the seeds and the assertions use
   `LearnedTermProjectResolver.shared`. If `resolve` returns a keyed identity
   under some conditions, the dictation folds under a different key and both
   assertions fail exactly as observed. This fits the evidence best; it needs
   someone to check what `resolve` does when the override is set and what the
   process's git root is on each host.
2. **Unawaited seed writes.** `LearnedTermStore.record` is `writeQueue.async`
   and the three seeding calls never `waitForPendingWrites()`; only the
   `commit` helper does, at the end. The serial queue argues against a lost
   update, so this is the weaker candidate — but the seeding loops are the one
   place in the suite that writes without waiting.

## Constraints

Per the repo's test rules, the fix cannot be a tolerance widened or an
assertion dropped, and must not poll on wall-clock. The store already exposes
`waitForPendingWrites()` as its test seam; if the cause turns out to be the
project key, the seam that is missing is a deterministic resolver for tests.

T0mSIlver:
## What

`LearnedTermWiringTests` waited out a stopped session with a 1 s wall-clock
poll. On the GitHub runner that expired, and the assertions then read a
half-finished commit: the dictation had not folded into the memory yet, so a
term counted 3 instead of 4, and a remembered spelling had not been applied.
Three of the last four hosted runs went red on it while the same commit passed
on the Mac — it blocked #394 twice.

The app now says when it is done. Both stopped-session reset paths go through
one `markStoppedSessionCompleted`, which fires a DEBUG completion handler, and
the test awaits that instead of a clock. No tolerance was widened: the wait has
no deadline to widen any more.

Closes #392.

## Proof

- [x] Cause, demonstrated

Collapsing the deadline to zero (`ContinuousClock.now`) makes the expiry
certain. The suite then fails with CI's exact signature, line for line:

```
LearnedTermWiringTests.swift:200: error: testGroundingFromMemoryRefreshesWithoutRewritingProvenance :
XCTAssertEqual failed: ("Optional(3)") is not equal to ("Optional(4)")
LearnedTermWiringTests.swift:160: error: testConfirmedTermGroundsALaterDictationWithNoLiveSource :
XCTAssertEqual failed: ... "the remembered spelling is placed before the model call"
Executed 7 tests, with 6 failures (0 unexpected)
```

Both tests that failed in CI are there, with the same assertions and the same
values — `Optional(3)` against `Optional(4)` is the run 35617191537 failure
verbatim. The other four failures are the same cause reaching assertions that
had been getting lucky.

- [x] Fixed — `./scripts/remote-build.sh test --filter LearnedTermWiringTests`

```
Executed 7 tests, with 0 failures (0 unexpected) in 0.038 (0.038) seconds
```

0.038 s against 0.295 s before: the suite no longer sleeps, which is the point
— there is no longer a duration for load to exceed.

```
Executed 3728 tests, with 8 tests skipped and 0 failures (0 unexpected) in 68.602 seconds
```

## Not in this PR

Nine other waits across `DictationViewModelOverlayLifecycleTests`,
`PolishTokenGuardTests` and `AgentDictationE2EEvalTests` use the same 1 s
wall-clock poll. None has been seen failing, and several deliberately assert
the `isCompletingStoppedSession` transition around the wait, so each needs
reading rather than a sweep. The hook they would need now exists.


T0mSIlver:
Superseded by #397, which fixes this without touching production code — it awaits `polishAndCommitTask` rather than adding a DEBUG completion hook, and it also caught the unawaited seeding writes, a second cause I had wrongly ruled out (`record` folds on a `.utility` queue while `snapshot()` reads memory directly). Closing.

PR #400, opened 2026-09-21
tests: await the stop-commit's own task in the last four suites

### LINKED ISSUE #398: Four test files still wait for a stop-commit on a 1s wall-clock deadline
#397 fixed the 1 s poll in `LearnedTermWiringTests` by awaiting
`polishAndCommitTask` instead of a deadline. Four other test files still wait
for a stop-commit the old way, and its proof run showed the shape is unsound:
with the polishing service made slower than 1 s, the old helper returned
mid-commit and three of seven tests asserted on an unfinished session.

## Where it is

| File | Shape | Silent past the deadline? |
|---|---|---|
| `PolishTokenGuardTests.swift:2143` | shared helper, 14 call sites | yes |
| `DictationViewModelPolishFailureDiagnosticsTests.swift:313` | shared helper, 3 call sites | yes |
| `DictationViewModelOverlayLifecycleTests.swift:475` + inlined copies at 608, 659, 700, 745 | helper + 4 inlined | 4 of 5 |
| `AgentDictationE2EEvalTests.swift:558` | inlined, 240 s | no — throws `EvalInfraError` |

"Silent" is what matters. `DictationViewModelOverlayLifecycleTests.swift:357`
and the eval lane both check the flag after polling, so an outrun deadline
names itself. The rest fall through and assert on partial state, which is how
#392/#395 presented: a wrong value, on a rerun-green test, in a file whose diff
touched nothing.

The helper at `:475` carries its own note calling this known debt.

## What to do

Await the commit's own task, the way `DogfoodCaptureWiringTests` already does
and `LearnedTermWiringTests` now does:

```swift
let commitTask = viewModel.polishAndCommitTask
await commitTask?.value
XCTAssertFalse(viewModel.isCompletingStoppedSession)
```

Read the task before the first suspension — `finishStoppedSession` has just
stored it and the task clears it on its way out.

## The one case that needs thought

`DictationViewModelOverlayLifecycleTests.swift:339` deliberately asserts
`isCompletingStoppedSession == true` while a suspended polishing service holds
the commit open, then resumes it and waits. Awaiting the task is still right
there — the wait comes after `resumePendingRequest()` — but the poll at 340,
which waits for the service's first call while the commit is deliberately
mid-flight, has no task to await and needs a signal from the fake service
instead. That is why this was not folded into #397.

The eval lane's 240 s wait is the lowest priority: it is tier 2, it reports
itself, and its deadline is sized for a real model rather than a scheduler
hiccup.

T0mSIlver:
## What

Four test files still waited for a stop-commit the way #397 removed from the
fifth: poll `isCompletingStoppedSession` against a 1 s `ContinuousClock`
deadline, then return whichever way it went. Three of them fall through
silently, so a loaded runner asserts on partial state — a wrong value, on a
rerun-green test, in a file whose diff touched nothing. That is how #392/#395
presented.

All of them now await `polishAndCommitTask`, read directly after
`finishStoppedSession` while the value that call just stored is still there,
then assert the session is over before reading what it wrote. A stop that
commits synchronously leaves the task nil and has already finished, so one
helper covers both paths.

Two waits needed more than a swap:

- `testFinishStoppedSessionIgnoresDuplicateCallsWhilePolishingIsInFlight`
  asserts on a commit deliberately held open inside the polishing service, so
  there is no finished task to await at that point. `BlockingMockLLMPolishingService`
  now signals its own arrival, and the test awaits the commit task after
  resuming it.
- The eval lane's 240 s deadline reported itself rather than falling through,
  but a bound sized for a live 4B is a guess either way. It awaits the task
  too; the polish client's own timeout ends a wedged request, and the
  `llmFailed` check below already turns that into an `EvalInfraError`.

Fixes #398.

## Proof

The failure shape is a commit slower than the old deadline, so the regression
run makes every fake polishing service in the three unit suites sleep 1.5 s
before answering. That patch is proof scaffolding, not part of the diff.

Before the fix, with the slow services:

```
Executed 196 tests, with 101 failures (0 unexpected) in 60.318 (60.342) seconds
```

54 distinct tests failed, across all three suites — every one of them reading
state the commit had not written yet.

After the fix, same slow services:

```
Executed 196 tests, with 0 failures (0 unexpected) in 88.727 (88.755) seconds
```

The extra 28 s is the suites now actually waiting for the 1.5 s polish they ask
for instead of giving up at 1 s.

- [x] Unit suite green — `./scripts/remote-build.sh`:

```
Executed 3739 tests, with 8 tests skipped and 0 failures (0 unexpected) in 122.659 (123.107) seconds
```

- [x] Bug fix: failing run before, passing run after — above. No new test: the
  tests being fixed are the regression surface, same as #397.
- [ ] Realtime integration — not run. Test-harness-only diff; nothing reaches
  the model, the helpers, the prompts or the eval corpus, so no lane filter
  matches.
- [ ] UI change — none.

The eval-lane change (`AgentDictationE2EEvalTests`) compiles in the tier-0
build but only runs under `remote-build.sh eval-e2e`, which needs a packaged
app and a live model. It is the same swap as the other three, in the one place
that already reported its own timeout.


T0mSIlver:
The three copies of `awaitStoppedSessionCommit` this PR leaves behind are filed as #402 — the test target has no shared-helper home, which is also why `MockOverlayCoordinator` exists 7 times.

Issue #314, opened 2026-09-16
Long sessions get slower and heavier: the speech engine's unbounded per-utterance buffers


T0mSIlver:
Rescoped. [#319](https://github.com/T0mSIlver/localvoxtral/pull/319) shipped four of the five items originally listed here: the utterance limit is explicit and reports itself on the status line instead of going quiet, the helper sends an `error` event when it is reached, polish timeouts have their own failure kind and message, and `speechd-bench` measures a long run. What is left is the reason the limit has to exist at all.

## What is still open

Two of the engine's buffers grow without bound for the length of one utterance:

- **Conv rows and the adapter buffer** are concatenated on every step and never trimmed, so memory and per-step copy cost grow linearly with utterance length.
- **The transcript is re-decoded from the full token list on every step.**

Everything else is bounded — the decoder KV cache trims to the 8,192-token sliding window (about 11 minutes), the encoder KV caches reset every 750 conv frames.

Measured effect: past roughly 9 minutes the session loses real time, which is what set the 10-minute cap. The cap is a guard against the symptom; these buffers are the cause.

## Two ways out, and they compose

1. **Session rollover in the helper.** Roll to a fresh stream session at a natural pause, carrying the transcript forward. Ours to build, needs no upstream change, and is what makes an always-on session possible rather than just a longer one.
2. **Trim the buffers upstream** in mlx-audio-swift, and stop re-decoding the whole token list each step. The engine-side fix, and the one that helps every consumer.

## Proof expected

- `speechd-bench` over a long run, before and after, with step latency and memory.
- The accuracy gate (`integration-speechd`) unmoved: a rollover that drops or duplicates words at the seam is worse than the slowdown.

T0mSIlver:
Both open items are done.

**The buffers are bounded upstream.** Blaizzy/mlx-audio-swift #263 drops conv and adapter rows once consumed, #264 appends decoder KV rows in place, #265 decodes the transcript one token at a time. #343 pinned their merge (`01dec7c`) here; nothing else between that pin and `8ed8188` touches VoxtralRealtime. A `speechd-bench 10800` on that code held a mean step of 65-81 ms and active memory flat at 4.2 GB from 11 minutes through 3 hours, against 105 ms at 10 minutes and +23 MB per minute before.

**Session rollover is not needed.** It was the way to bound memory without an upstream change. Memory stopped growing, so the seam risk it carried buys nothing.

[...]

PR #401, opened 2026-09-21
speechd: raise the utterance limit to an hour

T0mSIlver:
## What

Raises the helper's default utterance limit from 10 minutes to an hour, and stops calling it a performance ceiling.

The 10-minute default was measured against the engine pinned at the time: decoding fell behind live speech past about 9 minutes, and memory climbed ~23 MB per minute. Blaizzy/mlx-audio-swift #263–#265 removed that growth, and #343 pinned them here — a 3-hour `speechd-bench` on that engine holds 65–81 ms steps and 4.2 GB flat from 11 minutes on. The limit still exists, now only as a guard against a session left running.

The app never passes `--max-utterance-seconds`, so the helper default is what ships. `docs/under-the-hood.md` and the `speechd-bench` usage comment follow the new number.

Refs #314.

## Proof

- [x] Helper suite green — `./scripts/remote-build.sh test --package-path SpeechHelper`: `Executed 58 tests, with 0 failures`.
- [x] Regression test, failing before and passing after. `testDefaultLimitIsAnHourAndPhrasesItselfInOneShortSentence` pins both the duration and the sentence the status line shows, which has to stay inside the popover's 44-character budget.

  Before (this test against `origin/main`'s `UtteranceLimit.swift`):

  ```
  UtteranceLimitTests.swift:26: error: XCTAssertEqual failed: ("600") is not equal to ("3600")
  UtteranceLimitTests.swift:27: error: XCTAssertEqual failed: ("10-minute limit reached; start again.") is not equal to ("60-minute limit reached; start again.")
  Executed 8 tests, with 2 failures
  ```

  After:

  ```
  Test Case '-[SpeechEngineTextTests.UtteranceLimitTests testDefaultLimitIsAnHourAndPhrasesItselfInOneShortSentence]' passed (0.000 seconds).
  Executed 8 tests, with 0 failures
  ```

- [x] Accuracy gate unmoved — `mac-lanes` ran `integration-speechd` by lane filter (`matched SpeechHelper/Sources/SpeechEngineText/UtteranceLimit.swift`), both checks green on 842df3a:

  ```
  Executed 5 tests, with 0 failures (0 unexpected) in 28.284 seconds
  speechd integration: word accuracy 0.804
  ```

  `testUtteranceLimitStopIsReportedOnceInsteadOfSilently` passes its own 3 s limit explicitly, so it still covers the reporting path at the new default.
- [ ] Long `speechd-bench` on the shipped pin, overnight on the build host: the 3-hour numbers this default rests on were measured on the fork branches before upstream merged them. Step latency and memory pasted here when it finishes.


T0mSIlver:
## Bench run 2026-09-23 05:01Z

Did not run: the build host never answered between 05:30 and 07:00 Paris.

```
== Mac build host: builder@192.168.1.167 ==
FAIL: build host unreachable over SSH:
  ssh: connect to host 192.168.1.167 port 22: Connection timed out
  (Mac asleep, owner logged out, or IP changed — update 'git config localvoxtral.buildhost')
CI:  latest run: completed/success 2026-09-23T00:20:19Z localvoxtralCore: the pure pieces build and test without AppKit (#432 step 9)
```

